PR #174, opened 2023-10-08
PyFixest version 0.10

s3alfisc:
- Runs the alternating projection algo through a custom demeaning function. Speeds gains of 5x or more.
- Adds `numba` as a dependency.
- Improves performance of preprocessing by adjusting `model_matrix_fixest`.


pyfixest/demean.py
```diff
@@ -62,7 +62,9 @@ def demean_model(
                 if var_diff.ndim == 1:
                     var_diff = var_diff.reshape(len(var_diff), 1)
 
-                YX_demean_new = algorithm.residualize(var_diff)
+                weights = np.ones(YX.shape[0])
```
s3alfisc, line 65: Weighted least squares is not yet implemented. With the current implementation of the algo, weights need to be provided -> vector of ones.

PR #194, opened 2023-10-22
PyFixest 0.10.7: Basic Did2s support

s3alfisc:
- Adds support for experimental `event_study()` function, which supports Difference-in-Differences estimation via two-way fixed effects and Gardner's two-stage estimator (2021). 
- To support the implementation of Gardner's estimator, now allows to estimate 'shallow' models as `feols("Y ~ 0 | f1)`.

To Do's: 

- [ ] allow for covariates
- [ ] allow for custom clustering variables
- [ ] implement proper event study with leads and lags. Pyfixest's `i()` operator currently only supports one reference level, do I have to adjust it to add a second level? Check. 
- [ ] Write unit tests. 

Example: 

```python
%load_ext autoreload
%autoreload 2
from pyfixest.experimental.did import event_study
from pyfixest.summarize import summary
import pandas as pd

df_het = pd.read_csv("pyfixest/data/df_het.csv")

fit_twfe = event_study(
    data = df_het,
    yname = "dep_var",
    idname= "state",
    tname = "year",
    gname = "g",
    estimator = "twfe"
)

fit_did2s = event_study(
    data = df_het,
    yname = "dep_var",
    idname= "state",
    tname = "year",
    gname = "g",
    estimator = "did2s"
)

summary([fit_twfe, fit_did2s])

# ###
#
# Estimation:  TWFE
# Dep. var.: dep_var, Fixed effects: state+year
# Inference:  CRV1
# Observations:  46500
#
# | Coefficient   |   Estimate |   Std. Error |   t value |   Pr(>|t|) |   2.5 % |   97.5 % |
# |:--------------|-----------:|-------------:|----------:|-----------:|--------:|---------:|
# | zz00_treat    |      2.135 |        0.044 |    48.803 |      0.000 |   2.049 |    2.220 |
# ---
# ###
#
#   Estimation:  DID2S
# Dep. var.: dep_var
# Inference:  CRV1 (GMM)
# Observations:  46500
#
# | Coefficient   |   Estimate |   Std. Error |   t value |   Pr(>|t|) |   2.5 % |   97.5 % |
# |:--------------|-----------:|-------------:|----------:|-----------:|--------:|---------:|
# | zz00_treat    |      2.152 |        0.048 |    45.208 |      0.000 |   2.059 |    2.246 |
# ---
```


s3alfisc:
@kylebutts on one hand, mostly fyi =) But also, some of the code is more or less directly translated from your R package, which I mention in the codebase. I think it would be best practice to actually add your name to the list of package authors (this is what I do with in my R packages if I more or less copy code from another package). What do you think?

kylebutts:
I'm not sure how multiple inheritance works in python, but is there a possibility to pass `first_stage` and `second_stage` directly? For example, I'm not sure how I could specify subgroup-specific time-trends in that notation. The nice thing about did2s (IMO) is the flexibility in modeling $Y_{it}(0)$. 

Though, I do see how it's nice to have a single `event_study` option. Could you have a `did2s` command that `event_study` calls with the standard TWFE formula? Happy to try and help, but my python skills are not very strong.

s3alfisc:
Hi Kyle, Thanks for offering your help! I might have some questions on reference levels for the event study with leads and lags down the line =) 

I thought about having a custom `did2s` function yesterday and basically came to the same conclusion as you: I should definitely support it! It indeed offers a loot of versatility that `event_study()` cannot provide (as you mention, subgroup specific time trends, "pooling" of time intervals in event studies, triple DIDs and probably quite a few other things I am not aware of). It should be fairly straightforward to set up - basically I need to copy the `event_study()` interface and call it `did2s` and add formula arguments to it + `DID2S.estimate()`. And that's it I think =)

s3alfisc:
Ah, maybe not as easy as I thought: mapping from an event study interface to calling did2s is fairly straightforward, but not the other way around. But should still be manageable =)

kylebutts:
What do you think is hard about making a did2s command? Your code looks pretty generic. From what I can tell, the main change you would need to make is in the vcov part:
https://github.com/s3alfisc/pyfixest/blob/f03deb3b8015efbdf7cfac9cda14f5c10db37830/pyfixest/experimental/did.py#L287-L291

s3alfisc:
Yes, you're right, it's just that the closer time gets to midnight, the faster I type and the worse I think 😅 

basically, the steps are the following: 

- create a did2s function 
- Initiate the DID2S object 
- Overwrite the formula attributes with the user selected formulas
- update estimate() and vcov() 

Really not too hard :)

kylebutts:
lol I'm getting push notifications of each attempt @s3alfisc 😅 godspeed on this

s3alfisc:
Oh no 🙈 I had my fingers crossed GitHub would stop notifying you after the 10th attempt or so 🙃 sorry about that! I'm definitely done for today 😅

PR #273, opened 2024-01-28
`WLS` for `feols()`

s3alfisc:


s3alfisc:
To do: 
- fix R2 for WLS
- adjust CRV3 inference
- error when `wildboottest` and `weights not None`

PR #286, opened 2024-02-04
added implementation of simultaneous confidence intervals

apoorvalal:
Added an implementation of the multiplier bootstrap for simultaneous confidence intervals based on sec 4.2 [here](https://arxiv.org/pdf/2205.09691.pdf#page=15). Let me know if it looks good; will write tests soon.


s3alfisc:
Hey that's super cool @apoorvalal! Somehow I missed this last night else I would have replied earlier =) I've left few small comments in the review.

apoorvalal:
Now more concise reference: remark 4.4.1 in [this new book](https://www.causalml-book.org/)

s3alfisc:
> Now more concise reference: remark 4.4.1 in [this new book](https://www.causalml-book.org/)

Nice, this is very helpful. I am super excited about this book btw! 😄

s3alfisc:
Hi @apoorvalal , I slightly refactored your code and introduced two new function arguments `drop` and `keep`, that allow you to select the coefficients for which you want to compute the simultaneous CIs (first implemented for etable by @Wenzhi-Ding). The syntax now is 
```py
# load autoreload extension 
%load_ext autoreload
%autoreload 2

from pyfixest.estimation import feols
from pyfixest.summarize import etable
from pyfixest.utils import get_data
import pandas as pd
import numpy as np

data = get_data()
fit = feols("Y ~ X1 + X2 + C(f1)", data=data)

fit.confint(keep = ["X1", "X2"], alpha = 0.05)

#     0.025%	0.975%
# X1	-1.079777	-0.819274
# X2	-0.208755	-0.139695

joint_cis = fit.confint(keep = ["X1", "X2"], alpha = 0.05, joint_cis = True)
joint_cis
#       	0.025%	  0.975%
# X1	-1.097158	-0.801893
# X2	-0.213363	-0.135087
```
I have also implemented some very basic tests of the simultaneous CIs - the best I could come up with was to check if the CIs get wider. Do you have a better idea how to test for correctness?

apoorvalal:
Excellent idea; this is a lot more user friendly than the janky integer indices in my implementation. 


Re: validation, i think the `hdm`  r package has this procedure, so we could probably compare results. Will report back.

apoorvalal:
Turns out the `doubleml` package [implements the same multiplier bootstrap](https://docs.doubleml.org/stable/guide/sim_inf.html), so I just decided to cross-check the `confint` results with this. Seems hard to line up the RNG exactly, but the computed intervals are quite comparable (compare this to [16] in python output [here](https://docs.doubleml.org/stable/guide/sim_inf.html#multiplier-bootstrap-and-joint-confidence-intervals))

```python
import numpy as np
import pandas as pd
from pyfixest.estimation import feols
import doubleml as dml
from sklearn.base import clone
from sklearn.linear_model import LinearRegression
# %%
# Simulate data
np.random.seed(1234)
n_obs = 500
n_vars = 100
X = np.random.normal(size=(n_obs, n_vars))
theta = np.array([3., 3., 3.])
y = np.dot(X[:, :3], theta) + np.random.standard_normal(size=(n_obs,))
dml_data = dml.DoubleMLData.from_arrays(X[:, 10:], y, X[:, :10])
learner = LinearRegression(); ml_l = clone(learner); ml_m = clone(learner)
dml_plr = dml.DoubleMLPLR(dml_data, ml_l, ml_m)

# %%
dml_res = dml_plr.fit().bootstrap(n_rep_boot = 10_000).confint(joint=True)
dml_res
# %%
df = pd.DataFrame(np.c_[y, X], columns = ['y'] + ['X_'+str(x) for x in range(n_vars)])
m = feols(f"y ~ -1 + {'+'.join(['X_'+str(x) for x in range(n_vars)])}", df)
m.confint(keep = 'X_.$', nboot = 10_000, joint_cis = True)
# %%
```


| |0.025%|0.975%|
|:----|:----|:----|
|X_0|2.780117|3.036934|
|X_1|2.817083|3.084475|
|X_2|2.864678|3.118481|
|X_3|-0.157368|0.098401|
|X_4|-0.104548|0.153656|
|X_5|-0.167871|0.091990|
|X_6|-0.155494|0.099452|
|X_7|-0.068445|0.201500|
|X_8|-0.097808|0.170749|
|X_9|-0.141763|0.116197|

vs pointwise intervals, which are naturally narrower
| |0.025%|0.975%|
|:----|:----|:----|
|X_0|2.818857|2.998195|
|X_1|2.857418|3.044140|
|X_2|2.902963|3.080196|
|X_3|-0.118787|0.059820|
|X_4|-0.065599|0.114707|
|X_5|-0.128672|0.052791|
|X_6|-0.117037|0.060994|
|X_7|-0.027725|0.160780|
|X_8|-0.057298|0.130238|
|X_9|-0.102851|0.077285|

apoorvalal:
A potential bug I noticed in `_select_order_coefs` is that all inputs to `keep` are interpreted as regex rather than string, which leads to unexpected behavior. In the above example, I first tried 

```python
keep = ['X_'+str(x) for x in range(10)]
m.confint(keep = keep, nboot = 10_000, joint_cis = True, seed = 42)
```

which should have selected `X0, ..., X9`. However, since all covariates contain X and numbers (which is all `_select_order_coefs` cares about since it calls `re.findall`) the output to the above line is 

|0.025%|0.975%|
|:----|:----|
|X_0|2.749130|3.067921|
|X_1|2.784821|3.116738|
|X_10|2.834055|3.149104|
|X_11|-0.188229|0.129262|
|X_12|-0.135702|0.184810|
|...|...|...|
|X_95|-0.162518|0.149868|
|X_96|-0.068570|0.254539|
|X_97|-0.076739|0.247851|
|X_98|-0.263077|0.057014|
|X_99|-0.136747|0.168490|

(100 X 2) - i.e. all coefficients, which is not what was requested.

Wenzhi-Ding:
> A potential bug I noticed in `_select_order_coefs` is that all inputs to `keep` are interpreted as regex rather than string, which leads to unexpected behavior. In the above example, I first tried
> 
> ```python
> keep = ['X_'+str(x) for x in range(10)]
> m.confint(keep = keep, nboot = 10_000, joint_cis = True, seed = 42)
> ```
> 
> which should have selected `X0, ..., X9`. However, since all covariates contain X and numbers (which is all `_select_order_coefs` cares about since it calls `re.findall`) the output to the above line is
> 
> 0.025%	0.975%
> X_0	2.749130
> X_1	2.784821
> X_10	2.834055
> X_11	-0.188229
> X_12	-0.135702
> ...	...
> X_95	-0.162518
> X_96	-0.068570
> X_97	-0.076739
> X_98	-0.263077
> X_99	-0.136747
> (100 X 2) - i.e. all coefficients, which is not what was requested.

Thanks for pointing out this problem! Maybe we should introduce a switcher called `exact_match=True` so that it will use exact match instead of regex? Alternatively, we can specify if want to do exact match, write the expression as `keep = [f'^X_{x}$' for x in range(10)]`.

apoorvalal:
Yep, I was leaning towards the former since it is simpler (at least on the user-facing side; will require a couple of changes to `_select_order_coefs`).

Wenzhi-Ding:
No problem! I will implement it today.

s3alfisc:
Thanks both! I think adding `exact_match=True` is a great idea, I recall occasionally struggling with fixest's regex to pick a set of specific coefficients ... 

@apoorvalal: I've added `doubleML` as a dev dependency and added your tests against `doubleML`. Results are even closer than what you report above when I increase the sample size, which suggests to me that `doubleML` and `pyfixest` might use different small sample adjustments. Do you know if by default `doubleML` uses "iid" or "HC" standard errors? I suppose that HC would be the sensible default, but their [vignettes are not really clear](https://docs.doubleml.org/stable/search.html?q=heteroskedasticity) (at least not to me at a cursory glance). 

I have two other related questions: 

- Based on their [vignettes](https://docs.doubleml.org/stable/guide/se_confint.html), it looks to me that the the implemented method works for all m-estimators, i.e. we can also support Poisson and IV regression? 
- The method should generalize to computing simultaneous CIs for multiple dependent variables (e.g. one experiment & multiple KPIs) if we were to provide the "seemingly unrelated regression" variance covariance matrix as [here](https://github.com/kylebutts/vcovSUR)? This is also of interest for computing simultaneous CIs for the linear projections approach to staggered DiD's in which each "event study point estimate" stems from a different regression.

apoorvalal:
Nice, more aligned results with larger sample sizes makes sense to me. The fully general linear-score-based inference in DML is equivalent to `HCO` in OLS afaik.
And yes, the multiplier boostrap does work for all m-estimators (the vignette links to Chernozhukov et al's 2014 paper that proves it works for a very general class of problems). I think we can start with this for now and I'd like to think through simultaneous inference for the 2-step-GMM procedures for a separate PR.

s3alfisc:
>  And yes, the multiplier boostrap does work for all m-estimators (the vignette links to Chernozhukov et al's 2014 paper that proves it works for a very general class of problems). I think we can start with this for now and I'd like to think through simultaneous inference for the 2-step-GMM procedures for a separate PR.

Sounds great! Then I will merge as is after the CI tests pass. Thanks for getting this started and all your work on this!

pyfixest/FixestMulti.py
```diff
@@ -547,7 +547,7 @@ def confint(self) -> pd.Series:
             A pd.Series with coefficient names and confidence intervals.
             The key indicates which models the estimated statistic derives from.
         """
-        return self.tidy()[["2.5 %", "97.5 %"]]
+        return self.tidy()[["2.5%", "97.5%"]]
```
apoorvalal, line 550: this change breaks `coefplot` btw
image

PR #288, opened 2024-02-04
update poetry dependencies

NKeleher:
Updates to poetry dependencies.

I was unable to get poetry to install dependencies. Suggesting these modifications, including adding the quartodoc. 

Note that adding quartodoc, requires bumping the python requirement to >=3.9.

I'm no sure how the poetry-extras portion of the pyproject.toml was being used, but integrating the dependencies into the dev dependencies helped make it possible to run tests. Could be that adding a separate dependencies group for tests is the way to go.


s3alfisc (approved):

s3alfisc:
Glad that you got it to run!

> Note that adding quartodoc, requires bumping the python requirement to >=3.9.
Yes, I know ... that's why I was reluctant to add it do the `pyproject.toml`. As it was just me working on the project, it seemed fine, but now that I have asked others to contribute (and you have started doing so!) installing all dependencies should be easy, and giving up on Python 3.8 for this seems fair =) 

> I'm no sure how the poetry-extras portion of the pyproject.toml was being used, but integrating the dependencies into the dev dependencies helped make it possible to run tests. Could be that adding a separate dependencies group for tests is the way to go.

Basically the ordering was: packages that are essential to run pyfixest are in tool.poetry.dependencies, packages to run tests and build documentation are in tool.poetry.group.dev.dependencies, packages for optional features (running a wild bootstrap or polars support) were in poetry-extras. But I am happy with your proposed solution!

s3alfisc:
To make this pass CI, could you stop running the actions workflow on Python 3.8 [here](https://github.com/s3alfisc/pyfixest/blob/71bc54300d0e8d9a3c0da74bb60dc6d4f0529113/.github/workflows/ci-tests.yaml#L20)? Thanks! =)

NKeleher:
Yeah, it's unfortunate that quartodoc imposes the restriction on Python 3.8. And, unfortunately, even if the docs dependencies are set to optional [ref](https://python-poetry.org/docs/managing-dependencies/#optional-groups), the dependency resolution still fails :-(. Seems like this a worthwhile cost for having a streamlined docs build with quarto.

Thanks for reviewing!

s3alfisc:
I've lost track where we discussed our struggles to install under Python 3.12 - this might have been related with numba not supporting 3.12? This might have changed with the most recent numba release: [link](https://numba.discourse.group/t/ann-numba-0-59-0-and-llvmlite-0-42-0/2387)

NKeleher:
Well, good news is that with the next release, numba is now 0.59.0 in the poetry.lock: https://github.com/s3alfisc/pyfixest/blob/d753027a8b76e5e65fe9313eb55d43d06a9ee954/poetry.lock#L1831-L1836

Bumping numpy to work with Python 3.9 to 3.12 might be trivial at this point.

PR #303, opened 2024-02-07
rename docs/readme.md

NKeleher:
I believe that the reason  we only see the readme page at https://s3alfisc.github.io/pyfixest/ is because github pages thinks that the README is the full structure of the website.

Looking at https://github.com/s3alfisc/pyfixest/tree/gh-pages the README is the html page. However, other gh-pages branches do not have this html:

- https://github.com/NKeleher/statsframe/tree/gh-pages
image

- https://github.com/rstudio/pins-python/tree/gh-pages 
image


NKeleher:
I've renamed docs/readme.md to docs/pyfixest.qmd. 

Confirmed that the website renders locally. I also ran quartodoc to update the references pages.

s3alfisc:
The website now render properly 🥳Thanks for all your work on this, it will save me an incredible amount of time in the future!

PR #314, opened 2024-02-18
Implement the Causal Cluster Variance Estimator following Abadie et al

s3alfisc:


s3alfisc:
Basic implementation seems to work now: 

```py
# load autoreload extension 
%load_ext autoreload
%autoreload 2

from pyfixest.estimation import feols
from pyfixest.ccv import ccv
import pandas as pd
import numpy as np

df = pd.read_stata("C:/Users/alexa/Downloads/census2000_5pc.dta")
fit = feols("ln_earnings ~ college", vcov = {"CRV1": "state"}, data = df)
fit.ccv(treatment = "college", pk = 0.05)
```

image

while [the Stata version reports ](https://github.com/Daniel-Pailanir/TSCB-CCV)
image.

Issue #323, opened 2024-02-25
Replace `coef` argument in plotting functions with `drop` and `keep` arguments


s3alfisc:
We should make use of regex-based coefficient selection whenever possible. 

The `coefplot()` function and method currently support a `coefficients` argument - I think it makes sense to deprecate these in favor of `drop`, `keep`, and `exact_match`. 

As `_select_order_coefs` will be a function not only used for `etable` but also `confint` and `coefplot`, maybe it makes sense to move it from `summarize.py` to `dev_utils.py` (which is the script in which I collect non-user facing development utilities)? 

@Wenzhi-Ding, maybe something for you to pick up? =)

PR #325, opened 2024-02-25
dev: add exact_match to _select_order_coefs and etable

Wenzhi-Ding:
Echo issues #323 and #286, I add `exact_match` parameter to `_select_order_coefs()` function and move this function to `dev_utils.py` now. I also add some tests in `test_summary()` to verify `_select_order_coefs()` gives desirable results.

I haven't implemented them into `confint` and `coefplot` as mentioned in #323 yet.

### DIFF
--- a/pyfixest/dev_utils.py
+++ b/pyfixest/dev_utils.py
@@ -1,4 +1,5 @@
-from typing import Union
+import re
+from typing import Optional, Union
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


s3alfisc (approved):

s3alfisc:
Very cool, thank you @Wenzhi-Ding!

PR #327, opened 2024-02-26
Fix coefplot() bug after deleting whitespace before % for CIs in `tidy()`

s3alfisc:


s3alfisc:
Thanks for reporting @apoorvalal. Should be fixed now. Not sure how this slipped through my unit tests tbh...

PR #334, opened 2024-03-02
Fix ci error with Python 3.10

s3alfisc:


Wenzhi-Ding:
> So there error stems from the `did2s` tests. Likely not the pooled estimate as it runs in the `event_study()` tests.

Yea, when I trace the log, it seems to be something went wrong in `rpy2` under Python 3.10, associated with the `threading` module. It may be caused by some thread unproved code in `rpy2`.

s3alfisc:
Looks like the `tqdm` package is causing all these issues. As it's really only a nice to have, I'll drop it from the dependency list! @Wenzhi-Ding

Wenzhi-Ding:
The debugging process is so impressive! I learned a lot!

s3alfisc:
Actually, I am not sure this solved the problem, now the same error occurs under 3.9... 😞 So it is likely not just rpy2 + tqdm causing the issue?

Wenzhi-Ding:
From the log it seems it is still related with `rpy2`. Let me also have a try to see what exactly happens. Maybe it is caused by the import structure change...

image

s3alfisc:
Though now the second run worked ... I'd say I just merge it as is and release version `0.17.0`?

s3alfisc:
> Maybe it is caused by the import structure change...
I wouldn't think so, `rpy2` is just very fickle, I had quite some problems with it in the past & have thought about moving away from it for testing a couple of times. But so far things have usually fallen into place 😄

Wenzhi-Ding:
The most head-aching random error 😂 Anyway, it's great to see it work now. Please feel free to publish new release. 🚀

PR #341, opened 2024-03-04
Add `keep`/`drop` to `coefplot` and refactor code to improve maintainability

Wenzhi-Ding:
Hi Alex,

I now replace `coefficients` parameter by `keep`, `drop`, and `exact_match` for both `coefplot` and `iplot` as mentioned in #323.

In the mean time, I see there are some references to these two functions in `FixestMulti` and `Feols`. Changing a parameter in the original function requires changing all functions in these modules if calling them explicitly. So I also do a refactor to these codes use `functools.partial` so we don't need to bother syncing the changes all across codebase now.

Please feel free to comment or change the code!

Best regards,
Dave

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,7 @@ ignore = [
   "D104",   # missing docstring in public package
   "SIM110", # Use all instead of `for` loop
   "TRY003", # Avoid specifying long messages outside the exception class
+  "D205",   # 1 blank line required between summary line and description
 ]
 
 [tool.ruff.lint.per-file-ignores]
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -78,6 +78,7 @@ ignore = [
   "D104",   # missing docstring in public package
   "SIM110", # Use all instead of `for` loop
   "TRY003", # Avoid specifying long messages outside the exception class
+  "D205",   # 1 blank line required between summary line and description
```
s3alfisc, line 81: Yes, I also did not love that one ... 😄

s3alfisc (approved):

PR #347, opened 2024-03-11
fix: number of observations format

Wenzhi-Ding:
Hi Alex,

I am fixing a stupid thing I overlooked in formatting the table. Previously I mistakenly make the number of observations displayed in scientific notation if the number is large.

image

After the fix, it will be:

image


s3alfisc (approved):

s3alfisc:
Thanks for fixing this Dave!

PR #352, opened 2024-03-16
`FixestFormulaParser` cleanup / refactoring

s3alfisc:


s3alfisc:
To do: 

- [ ] i() interactions
- [ ] update did2s' use of `fixest_model_matrix`
- [ ] lint

s3alfisc:
To do's: 

- clean up `i_ref1` and `i_ref2` args
- fix and update tests
- add docstrings, fix linting

s3alfisc:
Hi Dave (@Wenzhi-Ding), from here on, I believe things will remain "mostly" stable. I have done the following things: 

- reworked `FixestFormulaParser`. Hopefully, it should now be much easier for you to follow what's happening inside. It only deparses multiple estimation syntax. 
- I now pass a `FixestFormula` object to `model_matrix_fixest` instead of a pure formula. The object contains information on the "second_stage" (the "last" estimation) and "first_stage" (in case of IV), the dependent variables, fixed effects, covariates, endog.vars and instruments. 
- I now use `formulaic.get_model_matrix()` to automatically create matrices for first and second stage and a weights vector. 
- I have updated the "i()" formula parsing. It is now possible to have more than one variable wrapped in `i()` syntax; and one can now set reference levels within the formula via `i(var1, var2, ref)` syntax - just as in fixest. I have deprecated the `i_ref1` and `i_ref2` arguments because they are no longer needed. I had a lot of fun matching regex patterns! 😄 

I still have to clean this up here and there, implement new tests, and fix some currently failing ones (but these are all minor).

PR #374, opened 2024-04-25
Add ruff checks on notebooks via nbqa

juanitorduz:
Here is a suggestion (because I saw it as a TODO) to use https://github.com/nbQA-dev/nbQA to run ruff checks on notebooks. I have used it in other projects, and it works very well! 

I had to make small changes to the notebooks to make it work. Most of the changes were mainly putting the imports on the first cell. 

Please check the changes and let me know if there is anything you want to revert (or even not do this change at all)

P.S. Also, as a tip ... avoid `from foo import *` (this is not recommended 😉 )

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,3 +22,10 @@ repos:
       - id: ruff
         args: ["--fix", "--output-format=full"]
       - id: ruff-format
+
+  - repo: https://github.com/nbQA-dev/nbQA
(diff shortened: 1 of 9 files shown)


s3alfisc (approved):
Looks excellent, much appreciated! 

Where did you spot `from package import *` syntax? I think I would only use this for `lets_plot`, mostly because there are so many different functions to export. But surely I shouldn't do this in the docs 😄

PR #376, opened 2024-04-25
Add colors to tests and cov report

juanitorduz:
Here are some suggestions to make the test reports more colorful and also get code coverage reports on the GitHub actions runs.

This is how it looks:
image

This is especially helpful when some test fails because you can easily locate it :)


juanitorduz:
The `codecov/project` does not like me 😅

s3alfisc:
> The codecov/project does not like me 😅

Do you know where this is coming from? Simply an "old" coverage report that I have forgotten to update?

PR #377, opened 2024-04-25
PNG -> SVG Badge

juanitorduz:
This makes the badge looks nicer :)

Compare https://img.shields.io/pypi/dm/pyfixest.svg vs https://img.shields.io/pypi/dm/pyfixest.png


s3alfisc:
Oh nice, I had been wondering why the badge was so blurred (and how to fix it) for a while😄

PR #397, opened 2024-04-27
Remove pre-commit from GitHub action

juanitorduz:
After https://github.com/s3alfisc/pyfixest/pull/375 we have seen the pre-commit app is working as expected and therefore we are checking the pre-commit twice. Removing this file because is unnecessary :)


s3alfisc (approved):

s3alfisc:
I have to figure out at some point what's going wrong with the codecov checks. Will merge this now nevertheless =)

PR #402, opened 2024-04-27
Pass `MyPy` Checks

s3alfisc:
- Refactor all code in `pyfixest/estimation` to pass `mypy` checks
- add a `mypy.ini` file
- add `mypy` to `pre-commit`


s3alfisc:
Introduced error for CRV estimation: 

```py
>               if len(self._ssc) == 0:
E               TypeError: object of type 'float' has no len()
```

s3alfisc:
Three errors remaining: 

pyfixest\utils\dev_utils.py:12: error: Cannot assign multiple types to name "DataFrameType" without an explicit "Type[...]" annotation  [misc]
pyfixest\estimation\feols_.py:1754: error: Incompatible return value type (got "tuple[ndarray[Any, Any], Any, Any | None, Any | None]", expected "tuple[ndarray[Any, Any], list[str], list[str], list[int]]")  [return-value]
pyfixest\estimation\FixestMulti_.py:394: error: Argument "_ssc_dict" to "add_fixest_multi_context" of "Feols" has incompatible type "dict[str, str] | None"; expected "dict[str, str | bool]"  [arg-type]

pyfixest/utils/dev_utils.py
```diff
@@ -3,13 +3,9 @@
...
 import pandas as pd
+import polars as pl
 
-try:
-    import polars as pl
-
-    DataFrameType = Union[pd.DataFrame, pl.DataFrame]
-except ImportError:
-    DataFrameType = pd.DataFrame
+DataFrameType = Union[pd.DataFrame, pl.DataFrame]
```
s3alfisc, line 8: Hi @juanitorduz, the only way to get rid of `mypy` errors in this section was the change above; the "old" version would lead to this error: 

"Cannot assign multiple types to name "DataFrameType" without an explicit "Type[...]" annotation  [misc]"

As a result, I moved `polars` from a dev-dependency to a "hard" dependency. This is somewhat overkill, as I only use polars in this very spot. 

Do you think I have to bite the bullet or do you see a way to make mypy work while still assigning DataFrameType dynamically?

juanitorduz: mmm ... this one I am not sure 🙈 . I think the key question is **do you need or want to support polars?**

I think this library will just keep growing so having it as a dependency should be ok (especially for large datasets). An alternative (for a future PR) is to work with new ideas around generic dataframe API. For example, this looks very interesting https://github.com/MarcoGorelli/narwhals

 Now that I am thinking about it a bit more I think `DataFrameType = Union[pd.DataFrame, pl.DataFrame]` could cause more problems as we want to use mypy to check if methods exist and as polars and pandas do not share the same methods this can create confusion. An alternative would be to be strict about then you are expecting polars and where you are expecting pandas (or see if https://github.com/MarcoGorelli/narwhals actually solves this)

s3alfisc: Thanks for your quick feedback! I mostly added `polars` support so that the package would work seamlessly with [pymarginaleffects ](https://github.com/vincentarelbundock/pymarginaleffects) (which runs on `polars`) and as a convenience for users. 

`polars` is practically not used at all in the codebase - one of the first things that happens is that if a user provides a `polars` DataFrame, it is directly converted to `pandas` ([see here](https://github.com/s3alfisc/pyfixest/blob/b8d793789b5b5192f59a34c4e7a4c00e1381cc8f/pyfixest/estimation/FixestMulti_.py#L58) and [here](https://github.com/s3alfisc/pyfixest/blob/b8d793789b5b5192f59a34c4e7a4c00e1381cc8f/pyfixest/utils/dev_utils.py#L15)). So I think there should be no ambiguity for `mypy` the way the code is set up at the moment.  

So in short, there is no real need to support polars except for `pymarginaleffects` and convenience for users.

I think I might just keep it as a dependency for now - we can still drop it later in case it might cause any problems in the future (unless you have major concerns 😄)

juanitorduz: Well, a fair question is whether it is worth the hassle to have polars. As you are converting it to pandas, that can be done by the user. So, in my opinion, if we are just parsing to pandas as soon as we get the data, we can get rid of polars (as there is no real gain). Maybe you can sleep over and see how it looks tomorrow hehe

s3alfisc: ... but I'll do it in a separate PR.

juanitorduz:
@s3alfisc Fantastic work with the MyPy integration! I know is hard but is better do it sooner than later :D

s3alfisc:
Without the help of my LLM coworker, I'd probably have another week ahead of me =) But there are still some fixes to apply for the `report` and `did` modules, which is also something for separate PRs. Thanks for tipping me over to get this started, the code quality has increased already (and I even found some proper bugs as well!).

PR #481, opened 2024-06-06
Panelview Function

s3alfisc:
Code by @apoorvalal .

I have slightly renamed function arguments, added docstrings & type hinting, and a very basic API test.

This also took me twice because I forgot to check out the dev branch 😅 

@apoorvalal , do you think there would be value in adding this function as a post-estimation method to the DiD estimators?

### DIFF
--- a/pyfixest/did/visualize.py
+++ b/pyfixest/did/visualize.py
@@ -1,0 +1,63 @@
+from typing import Optional
+
+import matplotlib.pyplot as plt
+import pandas as pd
+
+
+def panelview(
+    data: pd.DataFrame,
+    unit: str,
+    time: str,
+    treat: str,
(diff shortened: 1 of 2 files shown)


s3alfisc:
> I like the idea of having it free-standing because ideally you make this plot before running any regressions

I agree 👍 to clarify my suggestion, it would have been to keep it as a standalone function, but to also add it as a post-estimation method, purely as convenience for users who might want to check post hoc. But maybe we shouldn't incentivize poor research practices so happy to drop the post-estimation method idea.

s3alfisc:
On the point of setting a good example - it would be to add an example of the function at the beginning of the DiD notebook =)

apoorvalal:
prob close this

PR #513, opened 2024-06-17
Added solver to feols and created new test file for it

saidamir:
Hi I added a solver to feols class and created a test file for it. Once approved, I will add the solver to other classes as well, as I want to complete one class at this stage.


tests/test_solver.py
```diff
@@ -0,0 +1,61 @@
+import unittest
```
s3alfisc, line 1: Could you port the tests from `unittest` to `pytest`? E.g. the tests could look like this: 

```python
from dataclasses import dataclass
import numpy as np
import pytest

import pyfixest as pf
from pyfixest.estimation.feols_ import Feols

@pytest.fixture()
def data():

    data = pf.get_data()
    Y = data["Y"]
    X = data[["X1", "X2"]]
    Z = data[["Z1"]]

    return X, Y, Z, weights

def test_solver_equivalence(data):
    """
    Test the equivalence of different solvers for the feols class.
    This function initializes an object with test data and compares the results
    obtained from two different solvers: np.linalg.lstsq
    and np.linalg.solve. It asserts that
    the results are identical or very close within a tolerance.
    """
    X, Y, Z, weights = data
    obj = Feols(
        Y = Y,
        X = X,
        weights = weights,
        collin_tol = 1e-08,
        coefnames = ["X1", "X2"],
        weights_name = "weights",
        weights_type = None
    )

    # Use np.linalg.lstsq solver
    obj._solver = "np.linalg.lstsq"
    obj.get_fit()
    beta_hat_lstsq = obj._beta_hat.copy()

    # Use np.linalg.solve solver
    obj._solver = "np.linalg.solve"
    obj.get_fit()
    beta_hat_solve = obj._beta_hat.copy()

    # Assert that the results are identical or very close within a tolerance
    np.testing.assert_array_almost_equal(
        beta_hat_lstsq, beta_hat_solve, decimal=6, err_msg="Solvers' results differ"
    )
```

pyfixest/estimation/feols_.py
```diff
@@ -302,12 +303,16 @@ def get_fit(self) -> None:
...
         _Y = self._Y
         _Z = self._Z
-
+        _solver = self._solver
         self._tZX = _Z.T @ _X
         self._tZy = _Z.T @ _Y
 
         # self._tZXinv = np.linalg.inv(self._tZX)
-        self._beta_hat = np.linalg.solve(self._tZX, self._tZy).flatten()
+        if _solver == "np.linalg.lstsq":
```
s3alfisc, line 311: Looks good! For other solvers, maybe we can add a dedicated function, e.g. 

```python
def solve_ols(tZX, tZY, solver):
   
    if solver== "np.linalg.lstsq":
        return np.linalg.lstsq(self._tZX, self._tZy, rcond=None)
    elif solver == "np.linalg.solve": 
        return np.linalg.solve(self._tZX, self._tZy).flatten()
    else: 
        raise ValueError(f"Solver {_solver} not supported.")
        


```

saidamir:
@s3alfisc would you please review my pr at your free time? Thanks!

s3alfisc:
Hi @saidamir, sorry, I did not understand that this was already ready for review. This looks good, thank you! =) What next step do you have in mind? Will you add `solver` arguments to `Fepois` and `Feiv` or add arguments for the `feols()` and `fepois()` APIs?

pyfixest/estimation/fepois_.py
```diff
@@ -113,6 +115,32 @@ def __init__(
         self.deviance = None
         self._Xbeta = np.array([])
 
+    def solve_ols(self, tZX, tZY, solver):
```
s3alfisc, line 118: It is sufficient to define `solve_ols` as a method for `Feols` - as the other classes (`Feiv`, `Fepois`) are derived from `Feols`, the other classes will inherit all of `Feols` methods.

s3alfisc:
Hi Aziz (@saidamir) - if you move the `solve_ols` method to `Feols` and delete it in the `Feiv` and `Fepois` classes, then this looks good to merge (pending CI tests not failing, which I don't expect). 

@all-contributors please add @saidamir for code.

pyfixest/estimation/fepois_.py
s3alfisc: Done with [945f611](https://github.com/py-econometrics/pyfixest/pull/513/commits/945f611a151866f5ee3c187093e41cfb4a385f1f).

saidamir (commented):
Agreed.

saidamir (commented):
agreed

saidamir (commented):
agreed

saidamir:
When running test_solver.py the following error: FAILED tests/test_solver.py::test_solver_fepois - ValueError: operands could not be broadcast together with shapes (997,997) (997,998).
Screenshot 2024-07-01 at 5 25 07 PM
The proposed fix would involve changes to fepois_.py, git_fit() method as follows:
# Ensure mu is reshaped to (n_samples, 1) for correct broadcasting:
```diff
- WX = np.sqrt(mu) * X_resid
+ WX = np.sqrt(mu.reshape(-1, 1)) * X_resid
```
I am not sure if I can change the git fit method. What do you think?

s3alfisc:
I think the error stems from 

```
delta_new = self.solve_ols(XWX, XWZ, _solver) 
```
now evaluating to a "flat" array whereas in the old version, it was of dim N x 1. In the second run of the for loop, we then get this shape error. 

I think changing `delta_new` back to a N x 1 array should do the trick, e.g. 

```python
delta_new = self.solve_ols(XWX, XWZ, _solver).reshape((-1,1))
```

s3alfisc:
I think everything should be fixed now. I also added support for [scipy.sparse.linalg.lsqr](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.lsqr.html) and added a test =)

s3alfisc:
And it's merged! Congrats @saidamir and thanks for your help! =)

saidamir:
@s3alfisc would you recommend me the next issue? I have to admit, still struggling a bit with the basics, as I have to combine work, jobs search and this very exciting project, but I love it!

s3alfisc:
Of course! The immediate follow up would consist of adding the `solver` argument to the [feols](https://github.com/py-econometrics/pyfixest/blob/405ba73787af6bac793afb5bdbb34d2e7ad6cd53/pyfixest/estimation/estimation.py#L13) and [fepois](https://github.com/py-econometrics/pyfixest/blob/405ba73787af6bac793afb5bdbb34d2e7ad6cd53/pyfixest/estimation/estimation.py#L347) functions. There's no PR for this yet but I can write one tomorrow evening =)

PR #514, opened 2024-06-17
Add "ri" resampling method to Romano-Wolf procedure,

Jayhyung:
Hi, Alex. 

I made some changes to the function `rwolf` to allow resampling method when computing p-value using Romano-Wolf procedure. 

There are a couple of concerns that I have regarding this change. I would appreciate if you could review this, and give me the feedback(if these are okay to ignore, please link the pull request to the repo!)

1. The comment in the original code(`multicomp.py` in `estimation` folder) was somewhat tailored to the bootstrapping method. For example, the comment section in the function `_get_rwolf_pval` has this comment "`Compute Romano-Wolf adjusted p-values based on bootstrapped t-statistics.`". Please, confirm that I can use this function to compute the RW adjusted p-value with "`ri`" method as I did in the line 177. According to my understanding of this [paper](https://docs.iza.org/dp12845.pdf), the RW procedure should be exactly the same for both resampling methods except for the part that we are using different sampling method for computing t-statistics from resampling.

2. Do you think I should add a testing file to check on  whether the two different resampling method produces similar results? I'm not sure how to do this, so I would appreciate it if you could give me any example code for testing. I think comparing the adjusted p-values from both methods would be enough. 

Thanks!


s3alfisc:
Hi! This looks pretty good! 🎉 thank you =) 

Indeed, the rwolf stepwise function should work for both bootstrap and randomization inference p values. =) 

Regarding tests - one could argue that the feature does not need to be tested as both the rwolf function and the rotest function are tested independently. 

But it's likely best to have a test anyways (I've learned that I should test everything, if not, it will bite us) 😀 the best place for tests is in test_multcomp.py. You can start with the tolerance thresholds and iterations there (and please tell me if you feel we should be stricter with the test 😅)  

For testing, I think you're strategy is the right one - under a dgp where the RI and bootstrap assumptions hold (ie treatment being conditionally randomly assigned), you should expect rwolf to produce the same pvalue, no matter if run via a bootstrap or randomization inference.

Jayhyung:
Thanks for the comment! I will prepare the test code. Stay tuned! :)

Jayhyung:
I made a minor change to `multcomp.py` and added a test code to `test_multcomp.py.` The newly added test passed on my local machine. Please let me know if I need to do any extra works on this PR. Thanks!

pyfixest/estimation/multcomp.py
```diff
@@ -142,34 +150,48 @@ def rwolf(
...
+            model.ritest(
+                resampvar=param,
+                reps=reps,
+                type="randomization-t",
+                store_ritest_statistics=True,
+            )
+
+            t_stats[i] = model._ritest_sample_stat
+            boot_t_stats[:, i] = model._ritest_statistics
+        else:
```
s3alfisc, line 174: Could you add a test to [test_errors.py](https://github.com/py-econometrics/pyfixest/blob/0c84f012e2cf72762893ecf121e447e306b54e10/tests/test_errors.py#L181) that triggers the ValueError? This will make the codecov tool happy =)

pyfixest/estimation/multcomp.py
```diff
@@ -142,34 +150,48 @@ def rwolf(
...
+                return_bootstrapped_t_stats=True,
+                seed=seed,  # all S iterations require the same bootstrap samples, hence seed needs to be reset
+            )
+
+            t_stats[i] = wildboot_res_df["t value"]
+            boot_t_stats[:, i] = bootstrapped_t_stats
+
+        elif sampling_method == "ri":
+            model.ritest(
+                resampvar=param,
```
s3alfisc, line 166: Can you please provide a `rng` here? Each iteration of the procedure must run on the identical permutation of the data. I.e. within each loop, we need to apply the same rng. 

Unfortunately there is a small inconsistency in the `wildboottest` and `ritest` APIs - `ritest` requires a [np.RandomGenerator](https://numpy.org/doc/stable/reference/random/generator.html) while `wildboottest` is happy with a seed. 

E.g. you would have to add smth like

```python
rng = np.random.default_rng(seed) 
model.ritest(
          resampvar=param,
          ...
)
```

tests/test_multcomp.py
```diff
@@ -143,3 +143,47 @@ def test_stepwise_function():
...
+
+    # Calculate percentage difference in p-values
+    percent_diff = 100 * (wb_pval - ri_pval) / ri_pval
+
+    print(
+        f"Percentage difference in p-values (seed={seed}, reps={reps}): {percent_diff}"
+    )
+
+    # Assert that the percentage difference is within an acceptable range
+    assert (
```
s3alfisc, line 187: Just an fyi, no need to change this - you could also use tools from the `np.testing` module, e.g. [np.testing.assert_allclose()](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html).

s3alfisc:
Thanks @Jayhyung 🎉I added a few more small comments. Very nicely, the CI tests ran succesfully! So I think we will be able to merge this after the next iteration =)

pyfixest/estimation/multcomp.py
Jayhyung: Could you elaborate on this? In `test_error.py`, we have two functions related to `multcomp.py` : `test_rwolf_error` and `test_multcomp_errors`. `test_multcomp_errors` also triggers a ValueError, but are we adding a new test that triggers the error in the sampling method feature? I would appreciate it if you let me know whether the following code will be enough. Thanks!

```
def test_multcomp_sampling_errors():
    data = get_data().dropna()
    # Sampling method not supported in "rwolf"
    fit1 = feols("Y + Y2 ~ X1 | f1", data=data)
    with pytest.raises(ValueError):
        rwolf(fit1.to_list(), param="X1", reps=999, seed=92, sampling_method = "abc")
```

s3alfisc: Yes, that's exactly what I had I mind =)

Jayhyung:
Thanks for the comment. I modified the code, and added the test according to you suggestion. Please, let me know if anything extra should be done. Thanks!

s3alfisc (approved):

s3alfisc:
And merged! 🥳

PR #525, opened 2024-06-23
Add 1st stage regression in Feiv class

Jayhyung:
Hi Alex @s3alfisc! More details with this PR will follow soon.


s3alfisc:
Very cool! Looking forward to it =)

Jayhyung:
There are a few concerns that I have working on this issue.

1. Does `feiv` class in the codebase allow multiple endogenous variables? I'm asking this question because the IV test cases in the test code(line 97 ~ 117 in `test_vs_fixest.py`) only allow the single case. The first stage in my code only allows the single case, but if you would like to generalize the code that allows the multiple cases, I'm happy to work more on this!
2. I defined several attributes within `feiv` class related with the first stage regression(`_pi_hat` : estimated coefficient, `X_hat` : predicted values, `_v_hat` : residuals). Please, let me know if you want me to define more/less attributes related with the first stage. 
3. I would add a test code that evaluates whether the result(e.g. `_pi_hat`) of the first stage regression is the same with the result from `feols`. Do you think this is a reasonable way to test the 1st stage part of the `get_fit` method in `feiv` class? Also, where do i have to add the test code in `test_vs_fixest.py` exactly?

Thanks!

s3alfisc:
Hi @Jayhyung - sorry for my delayed response, I have been swamped for work over tethe last days :/ 

> Does feiv class in the codebase allow multiple endogenous independent variables? I'm asking this question because the IV test cases in the test code(line 97 ~ 117 in test_vs_fixest.py) only allow the single case. The first stage in my code only allows the single case, but if you would like to generalize the code that allows the multiple cases, I'm happy to work more on this!

Yes, only one endogenous variable is allowed. It would be quite cool to allow for more, but this would require to rework the `FixestFormulaParser` and it might not be a small endeavor. So for now, I would recommend to maybe keep this as a separate issue? 

> I defined several attributes within feiv class related with the first stage regression(_pi_hat : estimated coefficient, X_hat : predicted values, _v_hat : residuals). Please, let me know if you want me to define more/less attributes related with the first stage.

For now, I think that coefficient, predicted values and residuals is all that we need, right? So this sounds good to me =) 

> I would add a test code that evaluates whether the result(e.g. _pi_hat) of the first stage regression is the same with the result from feols. Do you think this is a reasonable way to test the 1st stage part of the get_fit method in feiv class? Also, where do i have to add the test code in test_vs_fixest.py exactly?

Sounds like a good strategy =) I think that you could just create a new test file, e.g. `test_iv.py`?

Jayhyung:
Thanks for the feedback! I will work on my next iteration. Stay tuned. :)

Jayhyung:
Just added the error testing file! Please let me know if anything extra should be done.

pyfixest/estimation/feiv_.py
```diff
@@ -133,6 +139,28 @@ def get_fit(self) -> None:
         _Z = self._Z
         _Y = self._Y
 
+        #  Start First Stage
+
+        model1 = Feols(
```
s3alfisc, line 144: I think it correct that the endogenous variable is always the first one in `_X`. But it would likely be safer to use the [endogvar](https://github.com/py-econometrics/pyfixest/blob/b30512d63fc7186e0d5b6058b094fd0a0b7d6cc1/pyfixest/estimation/FixestMulti_.py#L232) created via the `model_matrix_fixest` function? 

This would require to add an `endogvar` argument to `Feiv.__init__` and to pass it to `Feiv.get_fit()`. 

We would also have to demean it in FixestMulti.

pyfixest/estimation/feiv_.py
```diff
@@ -133,6 +139,28 @@ def get_fit(self) -> None:
...
+            Y=_X[:, -1].reshape(-1, 1),
+            X=_Z,
+            weights=np.ones(self._weights.shape[0]).reshape(-1, 1),
```
s3alfisc, line 147: Here we do not want to use `self._weights` because the weights are already applied to `_X` and `_Z`, right? Nice catch =)

pyfixest/estimation/feiv_.py
```diff
@@ -142,13 +170,17 @@ def get_fit(self) -> None:
...
+        # Estimate coefficients (beta_hat)
         self._beta_hat = np.linalg.solve(A, B).flatten()
 
+        # Predicted values and residuals
         self._Y_hat_link = self._X @ self._beta_hat
         self._u_hat = self._Y.flatten() - self._Y_hat_link.flatten()
 
+        # Compute scores and hessian
         self._scores = self._Z * self._u_hat[:, None]
-        self._hessian = self._Z.transpose() @ self._Z
```
s3alfisc, line 151: Yup, `.transpose()` is a bad habit of mine... 😅 thx!

pyfixest/estimation/feiv_.py
Jayhyung: I think your practice is more sensible as it encompasses more general cases. I will take this into account in the next iteration!

pyfixest/estimation/feiv_.py
Jayhyung: Yes, this is exactly what I had in mind!

Jayhyung:
I added `endogvar` as you suggested to the next iteration. Let me know if we need anything extra to be done!

pyfixest/estimation/feiv_.py
```diff
@@ -31,7 +33,8 @@ class Feiv(Feols):
         Names of the coefficients of Z.
     collin_tol : float
         Tolerance for collinearity check.
-    weights_name : Optional[str]
+    weights_name : Op    endgvar : np.ndarray
```
s3alfisc, line 36: Small formatting issue here =)

Jayhyung:
Thanks! Just fixed the formatting issue. Let me know if anything comes up!

s3alfisc:
Perfect. It's merged =) thanks @Jayhyung!

PR #537, opened 2024-07-05
Clarify fixef() value error message

b-knight:
I had this fail on me when I passed PyFixest the following fixed effect : 

C(SHOPPER_PLATFORM)[T.[
  "ios"
]]

I imagine having escape sequences in your fixed effect variable is poor form. 
Hopefully this can save someone some troubleshooting time.

### DIFF
--- a/pyfixest/estimation/feols_.py
+++ b/pyfixest/estimation/feols_.py
@@ -1336,7 +1336,9 @@ class Feols:
                 level = matches.group(2)
             else:
                 raise ValueError(
-                    "Something went wrong with the regex. Please open a PR in the github repo!"
+                    f"The fixef() method failed after regex encountered the following value: {col}."
+                    + " This may due to the presence of line separation and/or escape sequences within the string."
+                    + " If so, consider recoding the underlying string. Otherwise, please open a PR in the github repo!"
                 )
 
             # check if res already has a key variable


s3alfisc:
Hi Ben, thanks for reporting this! And sorry to hear that you lost some time on  this 😞 Anyways I'm more than happy to accept this PR as it is clearly an improvement relative to the status quo! 

I suppose the error lies with the regex here https://github.com/py-econometrics/pyfixest/blob/dee5109965b87368e28521ffc87710d289955d4f/pyfixest/estimation/feols_.py#L1333. I wonder if we could re-write the code to make it more robust? The key functionality of the code section is to find the "fixed effect name" and level in the formulaic string. 

The formulaic encoding for fixed effects is `f"T[{fixef_name}]{fixef_level}"`, and I try to select "fixef_name" and "fixef_level". I guess the regex is simply failing in your case and we should be able to accommodate it by adjusting the regex. Could you maybe add a reproducible example so that I could run some checks? 

For example, this here seems to work for me: 

```python
import pyfixest as pf 
import numpy as np

data = pf.get_data()
data["f4"] = np.random.choice(["A[", "B-", "&C"], size = data.shape[0])

fit = pf.feols("Y ~ X1 | f4", data = data)
fit.tidy()
fit.fixef()
# {'C(f4)': {'&C': 0.9304501248551359,
#  'A[': 0.8679081078089161,
#  'B-': 0.9582724997599982}}
```

b-knight:
Hi Alex! I was able to reproduce the problem by adding a next line character (\n) to the string, e.g. 
`import pyfixest as pf 
import numpy as np

data = pf.get_data()
data["f4"] = np.random.choice(["[\n \"web\", \"ios\" \n]", "[\n \"web\", \"android\" \n]"], size = data.shape[0])

fit = pf.feols("Y ~ X1 | f4", data = data)
fit.tidy()
fit.fixef()`

This is a failure to do basic data sanitation on my part - not sure if there is an easy fix.

s3alfisc:
The problem indeed occurs because the regex pattern fails as `formulaic` seems to apply a different naming convention:

image. 

I think this regex pattern matching should work quite generally: 

```python
import re

def extract_variable_level(x: str):


    c_pattern = r"C\((.+?)\)"
    t_pattern = r'\[T\.(.*\])'

    c_match = re.search(c_pattern, x)  
    t_match = re.search(t_pattern, x, re.DOTALL)

    if not c_match or not t_match:
        raise ValueError(
            f"The fixef() method failed after regex encountered the following value: {col}."
            + " This may due to the presence of line separation and/or escape sequences within the string."
            + " If so, consider recoding the underlying string. Otherwise, please open a PR in the github repo!"
        )

    variable = c_match.group(1)
    level = t_match.group(1)

    return variable, level


cols = [
    'C(f4)[T.[\n "web", "ios" \n]]',
    'C(f1)[T.27.0]'
]

for col in cols: 
    try:
        var, lvl = extract_variable_level(col)
        print(f"Variable: {var}, Level: {lvl}")
    except ValueError as e:
        print(e)

# Variable: f4, Level: [
#  "web", "ios" 
# ]]
# Variable: f1, Level: 27.0]
```

b-knight:
Oh wow. I was wondering where those line separation characters came from. Thank you so much for looking into this!

PR #542, opened 2024-07-08
Add a `fixef_maxiter` function argument to feols(), fepois

### LINKED ISSUE #539: Helping demean() Fail Gracefully - Make 'maxiter' Accessible from feols()
I've encountered instances when [demean_.demean()](https://github.com/py-econometrics/pyfixest/blob/master/pyfixest/estimation/demean_.py#L187-L267) runs endlessly - probably because I passed it a poorly structured data shape. I would like to be able to call a service that uses PyFixest, so calling the package on problematic data is a distinct possibility. 

To this end, it would be great if there were ways to help the demean() method fail gracefully. Having some sort of mechanism for timing out could be handy, but the simplest approach that I can think of off hand is to make the 'maxiter' argument accessible from the feols() method in a manner similar to how the 'fixef_tol' argument is accessible to the user.

s3alfisc:
- This PR adds a `fixef_maxiter` function argument to both the `feols()` and `fepois()` API and adds basic unit tests. 
- Merging this PR should close #539 .

### DIFF
--- a/pyfixest/estimation/FixestMulti_.py
+++ b/pyfixest/estimation/FixestMulti_.py
@@ -25,6 +25,7 @@ class FixestMulti:
         copy_data: bool,
         store_data: bool,
         fixef_tol: float,
+        fixef_maxiter: int,
         weights_type: str,
     ) -> None:
         """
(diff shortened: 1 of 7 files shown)


s3alfisc:
A similar feature has been merged a long time ago.

PR #544, opened 2024-07-09
Improved robustness of feols_ regex.

b-knight:
This is a follow-up to the [PR Clarify fixef() value error message](https://github.com/py-econometrics/pyfixest/pull/537). It looked like the formulaic encoding was adding line breaks (\n) under certain circumstances which the feols_ regex was struggling with. [Alex did the heavy lifting](https://github.com/py-econometrics/pyfixest/pull/537#issuecomment-2212392559) in refactoring the regex, and I just made things pretty and spun up a PR.

### DIFF
--- a/pyfixest/estimation/feols_.py
+++ b/pyfixest/estimation/feols_.py
@@ -1288,6 +1288,38 @@ class Feols:
         -------
         None
         """
+
+        def extract_variable_level(fe_string: str):
+            """
+            Extract the variable and level from a given string.
+
+            Parameters
+            ----------
+            fe_string: str
+                The string encapsulating the fixed effect factor variable and level.
+
+            Returns
+            -------
+            tuple
+                A tuple containing the extracted variable and level for the fixed
+                effect.
+            """
+            c_pattern = r"C\((.+?)\)"
+            t_pattern = r"\[T\.(.*\])"
+            c_match = re.search(c_pattern, fe_string)
+            t_match = re.search(t_pattern, fe_string, re.DOTALL)
+
+            if not c_match or not t_match:
+                raise ValueError(
+                    f"feols() failed after regex encountered the following value as a fixed effect:\n {fe_string}."
+                    + "\nThis may due to the presence of line separation and/or escape sequences within the string."
+                    + " If so, consider recoding the underlying string. Otherwise, please open a PR in the github repo!"
+                )
+
+            variable = c_match.group(1)
+            level = t_match.group(1)
+            return variable, level
+
         _has_fixef = self._has_fixef
         _is_iv = self._is_iv
         _method = self._method
@@ -1330,14 +1362,10 @@ class Feols:
 
         res: dict[str, dict[str, float]] = {}
         for i, col in enumerate(cols):
-            matches = re.match(r"(.+?)\[T\.(.+?)\]", col)
-            if matches:
-                variable = matches.group(1)
-                level = matches.group(2)
-            else:
-                raise ValueError(
-                    "Something went wrong with the regex. Please open a PR in the github repo!"
-                )
+            try:
+                variable, level = extract_variable_level(col)
+            except ValueError as e:
+                print(e)
 
             # check if res already has a key variable
             if variable not in res:
@@ -1529,12 +1557,12 @@ class Feols:
         Return a tidy pd.DataFrame with the point estimates, standard errors,
         t-statistics, and p-values.
 
-        Parameters 
+        Parameters
         ----------
         alpha: Optional[float]
-            The significance level for the confidence intervals. If None, 
-            computes a 95% confidence interval (`alpha = 0.05`). 
-        
+            The significance level for the confidence intervals. If None,
+            computes a 95% confidence interval (`alpha = 0.05`).
+
         Returns
         -------
         tidy_df : pd.DataFrame
@@ -1825,10 +1853,12 @@ class Feols:
         sample_stat = sample_tstat if type == "randomization-t" else sample_coef
 
         if clustervar_arr is not None and np.any(np.isnan(clustervar_arr)):
-            raise ValueError("""
+            raise ValueError(
+                """
             The cluster variable contains missing values. This is not allowed
             for randomization inference via `ritest()`.
-            """)
+            """
+            )
 
         if type not in ["randomization-t", "randomization-c"]:
             raise ValueError("type must be 'randomization-t' or 'randomization-c.")
@@ -1944,11 +1974,13 @@ class Feols:
         Inference Statistics.
         """
         if not hasattr(self, "_ritest_statistics"):
-            raise ValueError("""
+            raise ValueError(
+                """
                             The randomization inference statistics have not been stored
                             in the model object. Please set `store_ritest_statistics=True`
                             when calling `ritest()`
-                            """)
+                            """
+            )
 
         ri_stats = self._ritest_statistics
         sample_stat = self._ritest_sample_stat
--- a/pyfixest/estimation/feols_.py
+++ b/pyfixest/estimation/feols_.py
@@ -1305,7 +1305,7 @@ class Feols:
                 effect.
             """
             c_pattern = r"C\((.+?)\)"
-            t_pattern = r"\[T\.(.*\])"
+            t_pattern = r"\[T\.(.*?)\]"
             c_match = re.search(c_pattern, fe_string)
             t_match = re.search(t_pattern, fe_string, re.DOTALL)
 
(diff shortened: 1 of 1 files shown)


b-knight:
@s3alfisc Did I break something here, or is this just an idiosyncrasy of the test suite?

s3alfisc:
It's not you but the code I gave to you. Fixing it now!

b-knight:
Do I need to add unit tests to satisfy Codecov?

b-knight:
If it is ok with you, [this](https://github.com/py-econometrics/pyfixest/pull/544#issuecomment-2220416756) could be an opportunity for me to get some unit-testing experience. I'll try my hand at resolving this tonight

s3alfisc:
Sounds great! Of course that's fine =) looking forward to the update!

Jayhyung:
Hi Ben! We actually talked about what might be the best way to put testing code, so I'm posting my explanation on this thread. I'm using  There are two parts that raise issues in codecov in your code. 

> pyfixest/estimation/feols_.py#L1313

> pyfixest/estimation/feols_.py#L1367-L1368

These two parts were raised as an issue because the error triggers in these codes were not tested. So you need to add testing scripts to test whether these error triggers works right when they are triggered. I can demonstrate how to add the testing scripts using the first error issue(L1313). 


First, I made your method `extract_variable_level()` standalone, because it was inside the method `fixef()` (I did this to make the test script simple as possible, because you need to call  `fixef()` when testing `extract_variable_level()` if we don't do this). So the codes related with this part in `feols_.py` will read as the followings;

```python

    def fixef(self) -> dict[str, dict[str, float]]:
        """
        Compute the coefficients of (swept out) fixed effects for a regression model.

        This method creates the following attributes:
        - `alphaDF` (pd.DataFrame): A DataFrame with the estimated fixed effects.
        - `sumFE` (np.array): An array with the sum of fixed effects for each
        observation (i = 1, ..., N).

        Returns
        -------
        None
        """
        _has_fixef = self._has_fixef
        _is_iv = self._is_iv
        _method = self._method
        _fml = self._fml
        _data = self._data

        if not _has_fixef:
            raise ValueError("The regression model does not have fixed effects.")

        if _is_iv:
            raise NotImplementedError(
                "The fixef() method is currently not supported for IV models."
            )

        # fixef_vars = self._fixef.split("+")[0]

        depvars, rhs = _fml.split("~")
        covars, fixef_vars = rhs.split("|")

        fixef_vars_list = fixef_vars.split("+")
        fixef_vars_C = [f"C({x})" for x in fixef_vars_list]
        fixef_fml = "+".join(fixef_vars_C)

        fml_linear = f"{depvars} ~ {covars}"
        Y, X = Formula(fml_linear).get_model_matrix(_data, output="pandas")
        if self._X_is_empty:
            Y = Y.to_numpy()
            uhat = Y

        else:
            X = X[self._coefnames]  # drop intercept, potentially multicollinear vars
            Y = Y.to_numpy().flatten().astype(np.float64)
            X = X.to_numpy()
            uhat = csr_matrix(Y - X @ self._beta_hat).transpose()

        D2 = Formula("-1+" + fixef_fml).get_model_matrix(_data, output="sparse")
        cols = D2.model_spec.column_names

        alpha = spsolve(D2.transpose() @ D2, D2.transpose() @ uhat)

        res: dict[str, dict[str, float]] = {}
        for i, col in enumerate(cols):
            matches = re.match(r"(.+?)\[T\.(.+?)\]", col)
            if matches:
                variable = matches.group(1)
                level = matches.group(2)
            else:
                raise ValueError(
                    "Something went wrong with the regex. Please open a PR in the github repo!"
                )

            # check if res already has a key variable
            if variable not in res:
                res[variable] = dict()
                res[variable][level] = alpha[i]
                continue
            else:
                if level not in res[variable]:
                    res[variable][level] = alpha[i]

        self._fixef_dict = res
        self._alpha = alpha
        self._sumFE = D2.dot(alpha)

        return self._fixef_dict
    

    def extract_variable_level(self, fe_string: str):
        """
        Extract the variable and level from a given string.
        Parameters
        ----------
        fe_string: str
            The string encapsulating the fixed effect factor variable and level.
        Returns
        -------
        tuple
            A tuple containing the extracted variable and level for the fixed
            effect.
        """
        c_pattern = r"C\((.+?)\)"
        t_pattern = r"\[T\.(.*\])"
        c_match = re.search(c_pattern, fe_string)
        t_match = re.search(t_pattern, fe_string, re.DOTALL)

        if not c_match or not t_match:
            raise ValueError(
                f"feols() failed after regex encountered the following value as a fixed effect:\n {fe_string}."
                + "\nThis may due to the presence of line separation and/or escape sequences within the string."
                + " If so, consider recoding the underlying string. Otherwise, please open a PR in the github repo!"
            )

        variable = c_match.group(1)
        level = t_match.group(1)
        return variable, level




```

The second step is to deploy test scripts that test the error flag `pyfixest/estimation/feols_.py#L1313` in  `extract_variable_level()` method by putting the following script to `test_errors.py`


```python

def test_extract_variable_level_invalid_no_C():
    data = pd.read_csv("pyfixest/did/data/df_het.csv")
    data = data.iloc[1:3000]
    fml = "dep_var ~ treat"
    fit = feols(fml, data, vcov={"CRV1": "year"}, ssc=ssc(adj=False))

    fe_string = "variable[T.level]"
    with pytest.raises(ValueError) as excinfo:
        fit.extract_variable_level(fe_string)
    assert "feols() failed after regex encountered the following value as a fixed effect:" in str(excinfo.value)


```

So this

s3alfisc:
Thanks @Jayhyung! I would 100% accept your suggested changes =) 

If we make `extract_variable_level` a standalone function, we could add an even more straightforward test by simply checking the function directly. I think the software engineers in my team would call this a "[unit test](https://en.wikipedia.org/wiki/Unit_testing)", whereas what @Jayhyung has suggest might be called an "[integration test](https://en.wikipedia.org/wiki/Integration_testing)". Generally `PyFixest` does the latter by directly calling `feols()` or `fepois()` to test most behavior - here I follow advise from the [testing chapter in Hadley Wickham's book ](https://r-pkgs.org/testing-design.html#what-to-test) where he writes 

> Focus on testing the external interface to your functions - if you test the internal interface, then it’s harder to change the implementation in the future because as well as modifying the code, you’ll also need to update all the tests.

But here I think it would be ok to direclty test the function. E.g. we could move `test_extract_variable_level` to `pyfixest.utils.dev_utils ` and then add the following to `test_predict_resid_fixef.py`:

```python
from pyfixest.utils.dev_utils import _extract_variable_level

def test_extract_variable_level():

    var = '[\n "web", "ios" \n]'
    assert _extract_variable_level(var) == ("f4", '.[\n "web", "ios" \n]')

    var = 'C(f3)[T.1.0]'
    assert _extract_variable_level(var) == ("f3", "1.0")

    var = 'C(f4)[T.1]'
    assert _extract_variable_level(var) == ("f4", "1")

```

b-knight:
@s3alfisc Are you confident that the regex is what you want? The method 
`def _extract_variable_level(fe_string: str):
    import re
    c_pattern = r"C\((.+?)\)"
    t_pattern = r"\[T\.(.*\])"
    c_match = re.search(c_pattern, fe_string)
    t_match = re.search(t_pattern, fe_string, re.DOTALL)

    if not c_match or not t_match:
        raise ValueError(
            f"feols() failed after regex encountered the following value as a fixed effect:\n {fe_string}."
            + "\nThis may due to the presence of line separation and/or escape sequences within the string."
            + " If so, consider recoding the underlying string. Otherwise, please open a PR in the github repo!"
        )

    variable = c_match.group(1)
    level = t_match.group(1)
    return variable, level`
...yields the following when I pass it **str = "C(f3)[T.1.0]"** (notice the right hand side bracket) 
Screenshot from 2024-07-13 18-51-25

s3alfisc:
Yep, you're right, the function above needs a small update. We need to change 

```py
    t_pattern = r"\[T\.(.*\])"
```
to 
```py
    t_pattern = r"\[T\.(.*?)\]"
```
And then things work! 

```py
def _extract_variable_level(fe_string: str):
    import re
    c_pattern = r"C\((.+?)\)"
    t_pattern = r"\[T\.(.*?)\]"
    c_match = re.search(c_pattern, fe_string)
    t_match = re.search(t_pattern, fe_string, re.DOTALL)

    if not c_match or not t_match:
        raise ValueError(
            f"feols() failed after regex encountered the following value as a fixed effect:\n {fe_string}."
            + "\nThis may due to the presence of line separation and/or escape sequences within the string."
            + " If so, consider recoding the underlying string. Otherwise, please open a PR in the github repo!"
        )

    variable = c_match.group(1)
    level = t_match.group(1)
    return variable, level

var = 'C(f4)[T.\n "web", "ios" \n]'
_extract_variable_level(var)
# ('f4', '\n "web", "ios" \n')
```

PR #546, opened 2024-07-11
lsqr() for fixed effects solver

### LINKED ISSUE #469: fixef: Implement Iterative Least Squares Solver
## Context

Currently, pyfixest can compute fixed effect estimates in the [fixef()](https://github.com/py-econometrics/pyfixest/blob/734a21c34a56a88ba12b0fa26cd3ed0e8c24bd51/pyfixest/estimation/feols_.py#L1206) method by using a sparse solver. 

A potentially faster alternative is to implement an iterative solver. See section 3 in the [lfe paper](https://cran.r-project.org/web/packages/lfe/vignettes/lfehow.pdf).

greenguy33:
Closes #469


greenguy33:
Looks like method `_matmul_dispatch()` in `scipy.sparse` is throwing an error based on this change:

```
        if issparse(other):
            if self.shape[-1] != other.shape[0]:
>               raise ValueError('dimension mismatch')
E               ValueError: dimension mismatch

../../../.cache/pypoetry/virtualenvs/pyfixest-va_uSH9_-py3.9/lib/python3.9/site-packages/scipy/sparse/_base.py:603: ValueError
```
I'll take a look at what's going asap.

s3alfisc:
I think lsqr might return a tuple? https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.lsqr.html so we'd have to pick the first return element.

pyfixest/estimation/feols_.py
```diff
@@ -1384,8 +1384,8 @@ def fixef(self) -> dict[str, dict[str, float]]:
         D2 = Formula("-1+" + fixef_fml).get_model_matrix(_data, output="sparse")
         cols = D2.model_spec.column_names
```
s3alfisc, line 1385: `lsqr` requires `uhat` to be a 1D array. You can solve many of the test error by changing 

```py
uhat = csr_matrix(Y - X @ self._beta_hat).transpose()
```
to 
```py
uhat = (Y - X @ self._beta_hat).flatten()
```

s3alfisc:
Another option to fix the failing tests is to tighten the convergence criteria for the iterative `lsqr` solver. I.e. by setting `atol=1e-10` and `btol=1e-10`, all tests pass: 

```python
        alpha = lsqr(D2, uhat, atol = 1e-10, btol = 1e-10)[0]
```

Note that `scikit-learns` `LinearRegession` class also builds on the `lsqr` solver, and they are satisfied with the default tolerances for `atol` and `btol` of 1e-06. 

I think I would suggest the following: we keep the 1e-06 defaults as our defaults, but add function arguments `atol` and `btol` to the `.predict()` and `.fixef()` methods so that users can specify their desired numerical accuracy. 

What do you think Hayden (@greenguy33)? Any feedback from a user's perspective on this, @b-knight?

greenguy33:
That sounds good to me. I am mostly offline this weekend but can definitely make this change in the coming week. I guess it will require some changes in the tests as well, to supply the passing values of 1e-10?

greenguy33:
@s3alfisc Looks like tests are passing now. However in order to get that to happen I changed the tolerance values for np.testing.assert_allclose in two test files. I'm not sure if these tests were passing before because it doesn't appear that they are running the lsqr solver as Supplying tolerance values of atol = 1e-05, rtol = 1e-05 seemed to work, and smaller values tend to fail

s3alfisc (approved):

s3alfisc:
Just merged, thank you @greenguy33!

PR #554, opened 2024-07-15

Jayhyung:
Hi @s3alfisc!

I will add more details to this PR soon!

Thanks,


Jayhyung:
Hmm, this may be the same issue that @b-knight raised in Discord? I re-added `.coverage` to the PR, but this does not resolve the issue ;)

PR #529, opened 2024-06-30
Vcov

s3alfisc:
1. Updates small sample adjustments to use (N-1) / (N-K) for heteroskedastic errors when `adj = True`, `1` if `adj=False`, which is in line with the `fixest` docs. 
2. Reorganizes unit tests to skip non-required tests in `test_vs_fixest.py`.
3. Adds dedicated tests for exact equality for vcov matrices in `test_vcov.py` and also in `test_vs_fixest.py`.

Note that this is labelled a "breaking change" as inference results will slightly deviate. But for medium N and small k, differences will be negligible. 

Example: 

```python
%load_ext autoreload
%autoreload 2

import numpy as np
import rpy2.robjects as ro
from rpy2.robjects import pandas2ri
from rpy2.robjects.packages import importr

import pyfixest as pf
from pyfixest.utils.utils import ssc
import pandas as pd

pandas2ri.activate()

fixest = importr("fixest")
stats = importr("stats")

data = pf.get_data(N = 100, model = "Feols")

cluster_adj = False
fml = "Y ~ X1 + X2 | f1"
vcov = "hetero"

py_mod = pf.feols(
    fml, data=data, vcov=vcov, ssc=ssc(adj=adj, cluster_adj=cluster_adj)
)
r_mod = fixest.feols(
    ro.Formula(fml),
    data=data,
    vcov=vcov,
    ssc=fixest.ssc(adj, "none", cluster_adj, "min", "min", False),
)

py_mod_vcov = py_mod._vcov
r_mod_vcov = stats.vcov(r_mod)

py_mod_vcov - r_mod_vcov
# array([[6.93889390e-18, 1.08420217e-19],
#        [3.25260652e-19, 0.00000000e+00]])

adj = False
cluster_adj = False
fml = "Y ~ X1 + X2 | f1"
vcov = "hetero"

py_mod = pf.feols(
    fml, data=data, vcov=vcov, ssc=ssc(adj=adj, cluster_adj=cluster_adj)
)
r_mod = fixest.feols(
    ro.Formula(fml),
    data=data,
    vcov=vcov,
    ssc=fixest.ssc(adj, "none", cluster_adj, "min", "min", False),
)

py_mod_vcov = py_mod._vcov
r_mod_vcov = stats.vcov(r_mod)

py_mod_vcov - r_mod_vcov
# array([[6.93889390e-18, 1.08420217e-19],
#        [3.25260652e-19, 0.00000000e+00]])
```

PR #556, opened 2024-07-15

Jayhyung:
Hi @s3alfisc!

I will add more details to this PR soon.


Jayhyung:
It was kind of a big battle for me to resolve this .coverage issue ;).

s3alfisc:
> It was kind of a big battle for me to resolve this .coverage issue ;).

But I've seen that you prevailed! 😄

Jayhyung:
Hmm, i guess I got my tests(testing discrepancy of 1st stage F tests between `wald_test()`  and IV_Diag R package, when covariance-variance matrix type is set to be "hetero") crashes after merging this change #529. I actually tested after I rolled back to the previous version in my local machine, and found out that this merge really affects the value of F statistics. The discrepancy is small, but big enough to make the tests fail. Do you think that we only test "iid" and "CRV1" cases? Thanks!

Jayhyung:
Could you please let me know where i should put a line of code that installs `ivDiag` R package? I think `ivDiag` was not installed in the testing environment.

s3alfisc:
Hi @Jayhyung, sorry for my delayed response, I am currently traveling =) There's two spots to mention `ivDiag`: 

- the justfile where you'd add the `just install-r` command: https://github.com/py-econometrics/pyfixest/blob/506cd51708c74e90be3c03800e02e2dc928293d0/justfile#L24
- then we should update the contributing.md and mention the r installations here: https://github.com/py-econometrics/pyfixest/blob/506cd51708c74e90be3c03800e02e2dc928293d0/docs/contributing.qmd#L87

And that should be it! =)

Jayhyung:
Hmm, I think that change in the `justfile` isn't reflected on the github action workflow. I guess that we have to add `ivDiag` package to the following location([here](https://github.com/py-econometrics/pyfixest/blob/master/.github/workflows/ci-tests.yaml#L43)). The relavant part is the line 43


```yaml
      - name: install R packages
        run: Rscript -e 'install.packages(c("fixest", "broom", "did2s", "clubSandwich", "wildrwolf"))'
        shell: bash
```

I acatually do not have permission to change this, so could you modify this part? In `justfile`, the R package installation part read like the following : 

```justfile
# Install R dependencies

install-r:
	Rscript -e "install.packages(c('broom', 'clubSandwich', 'did2s', 'fixest', 'wildrwolf', 'reticulate'), repos='https://cran.rstudio.com'); install.packages('ritest', repos=c('https://grantmcdermott.r-universe.dev', 'https://cloud.r-project.org')); install.packages('ivDiag', repos='http://cran.us.r-project.org')"

```

Jayhyung:
No hurry, Alex! I will be happy to get the reply when you return back :).

s3alfisc:
Yes, you're right, I completely forgot about the most important section 🙈 Committed the merge just now! Interesting to learn that you do not have write permission to the github action, I'll take a look at that later. Maybe the actions are extra fenced because changes to them might impact billing?

Jayhyung:
Thanks for the reply! There is another issue haha... I don't know why this happens, but the some tests in `test_i.py` are stuck at the python 3.10 version. I found out that `def test_i_vs_fixest()` is the issue(when I commentize this test, all the tests passed without an issue). My research on this issue is related with memory limit, but  I'm not sure about why this happens. Also, this test `def test_i_vs_fixest()` does not seem to be related with the tests related with IV regressions. Do you have any clue? 

PS : Could you also abort all the workflows in the queue that I made? I tried to check on this error by sending multiple commits, but it looks like they were stuck ;).. Sorry about the mess...

Jayhyung:
I reduced the number of observations used in `test_i_vs_fixest` by restricting the range of `year>=2005`, but now the test fails in the python 3.9...  😂

Jayhyung:
I reduced the number of obervations in `test_i.py` not to make it crash, [here](https://github.com/py-econometrics/pyfixest/pull/556/files#diff-d8ff1648a4cf1d5bacf6e61266ecd99ee9d06687d0bcbfedf32e678a31ae3944R55). I'm not sure whether this is the best practice, let me know if this may be an issue.

There are a couple of things to be mentioned about this PR. I added the F statistics(including effective F stat by Olea and Pflueger 2013, as this is frequently reported in the empirical papers that use IV) of 1st stage regression as default features. The usage example that I added to the code is self-evident, but here is how we implement this

```python
        import numpy as np
        import pandas as pd

        from pyfixest.estimation.estimation import feols

        #Set random seed for reproducibility
        np.random.seed(1)

        # Number of observations
        n = 1000

        # Simulate the data
        # Instrumental variable
        z = np.random.binomial(1, 0.5, size=n)
        z2 = np.random.binomial(1, 0.5, size=n)
        # Endogenous variable
        d = 0.5 * z + 1.5 * z2 + np.random.normal(size=n)

        # Control variables
        c1 = np.random.normal(size=n)
        c2 = np.random.normal(size=n)

        # Outcome variable
        y = 1.0 + 1.5 * d + 0.8 * c1 + 0.5 * c2 + np.random.normal(size=n)

        # Cluster variable
        cluster = np.random.randint(1, 50, size=n)

        weights = np.random.uniform(1, 3, size=n)

        # Create a DataFrame
        data = pd.DataFrame({
            'd': d,
            'y': y,
            'z': z,
            'z2' : z2,
            'c1': c1,
            'c2': c2,
            'cluster': cluster,
           'weights' : weights
        })

        vcov_detail = "iid"

        fit_iv = feols("y ~ 1 + c1  + c2 | d ~ z ", data=data,
                 vcov=vcov_detail, weights= "weights")
        fit_iv.IV_Diag()

        print("(Unadjusted) F stat : " ,fit_iv._f_stat_1st_stage)
        print("Effective F stat : " ,fit_iv._eff_F )
        # The example above generates the following results
        # (Unadjusted) F stat :  52.81535560457482
        # Effective F stat :  48.661542741328205
```

After calling the iv regression model `fit_iv `, we need to call an extra method `fit_iv.IV_Diag()` to compute F statistics. 

Other diagnostics tests of IV will be added as default attributes, so that we can easily dump the releavant statistics into this method.

Please let me know if there is anything extra should be done on my part!

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,225 @@ def get_fit(self) -> None:
...
+
+        model1 = feols(
+            fml=formula,
+            data=data,
+            vcov=vcov_detail,
+            weights=weight_detail,
+            collin_tol=1e-10,
+        )
+        # Ensure model1 is of type Feols
+        if isinstance(model1, Feols):
```
s3alfisc, line 272: Would it make sense to call `IV_Diag` within the first stage? E.g. here we could add a line 

```python 
self.IV_diag()
```
Motivation is that I assume that almost all users who run an IV regression will be interested in obtaining those statistics. 

Additionally, we could add a function argument `statistics` to `IV_Diag` which allows users to select **additional** statistics post estimation that we will add in the future. E.g. we'd have smth as 

```python 
def IV_Diag(self, statistics: Optional[list[str]] = None): 
   # default
   iv_diag_statistics = ["effective_f", "f_stat"]
   if not statistics: 
      iv_diag_statistics += statistics
```

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,225 @@ def get_fit(self) -> None:
...
+            self._v_hat = model1._u_hat
+
+            # Store 1st stage model for further use
+            self._model_1st_stage = model1
+        else:
+            raise TypeError("The first stage model must be of type Feols")
+
+    def IV_Diag(self) -> None:
+        """Implement IV weakness test(F-test)."""
+        # This method covers hetero-robust and clustered-robust F stats
```
s3alfisc, line 289: Could you move the description of the method into the string above? We're using the `numpy` format, e.g. the docstring could look like this: 

```python
def IV_weakness_test():
    """
    Implement IV weakness test (F-test).

    This method covers hetero-robust and clustered-robust F statistics. It produces two statistics:
    
    - self._f_stat_1st_stage: F statistics of first stage regression
    - self._eff_F: Effective F statistics (Olea and Pflueger 2013) of first stage regression
    
    Notes
    -----
    "self._f_stat_1st_stage" is adjusted to the specification of vcov. If vcov_detail = "iid", F statistics is not adjusted,
    otherwise it is always adjusted.
    
    Example
    -------
    The following is an example usage of this method:
    
        ```{python}
   
        import numpy as np
        import pandas as pd
        from pyfixest.estimation.estimation import feols

        # Set random seed for reproducibility
        np.random.seed(1)

        # Number of observations
        n = 1000

        # Simulate the data
        # Instrumental variable
        z = np.random.binomial(1, 0.5, size=n)
        z2 = np.random.binomial(1, 0.5, size=n)

        # Endogenous variable
        d = 0.5 * z + 1.5 * z2 + np.random.normal(size=n)

        # Control variables
        c1 = np.random.normal(size=n)
        c2 = np.random.normal(size=n)

        # Outcome variable
        y = 1.0 + 1.5 * d + 0.8 * c1 + 0.5 * c2 + np.random.normal(size=n)

        # Cluster variable
        cluster = np.random.randint(1, 50, size=n)
        weights = np.random.uniform(1, 3, size=n)

        # Create a DataFrame
        data = pd.DataFrame({
            'd': d,
            'y': y,
            'z': z,
            'z2': z2,
            'c1': c1,
            'c2': c2,
            'cluster': cluster,
            'weights': weights
        })

        vcov_detail = "iid"
        
        # Fit OLS model
        fit_ols = feols("y ~ 1 + d + c1 + c2", data=data, vcov=vcov_detail)

        # Fit IV model
        fit_iv = feols("y ~ 1 + c1 + c2 | d ~ z", data=data, vcov=vcov_detail, weights="weights")
        fit_iv.IV_Diag()

        print("(Unadjusted) F stat :", fit_iv._f_stat_1st_stage)
        print("Effective F stat :", fit_iv._eff_F)

        # The example above generates the following results
        # (Unadjusted) F stat : 52.81535560457482
        # Effective F stat : 48.661542741328205
   ```
    """
    pass
   
```

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,225 @@ def get_fit(self) -> None:
...
+        _Y = self._Y
+
+        # Store names of instruments from Z matrix
+        self._non_exo_instruments = list(set(self._coefnames_z) - set(self._coefnames))
+
+        # Prepare your data
+        # Select the columns from self._data that match the variable names
+        data = pd.DataFrame(self._Z_1st_stage, columns=self._coefnames_z)
+
+        # Store instrument variable matrix for futyre use
```
s3alfisc, line 226: > futyre
small typo =)

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,225 @@ def get_fit(self) -> None:
...
+        """
+        ##############################################################
+
+        self.first_stage()
+        self._p_iv = len(self._non_exo_instruments)
+
+        # Create an identity matrix of size p_iv by p_iv
+        # Pad the identity matrix with zeros to make it of size p_iv by k
+        p_iv = self._p_iv  # number of IVs
+        k = self._model_1st_stage._k  # number of estimated coefficients of 1st stage
```
s3alfisc, line 365: Here, we do not count the number of fixed effects, which is in line with pyfixest defaults (`fixef_K = 'none')` in the `pf.ssc()` function. Once we allow for `fixef_K = 'full'`, we need to update this as well. The number of estimated fixed effects is stored in `self._k_fe`. Or would we never include the "estimated fixed effects" here? Maybe @apoorvalal can guide us here :)

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,225 @@ def get_fit(self) -> None:
...
+
+        #print("(Unadjusted) F stat : " ,fit_iv._f_stat_1st_stage)
+        #print("Effective F stat : " ,fit_iv._eff_F )
+        ## The example above generates the following results
+        ## (Unadjusted) F stat :  52.81535560457482
+        ## Effective F stat :  48.661542741328205
+        """
+        ##############################################################
+
+        self.first_stage()
```
s3alfisc, line 359: Should we not call `self.first_stage()` outside of `IV_Diag`? My intuition would have been to call `IV_Diag` within `first_stage()` and not the other way around? 🤔

tests/test_iv.py
```diff
@@ -1,31 +1,215 @@
...
+    )
+
+    return {
+        "with_weights": result_with_weights.rx2("F_stat"),
+        "without_weights": result_without_weights.rx2("F_stat"),
+        "data": data,
+    }
+
+
+@pytest.mark.parametrize("has_weight", [0.0, 1.0])
```
s3alfisc, line 102: Maybe annoyingly nit-picky (sorry about that! 😅 ): 

I would rename the test parametrization to 
```python
@pytest.mark.parametrize("has_weight", [False, True])
@pytest.mark.parametrize("adj_vcov", ["iid", "hetero", {"CRV1":"cluster"}])
```
as it is easier to read?

pyfixest/estimation/feiv_.py
Jayhyung: Oh yes, you are right. I should have put `IV_Diag` to `first_stage` method to make it a default option. Thanks for pointing out :)

pyfixest/estimation/feiv_.py
Jayhyung: This layout looks far great! Yes, we also have to follow the numpy docstring. I will place the new method `IV_weakness_test()` as a separate method, and put the documentation as you suggested!

pyfixest/estimation/feiv_.py
Jayhyung: 🙈

pyfixest/estimation/feiv_.py
Jayhyung: Could you elaborate a bit more about this? My understanding is that we currently do not allow any other option other than `None` in the ssc in the `fixef_k` argument , but once we allow for that in the future, this may affect the number of estimated coefficients counted in `k`, right? Currently, this `k` is equal to the # of instruments + # of exogenous covariates + intercept(if fixed effects are not identified). My question is, are we counting the number of fixed effects in computing `k` in the future? If this is the case, then I will add the code that deals with this as the following


```python

k = self._model_1st_stage._k - self._model_1st_stage._k_fe

# When there is not fixed effects, then _k_fe = 0
```

pyfixest/estimation/feiv_.py
Jayhyung: Yes, you are right. I will handle this!

tests/test_iv.py
Jayhyung: This is a bad habit of mine that I always tend to work on numeric value ;). I will fix this!

s3alfisc:
> I reduced the number of obervations in test_i.py not to make it crash, [here](https://github.com/py-econometrics/pyfixest/pull/556/files#diff-d8ff1648a4cf1d5bacf6e61266ecd99ee9d06687d0bcbfedf32e678a31ae3944R55). I'm not sure whether this is the best practice, let me know if this may be an issue.

Probably ok, but did it crash because of an error, or because of a test failure? Overall quite strange behavior as the base `Feols` class, which is only called in the test, is untouched by your PR? 🤔

PR #563, opened 2024-07-21

Jayhyung:
I re-uploaded commits to resolve .coverage issue #556.


Jayhyung:
I'm sorry,  @s3alfisc. I had to make a new dev branch due to `.coverage` issue. Due to this change, we need to add again a command line that installs `ivDiag` package in the following [line](https://github.com/py-econometrics/pyfixest/blob/master/.github/workflows/ci-tests.yaml#L43). Could you make commit merge again? Thank you so much for your help!

Jayhyung:
Thanks! I think this should be the end of this PR, but let me know if there is anything! Just for the record, I was just curious about this issue you raised [here](https://github.com/py-econometrics/pyfixest/pull/556#discussion_r1685696137). Is this good to leave at it is now? Thanks!

apoorvalal:
Nice; good to see my variable names in ivDiag carry over into pyfixest.

While you're at it, any interest in implementing Anderson Rubin? Only really practical for a single endog variable but pretty intuitive https://github.com/apoorvalal/ding_causalInference_python/blob/master/Chapter23IVeconometrics.ipynb

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,263 @@ def get_fit(self) -> None:
...
+
+        # Set fix effects, cluster, and weight options to be passed to feols.
+
+        if self._has_fixef:
+            FE_vars = self._fixef.split("+")
+            data[FE_vars] = self._data[FE_vars]
+            formula = f"Y ~ {independent_vars} | {self._fixef}"
+        else:
+            formula = f"Y ~ {independent_vars}"
+
```
s3alfisc, line 245: Here, an alternative option would be to add full `FixestFormula` to the `Feiv` model via the [FixestMulti.add_fixest_multi_context](https://github.com/py-econometrics/pyfixest/blob/d77a7efee9397c860303ede8f44d83bad15a52a7/pyfixest/estimation/FixestMulti_.py#L395C21-L395C45) method. 

Then we could obtain the first stage as stored in the `FixestFormula` and as used in [model_matrix_fixest](https://github.com/py-econometrics/pyfixest/blob/d77a7efee9397c860303ede8f44d83bad15a52a7/pyfixest/estimation/model_matrix_fixest_.py#L74). 

I'm open to both the current and the FixestFormula solution, with slight preference for the FixestFormula =) What do you think?

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,263 @@ def get_fit(self) -> None:
...
+        # Do first stage regression
+
+        model1 = feols(
+            fml=formula,
+            data=data,
+            vcov=vcov_detail,
+            weights=weight_detail,
+            collin_tol=1e-10,
+        )
+
```
s3alfisc, line 271: This check is only to avoid that the first stage model might be an IV model itself?

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,263 @@ def get_fit(self) -> None:
...
+        iv_diag_statistics = ["f_stat", "effective_f"]
+
+        if statistics:
+            iv_diag_statistics += statistics
+
+        self.IV_weakness_test(iv_diag_statistics)
+
+    def IV_weakness_test(self, iv_diag_statistics: Optional[list[str]] = None) -> None:
+        """Implement IV weakness test (F-test).
+
```
s3alfisc, line 310: Can we add a docstring note for the `iv_diag_statistics` argument?

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,263 @@ def get_fit(self) -> None:
...
+            )  # number of estimated coefficients of 1st stage
+
+            # Generate matrix R that tests the following;
+            # H0 : \beta_{z_1} = 0 & ... & \beta_{z_{p_iv}} = 0
+            #      where z_1, ..., z_{p_iv} are the instrument variables
+            # H1 : H0 does not hold
+
+            identity_matrix = np.eye(p_iv)
+            R = np.zeros((p_iv, k))
+            if self._has_fixef:
```
s3alfisc, line 411: Here we implicitly encode that the order of the first stage design matrix is [intercept, instruments, exogenous], right? I am afraid this might be "unsafe" in case this order was ever to change. Have been thinking about this about the last 5 minutes but have no great answer.

s3alfisc: Could we maybe do something like 

```python
%load_ext autoreload
%autoreload 2
import pyfixest as pf 
import numpy as np

data = pf.get_data()

fit = pf.feols("X1 ~ Z1", data = data)

k = fit._k 
p_iv = 1

identity_matrix = np.eye(p_iv)
R = np.zeros((p_iv, k))

# those we might store in the Feiv object? 
instruments = ["Z1"]
idx = [fit._coefnames.index(x) for x in instruments if x in fit._coefnames]
idx

R[:, 1 : p_iv + 1] = identity_matrix
R
# array([[0., 1.]])
```
?

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,263 @@ def get_fit(self) -> None:
...
+        if "effective_f" in iv_diag_statistics:
+            self.eff_F()
+
+    def eff_F(self) -> None:
+        """Compute Effective F stat (Olea and Pflueger 2013)."""
+        # If vcov is iid, redo first stage regression
+        if self._vcov_type_detail == "iid":
+            self._vcov_type_detail = "hetero"
+            self.first_stage()
+
```
s3alfisc, line 429: If running `self.first_stage()` is only about computing inference, we could also just run `self._model_1st_stage.vcov("hetero")`, which will only update all of the vcov information + residual standard errors + everything around inference (t-stats, ses, pvalues, etc), but not re-fit the entire model. Might be slightly more efficient?

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,263 @@ def get_fit(self) -> None:
...
+        # Calculate the cross-product of the instrument matrix
+        Q_zz = Z.T @ Z
+
+        # Extract the robust variance-covariance matrix
+        vcv = self._model_1st_stage._vcov
+
+        # Map the instrument names to their indices in the parameter list
+        # Number of rows/columns in vcv
+
+        # Extract the submatrix
```
s3alfisc, line 461: Here again it might be "safer" to extract the position of the instruments based on "explicit" knowledge rather than "implicit" knowledge"?

pyfixest/estimation/feiv_.py
```diff
@@ -193,3 +206,263 @@ def get_fit(self) -> None:
...
+            #      where z_1, ..., z_{p_iv} are the instrument variables
+            # H1 : H0 does not hold
+
+            identity_matrix = np.eye(p_iv)
+            R = np.zeros((p_iv, k))
+            if self._has_fixef:
+                R[:, :p_iv] = identity_matrix
+            else:
+                R[:, 1 : p_iv + 1] = identity_matrix
+
```
s3alfisc, line 415: Here things very nicely come together I think! =)

pyfixest/estimation/feiv_.py
s3alfisc: Ah, I think that we should actually have what I call "instruments in `self._non_exo_instruments` above, or not?

pyfixest/estimation/feiv_.py
s3alfisc: Actually, you pretty much do this here I think 😄 

```python
            [
                self._model_1st_stage.coef()[instrument]
                for instrument in self._non_exo_instruments
            ]
```

pyfixest/estimation/feiv_.py
Jayhyung: Oh yes, you are right. The only purpose of this code here is to computing new `vcov`. I will update this part as you say!

pyfixest/estimation/feiv_.py
Jayhyung: Yes, you are right. I think there are two points that implicitly assume the position of instruments 

```python 

....
            if self._has_fixef:
                R[:, :p_iv] = identity_matrix
            else:
                R[:, 1 : p_iv + 1] = identity_matrix
....

           iv_pos = range(0, self._p_iv) if self._has_fixef else range(1, self._p_iv + 1)


```

This will be problematic if there is any chance that the positions of IVs changes, so my strategy will be 1) extract the first index of IVs from the 1st stage regression and the number of IVs (i.e. `_p_iv`), 2) and make range object using this to locate identity matrix and and extract submatrix of `vcov`.

The change code will be like the following : 

```python
....
    def extract_indices_IVs(self):
        """Extract indices of instrument variales in the first stage regression

        Returns
        -------
        None

        """
                
        indices = []
        for instrument in self._non_exo_instruments:
            if instrument in self._coefnames_z:
                index = self._coefnames_z.index(instrument)
                indices.append(index)
        return indices

....

       self._iv_loc_first = np.min(iv_loc)
       iv_loc_first = self._iv_loc_first
       R[:,iv_loc_first:p_iv+iv_loc_first] = identity_matrix

....

        # Extract the submatrix
        iv_pos = range(self._iv_loc_first, self._p_iv+self._iv_loc_first)
```

pyfixest/estimation/feiv_.py
Jayhyung: Please, correct me if I didn't understand this well. Originally, I directly initiated Feols class to do the first stage like the following; 

```python
        #  Start First Stage

        model1 = Feols(
            Y=_endogvar,
            X=_Z,
            weights=np.ones(self._weights.shape[0]).reshape(-1, 1),
            collin_tol=self._collin_tol,
            coefnames=self._coefnames_z,
            weights_name=None,
            weights_type=None,
        )
        model1.get_fit()
```

The issue that I had is whenever I tried to implement inference such as `wald_test`, I always encountered the issues that the necessary attributes are not defined, and had to define the attributes to proceed. I think [FixestMulti.add_fixest_multi_context](https://github.com/py-econometrics/pyfixest/blob/d77a7efee9397c860303ede8f44d83bad15a52a7/pyfixest/estimation/FixestMulti_.py#L395C21-L395C45) method allows to add more arguments(i.e. context) to define the attributes that might be needed for doing other tasks(such as `wald_test`)? I was actually not too familiar with `FixestFormula` so this will take more time to fix the code, but I would implement this way if this is the better option.

pyfixest/estimation/feiv_.py
Jayhyung: This is actually to avoid `mypy` error. If I do not force `model1` to be the `Feols` class like this, it looks like mypy also checks whether these attributes are defined in `FixestMulti` class, and outputs the error flags like the followings; 

```mypy
> git -c user.useConfigOnly=true commit --quiet
debug statements (python)................................................Passed
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...........................................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
check for added large files..............................................Passed
Validate GitHub Workflows............................(no files to check)Skipped
ruff.....................................................................Passed
ruff-format..............................................................Passed
nbqa-ruff............................................(no files to check)Skipped
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

pyfixest/estimation/feiv_.py:286: error: Item "FixestMulti" of "Feols | FixestMulti" has no attribute "_beta_hat"  [union-attr]
pyfixest/estimation/feiv_.py:289: error: Item "FixestMulti" of "Feols | FixestMulti" has no attribute "_Y_hat_link"  [union-attr]
pyfixest/estimation/feiv_.py:292: error: Item "FixestMulti" of "Feols | FixestMulti" has no attribute "_u_hat"  [union-attr]
pyfixest/estimation/feiv_.py:434: error: Item "FixestMulti" of "Feols | FixestMulti" has no attribute "_k"  [union-attr]
pyfixest/estimation/feiv_.py:460: error: Item "FixestMulti" of "Feols | FixestMulti" has no attribute "wald_test"  [union-attr]
pyfixest/estimation/feiv_.py:461: error: Item "FixestMulti" of "Feols | FixestMulti" has no attribute "_f_statistic"  [union-attr]
pyfixest/estimation/feiv_.py:462: error: Item "FixestMulti" of "Feols | FixestMulti" has no attribute "_p_value"  [union-attr]
pyfixest/estimation/feiv_.py:500: error: Item "FixestMulti" of "Feols | FixestMulti" has no attribute "_vcov"  [union-attr]
Found 8 errors in 1 file (checked 1 source file)

```

Jayhyung:
@apoorvalal Oh yes, I'm planning to do this for my next PR. `ivDiag` package was a good reference for me, so I borrowed your variable names and effective F stat part a lot :).

pyfixest/estimation/feiv_.py
```diff
@@ -401,17 +437,20 @@ def IV_weakness_test(self, iv_diag_statistics: Optional[list[str]] = None) -> No
...
+            iv_loc_first = self._iv_loc_first
+
             # Generate matrix R that tests the following;
             # H0 : \beta_{z_1} = 0 & ... & \beta_{z_{p_iv}} = 0
             #      where z_1, ..., z_{p_iv} are the instrument variables
             # H1 : H0 does not hold
 
+            # Pad identity matrix to implement wald-test
             identity_matrix = np.eye(p_iv)
             R = np.zeros((p_iv, k))
```
s3alfisc, line 452: Can we not directly subset as `R[:, iv_loc]`?

pyfixest/estimation/feiv_.py
```diff
@@ -460,7 +499,9 @@ def eff_F(self) -> None:
 
         # Extract the submatrix
```
s3alfisc, line 501: Could we not use the output of `iv_loc = self.extract_indices_IVs()` here as well?

Jayhyung: Thanks, I just fixed this part by replacing with one line of code!

Jayhyung:
Any chance that this can be merged soon? There is one issue that was not resolved yet, but I left my thought. Thanks!

s3alfisc:
Wow @Jayhyung, I completely missed your status update, sorry! Will review this tomorrow evening =)

pyfixest/estimation/feiv_.py
s3alfisc: Yes, absolutely fine to stick with `feols()` here I think! I'll have to check though what's been going wrong with `Feols` and `add_fixest_multi_context` - in my mind, you should have all the attributed available after calling `get_fit()`,  `vcov()` and `add_fixest_multi_context` on a `Feols` object 🤔 `get_fit()` computes the estimates and residuals, `vcov()` the vcov and adds attributes for small sample corrections & also calls `get_inference()`, `add_fixest_multi_context` adds context outside of the scope of `Feols` (i.e. the model formula, which `Feols` is agnostic of, or the input data frame if specificed by the user).

Jayhyung: Thanks for the comment. I will look into this!

s3alfisc:
Hi @Jayhyung , I have pushed three changes: 

- The first is completely unrelated and solved a mypy type check error for the predict method. WIthout changing it, I wasn't able to commit anything. 
- The second commit computes the first stage by default in the `FixestMulti` class. I.e. whenever a user calls an IV model via `pf.feols()`, the code goes through `FixestMulti`, calls `Feiv`, and then by default calls `Feiv.first_stage()`. I think this makes sense as users would expect first stage regression results in the default results? What do you think? 
- The last commit adds a [self._iv_loc](https://github.com/py-econometrics/pyfixest/pull/563/commits/9434aba81b96f4b7ed128decb6a7ee847270a1d4) attribute and uses it for subsetting. 

If you approve (and if all tests pass), I would then merge this PR!

Jayhyung:
According to the error message, `test_multicol_overdetermined_iv()` in `test_others.py` displays the following error 

```python
{
	"name": "ValueError",
	"message": "Shape of passed values is (997, 4), indices imply (997, 3)",
	"stack": "---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[88], line 10
      8 data = get_data()
      9 fit = feols(\"Y ~ X2 + f1 | f1 | X1 ~ Z1 + Z2\", data=data, ssc=ssc(adj=False))
---> 10 fit.first_stage()

File ~/pyfixest/pyfixest/estimation/feiv_.py:227, in Feiv.first_stage(self)
    225 print(self._non_exo_instruments)
    226 print(self._Z_1st_stage)
--> 227 data = pd.DataFrame(self._Z_1st_stage, columns=self._coefnames_z) 
    229 # Store instrument variable matrix for future use
    230 _Z_iv = data[self._non_exo_instruments]

File ~/.local/lib/python3.11/site-packages/pandas/core/frame.py:827, in DataFrame.__init__(self, data, index, columns, dtype, copy)
    816         mgr = dict_to_mgr(
    817             # error: Item \"ndarray\" of \"Union[ndarray, Series, Index]\" has no
    818             # attribute \"name\"
   (...)
    824             copy=_copy,
    825         )
    826     else:
--> 827         mgr = ndarray_to_mgr(
    828             data,
    829             index,
    830             columns,
    831             dtype=dtype,
    832             copy=copy,
    833             typ=manager,
    834         )
    836 # For data is list-like, or Iterable (will consume into list)
    837 elif is_list_like(data):

File ~/.local/lib/python3.11/site-packages/pandas/core/internals/construction.py:336, in ndarray_to_mgr(values, index, columns, dtype, copy, typ)
    331 # _prep_ndarraylike ensures that values.ndim == 2 at this point
    332 index, columns = _get_axes(
    333     values.shape[0], values.shape[1], index=index, columns=columns
    334 )
--> 336 _check_values_indices_shape_match(values, index, columns)
    338 if typ == \"array\":
    339     if issubclass(values.dtype.type, str):

File ~/.local/lib/python3.11/site-packages/pandas/core/internals/construction.py:420, in _check_values_indices_shape_match(values, index, columns)
    418 passed = values.shape
    419 implied = (len(index), len(columns))
--> 420 raise ValueError(f\"Shape of passed values is {passed}, indices imply {implied}\")

ValueError: Shape of passed values is (997, 4), indices imply (997, 3)"
}
```

This is due to the mismatch of the number of columns of `self._Z_1st_stage` and the number of elements in `self._coefnames_z`. The following picture displays  `self._Z_1st_stage`(which is the (unweighted) 2-D array of  `Z` array in `feiv_.py`  )

```python

['Z1', 'Z2', 'X2'] # this is "self._coefnames_z"


[[ 1.29816431  0.61720836  1.59227661  0.        ] # This is "self._Z_1st_stage"
 [-0.20504835 -3.78398468 -3.88557536  0.        ]
 [ 1.39615258 -1.35467391 -1.46324726  0.        ]
 ...
 [ 1.84211198  2.75574557  1.43801845  0.        ]
 [-0.65536947  3.99906064  3.6063778   0.        ]
 [ 1.22667048  0.72047835 -0.15614689  0.        ]]

```

It looks like if there is multi-colinearity issue, they drop the variable `f1`, and add `0` vector  to `self._Z_1st_stage` array instead. I guess the issue is in dealing with constructing `Z` array  when there is multi-colinearity issue. Any thought?

Jayhyung:
I just fixed the error. After this iteration, i think it is ready to be merged. Let me know if there is anything!

s3alfisc:
I think now Z is already multiplied with the weights, so we have to drop the weights argument from the feols() call at the end? 

Also, can we add the weights_type function argument to the feols call()? 

Btw, this is a much more intricate PR than I initially assumed, but we'll get there 😀

Jayhyung:
I actually differentiated `Z` that is unweighted(which is `self._Z_1st_stage`) and `Z` that i weighted(which is `self._Z`), so that this unweighted `Z` is passed to the first stage regression. The reason I do this is that `Z` in the beginning of the `feiv` class contains `intercept` term, which is also weighted if the weight feature is specified, so that we end up having "weighted" intercept term. So the first stage regression take this weighted intercept as a new independent variable. I actually wanted to prevent this by using unweighted `Z` in the first stage regression.

Yes, I added the weight type to the first stage!

Please, feel free to let me know  if there is anything that should be corrected and improved!

s3alfisc:
Perfect! The `self._Z = self._Z * w` part was what I was missing =) Just looked before having dinner and got quite confused 🤕 I'll merge this after the CI run! This was really a lot of work, thanks a lot! =)

PR #567, opened 2024-07-28
implement update method for coef vector 

apoorvalal:
added simple `update` method to feols object to update coefficient vector in online-learning settings. Use demonstrated in quickstart notebook. 
Also fleshed out `confint` section of quickstart notebook.


s3alfisc:
Thanks for adding the tests @apoorvalal - should I merge this as is & we can add inference for the `update` method in a separate PR?

apoorvalal:
I think merging as is should be fine; the update method is largely about
coefficient vectors anyway. I think inference for new observations is a
bigger PR (might need either conformal inference or anytime-valid
regression methods for that).

--

Apoorva Lal

***@***.*** | apoorvalal.github.io

Issue #566, opened 2024-07-25
Adding Standard Error of Prediction


greenguy33:
Pyfixest currently has `predict()` function but there is not currently functionality to provide estimates of the variance of the predictions. (Some) other tools do include this, such as:

-  Statsmodels. The `get_prediction()` function returns a `PredictionResults`object that [includes an estimate of the variance of the prediction](https://www.statsmodels.org/dev/generated/statsmodels.regression.linear_model.PredictionResults.html) as the attribute `var_pred_mean`.

- Stata. Includes stdp option for generating standard error of the prediction (see bottom of Pg. 1899 of [this link](https://www.stata.com/manuals13/r.pdf)).

This functionality would be useful for users who want to compute prediction intervals to estimate uncertainty in the prediction. If folks are interested in this feature, I am happy to work on adding this to Pyfixest. It would require calculating the full covariance matrix including all fixed effects. Then the variance of each prediction is simply `X * M * X'` where X is a row of covariate data to predict on and M is the full covariance matrix.

s3alfisc:
Hy Hayden, I think this would be an awesome addition to pyfixest! 

I think the best place to add it would be via the 'predict' method? We could add an argument, eg named "ci", that is by default False but if set to True, returns the prediction interval with the predicted values. What do you think? 

I agree that the key step will be to compute the full vcov matrix, including all estimates fixed effects. Do you already have an idea how to best built this based on the existing code base? I've thought about it for the last few minutes and am not sure if there is a "quick win". But I'll have to think about this more in the next days. Maybe one option would be to build a first MVP implementation that only works without fixed effects? 

Anyways, I'd be super happy to have this feature! Cc'ing @b-knight as this might be of interest for him as well =)

greenguy33:
Hey Alex,

Glad you think this would be a worthwhile addition!

I agree with you that the calculation of the full covariance matrix is the challenging aspect of this task. I don't have an idea how to do it right now - but I view it as an opportunity to learn more about how covariance matrices work. The nice thing is that we can compare results to the existing packages that already offer this functionality to validate/double check.

I guess there are two ways to approach the task:
1) start with the "easy win" of computing the prediction error without fixed effects, and expand later to include the full covariance matrix
2) tackle the difficult part first and roll out the full covariance matrix before adding the standard error option to the predict function

In general, my software engineer instincts tell me to get the hard part done first, but I am ok with either approach here and will cede to your opinion as to which order you prefer to do things.

[...]

PR #573, opened 2024-08-01
Prediction error for non-fe model

greenguy33:
This pull request references issue #566 

This is not a finished product but I took a first pass at implementing prediction standard error (stdp) within the predict function. As discussed it is only implemented for models without fixed effects for now. I'm creating this pull request to solicit feedback on the approach used. I confirmed that results returned are the same as statsmodels' `get_prediction()` function for a test dataset.

Note that CI's are not implemented, but the prediction error is returned. In order to implement CIs, we need to provide a z-score based on the percentage CI we want and then do some simple arithmetic. For the sake of keeping things simple and not adding too many new arguments to the `predict()` function, I left that out for now.


s3alfisc:
Looks very good so far! 👍😄  

I guess at some point we might want to clean up the different concepts of predicted values, i.e. `self._Y_hat`, `self._`self._Y_untransformed`, and I also think there is also `self._Y_hat_link`? 

I suppose next steps would be to implement some basic tests (i.e. against `statsmodels`), to add an example, and maybe update the docs? I am not sure if the `predict` method is even mentioned in `quickstart.ipynb` 🤔

pyfixest/estimation/feols_.py
```diff
@@ -1436,6 +1436,7 @@ def predict(
         atol: float = 1e-6,
         btol: float = 1e-6,
         type: str = "link",
+        compute_stdp: bool = False
     ) -> np.ndarray:
```
s3alfisc, line 1440: Btw, I think that a few unit tests fail because the code assumes that `predict` returns a `np.ndarray` - if we change to a `pd.DataFrame` (which I think makes most sense), we need to change these. 

Alternatively, we could change to `-> Union[np.ndarray, pd.DataFrame]` and simply return the numpy array if `compute_stdp = False`.

greenguy33:
Yep, I didn't touch the tests or docs yet but happy to do that as a next step!
I don't think predict is mentioned in the docs because when I started using the library I had to go digging for it :) I'll make sure to add that when I have a chance to make the next update.
As a github related sidenote, do you know if there is a way to squash commits on my end when I make a pull request from my fork? It feels very dirty to be submitting a pull request with 25 commits when only 3-4 of them are relevant to the current project.

s3alfisc:
>  As a github related sidenote, do you know if there is a way to squash commits on my end when I make a pull request from my fork? It feels very dirty to be submitting a pull request with 25 commits when only 3-4 of them are relevant to the current project.

Yes, I can do a squash commit on my end when merging, so many many commits are no problem, I have this all the time :D 
image

> Yep, I didn't touch the tests or docs yet but happy to do that as a next step!
> I don't think predict is mentioned in the docs because when I started using the library I had to go digging for it :) I'll make sure to add that when I have a chance to make the next update.

Sounds great!

s3alfisc:
@pre-commit.ci autofix

s3alfisc:
pre-commit.ci autofix

s3alfisc:
Hi Hayden (@greenguy33), I am finally getting around to pick this up - would it be ok if I took the newly added functions and opened a new PR?

greenguy33:
> Hi Hayden (@greenguy33), I am finally getting around to pick this up - would it be ok if I took the newly added functions and opened a new PR?

@s3alfisc Absolutely, and sorry for dropping off the face of the earth after submitting this. I've been absolutely swamped with my grad work since then.

s3alfisc:
Cool! And no worries at all, I've been there myself... hope all your papers will progress smoothly!

s3alfisc:
pre-commit.ci autofix

s3alfisc:
pre.commit-ci autofix

s3alfisc:
pre-commit.ci autofix

PR #574, opened 2024-08-03
Bootstrap & Regression based on Sufficient Statistics

s3alfisc:
1. Adds a `bootstrap` method to `Feols`. It is possible to pass **custom** evaluation functions to the bootstrap as well as kwargs. 
2. This allows us to easily implement a function `feols_compressed`, that allows to run regressions based on sufficient statistics and compute standard errors by bootstrapping. 

There are still quite a lot of things to do: 
- The API of `feols_compressed` needs refinement. Function arguments need to be added (`reps`, `seed`, etc). 
- Interacted variables will fail. 
- Missing data leads to problems. 
- The function probably fails if there are not fixed effects in the model.
- Other things? 
- We also needs unit tests 😄 

Example: 

```python
import pyfixest as pf
from pyfixest.estimation.estimation import feols_compressed
import polars as pl

fml = "Y~ f1 | f2"
vcov = {"CRV1": "f2"}

data = pf.get_data().dropna().reset_index(drop=True)
fit = pf.feols(fml, data = data, vcov = vcov)
fit_compressed = feols_compressed(fml = fml, data = data, vcov = vcov)

pf.etable([fit, fit_compressed], digits = 6)
```
image


pyfixest/estimation/feols_.py
```diff
@@ -956,6 +957,114 @@ def wald_test(self, R=None, q=None, distribution="F"):
...
+        if block_bootstrap:
+            unique_groups = _data.select(_clustervar).unique().to_numpy().flatten()
+            grouped_data = {
+                group: _data.filter(pl.col(_clustervar) == group)
+                for group in set(unique_groups)
+            }
+
+        boot_stats = np.zeros((reps, _k))
+
+        for b in tqdm(range(reps)):
```
apoorvalal, line 1036: qq about multiprocessing and polars: would we get any performance gains if we parallelized this operation with `joblib` or `multiprocessing`? I don't know the internals of polars well enough to decide whether something basic like indexing `_data[indices]` will be multi-processed by default, so making the bootstrap operation parallel would induce additional overhead? 

If so, mabye the

apoorvalal:
i have an implementation of the parallelization; can you give me write access to the repo?

apoorvalal:
here it is 
https://github.com/py-econometrics/pyfixest/compare/bootstrap...apoorvalal:pyfixest:bootstrap

apoorvalal:
actually don't really need parallelization anymore since the analytic SEs work fine [good to see an implementation in feols already].

pyfixest/estimation/feols_.py
```diff
@@ -943,6 +944,114 @@ def wald_test(self, R=None, q=None, distribution="F"):
 
         return res
 
+    def bootstrap(
+        self,
+        reps: int,
+        seed: Optional[int] = None,
+        inplace: bool = False,
+        fit_func: Optional[Callable] = None,
+        **kwargs,
```
apoorvalal, line 953: would it make sense to offer a 'compressed' flag for the bootstrap where the bootstrap replicates are (optionally) run in compressed fashion? would speed things up a lot.

s3alfisc: Yes,agreed this would be cool to have. From the compressed version, we would sample from a multinomial to reflect the weights? Maybe there even is a smart way to resample from the compressed data that is identical to the uncompressed bootstrap that I'm not seeing?

s3alfisc:
Yes I am very shamelessly stealing from duckreg 😅

s3alfisc:
I've moved the compression steps "inside" the package - this allows to make use of the full power of formulaic. I.e. it is now possible to have interactions or "C()" syntax in the formula call. This works through a new argument to feols, "use_compression". 

If `use_compression = True` and users specify **one** fixed effect, this will now run a Mundlak style regression, which can be a good notch faster than the default: 

```python
%load_ext autoreload
%autoreload 2

import pyfixest as pf
import time 
import numpy as np
import pandas as pd

# Generate sample data
def generate_sample_data(N=40_000_000, seed=12345):
    rng = np.random.default_rng(seed)
    D = rng.choice([0, 1], size=(N, 1))
    X = rng.choice(range(20), (N, 2), True)
    Y = D + X @ np.array([1, 2]).reshape(2, 1) + rng.normal(size=(N, 1))
    Y2 = -1 * D + X @ np.array([1, 2]).reshape(2, 1) + rng.normal(size=(N, 1))
    df = pd.DataFrame(
        np.concatenate([Y, Y2, D, X], axis=1), columns=["Y", "Y2", "D", "f1", "f2"]
    ).assign(rowid=range(N))
    return df

data = generate_sample_data()

time1 = time.time()
fit_c = pf.feols(
    "Y ~ f1 | f2",
    data = data,
    vcov = vcov,
    use_compression = True, 
    ssc = pf.ssc(adj = False, cluster_adj = False)
)

time2 = time.time()

fit = pf.feols(
    "Y ~ f1 | f2",
    data = data,
    vcov = vcov,
    use_compression = False, 
    ssc = pf.ssc(adj = False, cluster_adj = False)
)
time3 = time.time()
print(f"Uncompressed: {time3-time2} seconds")
print(f"Compressed: {time2-time1} seconds")

# Uncompressed: 17.647636890411377 seconds
# Compressed: 7.318124055862427 seconds
```
SEs seem to not match 100%. Likely dof corrections? 
image

pyfixest/estimation/feols_.py
apoorvalal: I was thinking more along the lines of each bootstrap iteration involving (take uncompresseddata > sample (either IID or cluster) > compress > call compressed regression) as in duckreg. 
With polars, I'd assume the sample+compress step is fast enough to make this effective? If not, we can punt on this.

s3alfisc: I think / hope that that is what is already implemented? The function argument `fit_func` allows to pass a custom "evaluation function" to the bootstrap, and I think that I pass `feols_compressed` with the "long" data set. So we resample from the long data -> compress -> fit on the compressed data using sufficient statistics. But maybe I am still misunderstanding your suggestion? 🤔 

There's a few tricks that we could still apply, i.e. we could pre-compute $A_{g} = inv(X_{g}'X_{g})$ where $X_{g}$ contains all observations from cluster g, where X is already compressed. There is one caveat - I think that $A_{g}$ might turn out difficult to compute when there are cluster fixed effects. 

Then we'd only have to recompute $X_{g} y_{g\text{mean}}$ for each cluster that is drawn in the boostrap and can compute the "overall" coefficient as 

$$
\hat{\beta} = \sum_{g \in E} A_{g} \times X_{g} y_{\text{mean}}
$$

where $E$ are all indices of the resampled clusters, i.e. $E = {1, 2, 2, 3, 4, 8, 9}$. 

If we **do not** use the compression strategy, we could even pre-compute $\times X_{g} y_{g}$ for each g before running the bootstrap, with both $X$ and $y$ in "long" format. 

Key ideas are from [this MacKinnon paper](https://www.econstor.eu/bitstream/10419/247206/1/qed-wp-1465.pdf), equation (12) is the key equation for us:

s3alfisc:
Closed in favor of https://github.com/py-econometrics/pyfixest/pull/574.

PR #577, opened 2024-08-09
Small extension to etable that allows to relabel variables

dsliwka:
Allows to pass a dictionary to the labels argument in etable. This dictionary is used to relabel the variables in the displayed regression table. Applies to dependent variable, explanatory variables and fixed effects. Fixed effects can also be separately relabeled with felabels dictionary. Interaction terms will be relabeled using the labels of the interacted variables.


s3alfisc:
This looks great, thank you @dsliwka! I slightly updated the docs and added some very simple API tests. Will merge it tomorrow morning in case the CI passes after my minor changes=)

s3alfisc:
@all-contributors please add @dsliwka for code and docs

PR #581, opened 2024-08-14
Implementing outcome plot for panelview function

rafimikail:
Hi @s3alfisc , this will be the PR for Issue #582 to enhance panelview to be able to do outcome plot.

I still need to finalize this by doing sanity check on the results and might also need to add the run tests in `test_visualize.py`, I will update you once its complete


rafimikail:
#582

pyfixest/did/visualize.py
```diff
@@ -74,24 +77,72 @@ def panelview(
     )
     ```
     """
```
s3alfisc, line 79: I think the `subsamp` and `collapse_to_cohort` should start working if you move some of this code 

```python
    treatment_quilt = data.pivot(index=unit, columns=time, values=treat)
    treatment_quilt = treatment_quilt.sample(subsamp) if subsamp else treatment_quilt
```
above your new code?

s3alfisc:
Hi @rafimikail , thanks a lot! I have added an example to the DiD notebook =) Looks already quite good! Next steps would be to add docstrings, tests, and make sure that the `subsamp` and `sort_by_cohort` arguments work as intended?

pyfixest/did/visualize.py
rafimikail: Hi @s3alfisc , i think its not as simple as that with my current code, i have just added some changes to code though:
- Add subsamp and collapse_to_cohort for outcome plot
- Adding docstring for figsize,  outcome, and type

But im thinking to pivoting my data first before plotting it so that the implementation of the subsamp and collapse_to_cohort can be the same with the apoorva's code (just like you mentioned above)

I think another things to do from my end:
- Im considering to use different data for plotting the outcome plot to have more visible control vs treatment (pre) vs treatment (post) line, what do you think? @s3alfisc 
- Adding parameter to flexibly choose which unit/group to plot --> So user can specify which group to be plotted/compare

s3alfisc: >     Add subsamp and collapse_to_cohort for outcome plot --> I have tested it and it works just fine
>     Adding docstring for figsize, outcome, and type

Great, thank you! 

> But im thinking to pivoting my data first before plotting it so that the implementation of the subsamp and collapse_to_cohort can be the same with the apoorva's code (just like you mentioned above) just to make it more consistent

I think this would be great, if only for the reason to keep the code base simple. 

> Adding parameter to flexibly choose which unit/group to plot --> So user can specify which group to be plotted/compare

This would be very cool I think! Would you like to add this in this or in a separate PR?

> Im considering to use different data for plotting the outcome plot to have more visible control vs treatment (pre) vs treatment (post) line, what do you think? @s3alfisc 

Here I have to admit that I am not really following 😅 Did you mean to say "colors" instead of "data" 😄

s3alfisc: Yeah both args now work very nicely. Cool!

Subsamp: 
image

collapse: 
image

Maybe you can update the did notebook and update the example to `collapse_to_cohort = True`?

rafimikail:
Hi @s3alfisc , i have finished the development, changes i made:
- Adding option to specify unit for outcome plot
- Adding xlim and ylim when plotting the outcome plot
- Fixing the collapse_to_cohort for outcome plot
- Adding test
- Adding docstring
- Adding specify unit and collapse_to_cohort for outcome plot on the documentation

I would appreciate it if you could review this when you have a moment. Thank you!

s3alfisc:
Awesome, thank you Rafi! Will review this tomorrow!

s3alfisc:
Plots look fabulous at first sight!

rafimikail:
I’m not entirely sure why the documentation isn’t showing the results for the previous examples when i re-ran the code, i think that's the reason why it says conflicting, would greatly appreciate it if you could kindly check that one later, Thanks!

rafimikail:
I think its already been solved, apparently it conflicted with the `difference-in-differences.ipynb` by another branch merged last night, i think should be good to go now @s3alfisc

s3alfisc:
Hi @rafimikail, two short comments: 

1. Could you add info on `panelview` to the docs? You can do to by adding a bullet point `did.visualize.panelview` to the [_quarto.yml](https://github.com/py-econometrics/pyfixest/blob/9b0554ba286416b06f423ae2a49e94784b4b894e/docs/_quarto.yml#L78). 
2. I get some linting errors when running `mypy pyfixest/did/visualize.py` and `just lint`. Have you installed pre-commit? If you do, the linter (`ruff`) and `mypy` will always run before you commit (basically you will be notified of these things in advance). [See here on how to get started with this.](https://py-econometrics.github.io/pyfixest/contributing.html#code-style). Here are the `mypy` errors: 
```python
error: Recipe `lint` failed on line 62 with exit code 1
PS C:\Users\alexa\Documents\pyfixest> mypy pyfixest/did/visualize.py
pyfixest\did\visualize.py:113: error: Value of type "Iterable[Any]" is not indexable  [index]
pyfixest\did\visualize.py:187: error: No return value expected  [return-value]
```
and the linter errors via `ruff`: 
```python
pyfixest\did\visualize.py:113:69: E712 Avoid equality comparisons to `True`; use `if x[treat]:` for truth checks
    |
111 |             data_pivot = data_pivot.sample(subsamp)
112 |         if collapse_to_cohort:
113 |             treatment_starts = data.groupby(unit).apply(lambda x: x[x[treat] == True][time].min(), include_groups=False)
    |                                                                     ^^^^^^^^^^^^^^^^ E712
114 |             treatment_starts = treatment_starts.reset_index(name='treatment_start')
115 |             data = data.merge(treatment_starts, on = unit, how='left')
    |
    = help: Replace with `x[treat]`

pyfixest\did\visualize.py:125:63: E712 Avoid equality comparisons to `True`; use `if data[treat]:` for truth checks
    |
123 |         for unit_id in data_pivot.index:
124 |             unit_data = data_pivot.loc[unit_id]
125 |             treatment_times = data[(data[unit] == unit_id) & (data[treat] == True)][time]
    |                                                               ^^^^^^^^^^^^^^^^^^^ E712
126 |
127 |             # If the unit never receives treatment, plot the line in grey
    |
    = help: Replace with `data[treat]`
```
Let me know if you have any questions regarding this!

s3alfisc:
Just brought back support for pre-commit. Will try something.

rafimikail:
Hi @s3alfisc , i have added it to the docs as well as fixed the code so it passed the ruff when committing, kindly have a check again, Thanks!

pyfixest/did/visualize.py
```diff
@@ -72,26 +93,132 @@ def panelview(
...
-        treatment_quilt = treatment_quilt.loc[
-            treatment_quilt.sum(axis=1).sort_values().index
-        ]
-    if not ax:
-        f, ax = plt.subplots()
-    cax = ax.matshow(treatment_quilt, cmap="viridis", aspect="auto")
-    f.colorbar(cax) if legend else None
-    ax.set_xlabel(xlab) if xlab else None
-    ax.set_ylabel(ylab) if ylab else None
+    if type == "outcome" and outcome:
```
s3alfisc, line 110: Can we raise an informative error message when a user specifies `type = "outcome"` but does not provide an outcome variable? 

I.e. we could do 

```python
if type == "outcome" and not outcome: 
   raise ValueError("You specified ... but ...")
```
We would then also add a test in `test_error_warnings.py`

pyfixest/did/visualize.py
```diff
@@ -72,26 +93,132 @@ def panelview(
...
+        if units_to_plot:
```
s3alfisc, line 111: For the `units_to_plot` argument, I would say that we should not allow subsampling and aggregation, right? Because if a user specifies specific units, does it make sense to still aggregate them / to subsample from them? (I think you could maybe argue for the first?). Still I would rather restrict this behavior.

pyfixest/did/visualize.py
```diff
@@ -72,26 +93,132 @@ def panelview(
...
+            data = data[data[unit].isin(units_to_plot)]
+        data_pivot = data.pivot(index=unit, columns=time, values=outcome)
+        if subsamp:
+            data_pivot = data_pivot.sample(subsamp)
+        if collapse_to_cohort:
+
+            def get_treatment_start(x: pd.DataFrame) -> pd.Timestamp:
```
s3alfisc, line 118: Maybe we can be more explicit here? E.g. change to 

```python
treat_bool = x[treat]
return x.iloc[treat_bool, time].min()
```

s3alfisc:
Hi @rafimikail , lots of confusion about this one, but the main outcome is that it's merged!

s3alfisc:
Thank you!

PR #586, opened 2024-08-23
overhaul mpl backend; fix coordinate flip argument

apoorvalal:
`coord_flip` did not work for the matplotlib backend [e.g. replacing all figs [here](https://py-econometrics.github.io/pyfixest/difference-in-differences.html) with MPL backends made everything unreadable]. 

I overhauled it to use `ax.errorbar`. Fixed(L) v old (R)

image

Also added an optional `ax` argument so that plots can be faceted mpl style

image

### DIFF
--- a/pyfixest/report/visualize.py
+++ b/pyfixest/report/visualize.py
@@ -19,7 +19,6 @@ from lets_plot import (
     theme,
     ylab,
 )
-from scipy.stats import norm
 
 from pyfixest.estimation.feiv_ import Feiv
 from pyfixest.estimation.feols_ import Feols
@@ -480,8 +479,9 @@ def _coefplot_matplotlib(
             lambda x: _relabel_expvar(x, labels, interactionSymbol)
         )
 
+    ub, lb = (f"{round(x * 100, 1)}%" for x in [1 - alpha / 2, alpha / 2])
+
     yintercept = yintercept if yintercept is not None else 0
-    critval = norm.ppf(1 - alpha / 2)
     title = title if title is not None else "Coefficient Plot"
 
     if ax is None:
@@ -509,13 +509,15 @@ def _coefplot_matplotlib(
             dodge_val = dodge_start + i * dodge
             x_pos = [coef_positions[coef] + dodge_val for coef in group["Coefficient"]]
         else:
-            x_pos = range(len(group))
+            x_pos = list(map(float, range(len(group))))
+
+        err = [group["Estimate"] - group[lb], group[ub] - group["Estimate"]]
 
         if flip_coord:
             ax.errorbar(
                 x=group["Estimate"],
                 y=x_pos,
-                xerr=group["Std. Error"] * critval,
+                xerr=err,
                 fmt="o",
                 capsize=5,
                 color=color,
@@ -525,7 +527,8 @@ def _coefplot_matplotlib(
             ax.errorbar(
                 y=group["Estimate"],
                 x=x_pos,
-                yerr=group["Std. Error"] * critval,
+                # yerr=group["Std. Error"] * critval,
+                yerr=err,
                 fmt="o",
                 capsize=5,
                 color=color,
--- a/pyfixest/report/visualize.py
+++ b/pyfixest/report/visualize.py
@@ -1,6 +1,7 @@
 from typing import Optional, Union
 
 import matplotlib.pyplot as plt
+import numpy as np
 import pandas as pd
 from lets_plot import (
     LetsPlot,
@@ -357,7 +358,7 @@ def _coefplot_lets_plot(
     title: Optional[str] = None,
     flip_coord: Optional[bool] = True,
     labels: Optional[dict] = None,
-    ax=None,
+    ax=None,  # for compatibility with matplotlib backend
 ):
     """
     Plot model coefficients with confidence intervals.
@@ -436,15 +437,17 @@ def _coefplot_matplotlib(
     flip_coord: Optional[bool] = True,
     labels: Optional[dict] = None,
     ax: Optional[plt.Axes] = None,
+    dodge: float = 0.5,
     **fig_kwargs,
 ) -> plt.Figure:
     """
-    Plot model coefficients with confidence intervals.
+    Plot model coefficients with confidence intervals, supporting multiple models.
 
     Parameters
     ----------
     df : pandas.DataFrame
         The dataframe containing the data used for the model fitting.
+        Must include a 'fml' column identifying different models.
     figsize : tuple
         The size of the figure.
     alpha : float
@@ -461,6 +464,8 @@ def _coefplot_matplotlib(
         Whether to flip the coordinates of the plot. Default is True.
     labels : dict, optional
         A dictionary to relabel the variables. The keys are the original variable names and the values the new names.
+    dodge : float, optional
+        The amount to dodge each model's points by. Default is 0.1.
     fig_kwargs : dict
         Additional keyword arguments to pass to the matplotlib figure.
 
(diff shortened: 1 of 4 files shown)


apoorvalal:
also the fact that the `Inf` interaction in the event study appears all the way to the right is quite disconcerting. Could we convert it into `-inf` @s3alfisc ?

apoorvalal:
Also was thinking about how to pair joint confints with `iplot`. 
Here's something I made with an annoying amount of manual work

```python
# interact treatment indicator with time dummies omitting the last period before treatment
m2 = pf.feols(
    """
              Y_it ~ i(time_id, ever_treated,  ref = 14) |
                unit_id + time_id
              """,
    df,
    vcov={"CRV1": "unit_id"},
)
# confidence regions
restable = m2.tidy()
timestamps = restable.index.str.extract(r'\[T\.(\d+)\]', expand=False).astype(int)
# create confidence envelopes for all pre-treat and all post-treat coefs
kv = [f"C(time_id, contr.treatment(base=14))[T.{t}]:ever_treated" for t in range(15)]
uniform_confint_pre = m2.confint(keep=kv, joint=True, exact_match=True)
kv = [f"C(time_id, contr.treatment(base=14))[T.{t}]:ever_treated" for t in range(15,30)]
uniform_confint_post = m2.confint(keep=kv, joint=True, exact_match=True)
uniform_res = pd.concat([uniform_confint_pre, uniform_confint_post])
uniform_res.columns = ['unif_25', 'unif_975']

f, ax = plt.subplots(1, 1, figsize=(12, 6))
ax.plot(
    np.arange(30),
    np.r_[np.repeat(0, 30 - 15), treat_effect_vector],
    marker=".",
    label="True ATE",
)
ax.errorbar(
    timestamps + .5,
    restable["Estimate"],
    yerr=restable["Std. Error"] * 1.96,
    fmt="o",
    label="pointwise 95% CI",
)
ax.errorbar(
    timestamps + .75,
    restable["Estimate"],
    yerr= [restable["Estimate"] - uniform_res["unif_25"], uniform_res["unif_975"] - restable["Estimate"]],
    fmt="o",
    label = "Joint 95% CI (pre/post)",
)
ax.axvline(x=14.5, color="k", linestyle="--")
ax.legend()
```

image


where the joint intervals are for all the pre-treatment coefficients at once (which is basically a test for parallel trends) and all the post-treatment coefficients at once (which is a test for the treatment effect **function**)

s3alfisc:
On the joint CIs: likely best to do a small code refactoring and move the core joint CI logic from the confint to the vcov method. This way, we can update the variance covariance matrix via `.vcov()`. 

If the used vcov matrix is not yet of type "simultaneous", we can then propagate it to `iplot()` by updating the `.tidy()` method with a `joint = True` argument, which is called by all plotting functions to get the coefficients, which in turn calls `.vcov()`, which calls `.get_inference()`. 

We can also have a flexible input argument `joint = "both"`, which would return a `Feols` instance with standard inference and a copy of the instance with simultaneous inference.

s3alfisc:
Hi, I have opened a PR that adds an argument to `iplot()` for joint CIs over at #587. Does not seem to work yet with matplotlib (I am really not an mlp expert).

s3alfisc:
> Could we convert it into -inf @s3alfisc ?

Yes, happy to do that =D

pyfixest/report/visualize.py
```diff
@@ -435,50 +439,57 @@ def _coefplot_matplotlib(
...
     """
     if labels is not None:
         interactionSymbol = " x "
         df["Coefficient"] = df["Coefficient"].apply(
             lambda x: _relabel_expvar(x, labels, interactionSymbol)
         )
+
     yintercept = yintercept if yintercept is not None else 0
-    ub, lb = alpha / 2, 1 - alpha / 2
+    critval = norm.ppf(1 - alpha / 2)
```
s3alfisc, line 452: Instead of computing the `critval` here, could we use the CIs returned via the `tidy()` method directly? Or would that not work with `ax.errorbar`?

s3alfisc:
Hi, generally happy to accept these changes! 

At the moment, the PR breaks multiple estimation functionality. Quick example: 

```python
import pyfixest as pf
from lets_plot import * 
LetsPlot.setup_html()

data = pf.get_data()
fit = pf.feols("Y + Y2~ i(f1)", data = data)

fit.iplot(coord_flip = False, plot_backend="matplotlib", figsize=[4, 3])
fit.iplot(coord_flip = False)
```
The `mlp` version returns 

image

but it should look as 

image

I like the option of being able to define subplots. Would this work with `seaborn` too?

s3alfisc:
Maybe not the most elegant, but we could simply to everything manually in `seaborn`? I did not manage to get any of the seaborn default args for flipping x- and y axis to work. 

```python
    if flip_coord:
        # Flip the x and y axes
        plot = (
            so.Plot(df, y="Estimate", x="Coefficient", color="fml")  # Swap x and y
            .add(so.Dot(), so.Dodge(empty="drop"))
            .add(
                so.Range(),
                so.Dodge(empty="drop"),
                ymin=str(round(lb * 100, 1)) + "%",  # Swap xmin with ymin
                ymax=str(round(ub * 100, 1)) + "%",  # Swap xmax with ymax
            )
            .label(
                title=title,
                y=rf"Estimate and {round((1-alpha)*100, 1)}% Confidence Interval",  # Swap x and y labels
                x="Coefficient",
                color="Model",
            )
            .on(ax)
        )
    else:
        plot = (
            so.Plot(df, x="Estimate", y="Coefficient", color="fml")
            .add(so.Dot(), so.Dodge(empty="drop"))
            .add(
                so.Range(),
                so.Dodge(empty="drop"),
                xmin=str(round(lb * 100, 1)) + "%",
                xmax=str(round(ub * 100, 1)) + "%",
            )
            .label(
                title=title,
                x=rf"Estimate and {round((1-alpha)*100, 1)}% Confidence Interval",
                y="Coefficient",
                color="Model",
            )
            .on(ax)
        )
```

pyfixest/report/visualize.py
apoorvalal: ax.errorbar expects an offset [so it computes UB and LB automatically under the hood], which is why I went with this. Alternatively, one can also pass a list as in the example above `
    yerr= [restable["Estimate"] - uniform_res["unif_25"], uniform_res["unif_975"] - restable["Estimate"]]`, which is a bit clunky.

apoorvalal:
I'm actually strongly in favour of getting rid of seaborn use entirely. 
MPL and letsplot(or other ggplot-likes, eg altair or plotnine) are two corner solutions: either roll your own plot with matrices and have very fine control over things (MPL: my preference) or have a user-friendly API that gets you 90% of the way there at the cost of finer control (eg `letsplot`). Seaborn is not very good at either, as illustrated here. Keeping the mpl backend as pure mpl for weirdos like me while most people use the default letsplot backend seems like a reasonable choice

s3alfisc:
Sounds good to me, let's switch to pure matplotlib then. I think the last PR I merged might need a slight refactor as we'll need the critical values for the joint cis, which is not available atm - though I think it could be easily added to the model_df frame that stores the coefficient and cis. Obviously some other larger refactors might also work =)

s3alfisc:
Btw I opted for letsplot back when because it's usually much faster than plotnine and looks prettier out of the box. I also like its  dynamic / html features. Have never used Altair though, just seen some nice plots produced by a colleague

apoorvalal:
fixed multiple estimation 
image

pyfixest/report/visualize.py
```diff
@@ -436,15 +437,17 @@ def _coefplot_matplotlib(
     flip_coord: Optional[bool] = True,
     labels: Optional[dict] = None,
     ax: Optional[plt.Axes] = None,
+    dodge: float = 0.5,
```
apoorvalal, line 440: hardcoded this for now since I didn't want to pass along many matplotlib-specific parameters from the `iplot` layer downwards. For sufficiently persnickety users (e.g. me), `tidy()` exists.

s3alfisc:
Pushed a small update to allow for joint CIs (& fixed a mypy complaint). Looks good to merge for me, though let me know if you'd prefer another solution for enabling joint= True, than the one I have now implemented.

s3alfisc (approved):

s3alfisc:
We also need to remove the seaborn dependency and replace it with matplotlib. Can be done via "poetry remove seaborn" and "poetry add matplotlib". I tried to do it myself but I'm in a train in the middle of nowhere in Germany and there is no WiFi around here 🤦‍♂️

s3alfisc:
Around Berlin now so wifi is back - updated the dependencies & created a new lock file. Ready to be merged from my end!

apoorvalal:
Oh yeah thx; I'll also update the mpl function to plot joint intervals. 
Also running into some issues with the buildfile (e.g. hardcodes powershell lol) so I'm glad you handled it.

s3alfisc:
Yeah the power shell is hard coded because I'm a windows guy 😀 I've implemented support for joint ci's btw! I can also give you a quick intro to the build tools one of these days, it's mostly very simple poetry + the just file commands (minus the power shell req)

pyproject.toml
```diff
@@ -16,11 +16,11 @@ numpy = ">=1.19.0"
 scipy = ">=1.6"
 formulaic = ">=1.0.0"
 lets-plot = ">=4.0.1"
-seaborn = ">=0.13.2"
```
apoorvalal, line 19: i think dropping this is breaking a handful of unit tests. no harm keeping it imo; esp since it also installs matplotlib automatically.

apoorvalal:
I ended up reverting the seaborn dependency drop commit since it produced a bunch of unit test failures (in corners of the package i know less well) btw. Will need a new patch to fully un-seaborn the package but I'm not in a hurry to do it since it is a fairly stable package that auto-installs matplotlib. You should go ahead with changes if you have strong opinions on the matter.

PR #587, opened 2024-08-24
Simultaneous CI option for iplot, coefplot

s3alfisc:
1. Adds a `joint` method to plot simultaneous CIs via the `iplot()` and `coefplot()` functions & methods (only for single methods). 
2. Renames the CI columns in `confint()` output to per cent, i.e. "0.025" becomes "2.5" (to standardize with the output of `tidy`). 
3. Updates the did vignette. 
4. Lots of linting from past PRs + etable now returns None at the end to accommodate mypy.
5. Working for mlp and lets_plot versions.

Relates to #586, fyi @apoorvalal


s3alfisc:
image

apoorvalal:
Very nice! I was thinking about contributing this too.
I computed coverage rates and find that only the joint method has nominal coverage here. 

https://gist.github.com/apoorvalal/8a7687d3620577fd5214f1d43fc740b3

s3alfisc:
Yes, this would be very cool to have in a vignette! Would you add it to the did/event study notebook or should we add a dedicated notebook on simultaneous inference correction where we show you simulation + also illustrate the rwolf & bonfetroni functions?

apoorvalal (approved):

PR #589, opened 2024-08-25
overhaul DID doc

apoorvalal:
Adds section on simultaneous vs pointwise inference in event studies, uses `resources` to avoid download, adds a separate file with some panel DGPs including the sharkfin used above.

### DIFF
--- a/pyfixest/utils/utils.py
+++ b/pyfixest/utils/utils.py
@@ -287,35 +287,6 @@ def get_data(N=1000, seed=1234, beta_type="1", error_type="1", model="Feols"):
     return df
 
 
-def get_blw():
-    """DGP for effect heterogeneity in panel data from Baker, Larcker, and Wang (2022)."""
-    n = np.arange(1, 31)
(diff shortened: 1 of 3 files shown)


pyfixest/utils/utils.py
```diff
@@ -354,42 +325,3 @@ def msqrt(C: np.ndarray) -> np.ndarray:
     return np.quantile(tmaxs, 1 - alpha)
 
 
-def _select_order_coefs(res: pd.DataFrame, keep: list, drop: list):
```
apoorvalal, line 357: this was added in my old simultaneous confint PR but eventually cloned to `dev_utils` so this was a duplicate afaict. Couldn't find anywhere this was used, but unit tests might say otherwise.

s3alfisc: Yes, it's a duplicate. Good catch!

s3alfisc (approved):
Looks great. Thanks @apoorvalal!

Issue #605, opened 2024-09-03
CI: Run Tests only on earliest and latest supported Python Versions & Parallelize CI Tests


s3alfisc:
CI tests currently run for about 1h, which is far too long for quick development. 

Following examples by [PyMC marketing](https://github.com/pymc-labs/pymc-marketing), we'd like to run all test only for the minimum and maximum supported Python versions (3.9 and 3.12). 

Additionally, we would like to split and parallelize the test suite, also following the [PyMC-Marketing example](https://github.com/pymc-labs/pymc-marketing/blob/main/.github/workflows/test.yml#L45). I suggest the following split: 

- Long running tests are all tests in [test_vs_fixest.py](https://github.com/py-econometrics/pyfixest/blob/master/tests/test_vs_fixest.py) and all tests marked with the `pytest.mark.extended` flag. 
- All other tests run relatively quickly. 

See also [here ](https://guicommits.com/parallelize-pytest-tests-github-actions/) for guidance on how to parallelize CI tests.

fyi @juanitorduz

PR #607, opened 2024-09-03
Try to split and paralelize tests

juanitorduz:
Closes https://github.com/py-econometrics/pyfixest/issues/605


.github/workflows/ci-tests.yaml
```diff
@@ -10,13 +10,13 @@ on:
...
   test:
-    name: "Tests"
+    name: "Tests Core"
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.12"]
```
juanitorduz, line 19: For the core tests its enough to test the min and max versions

.github/workflows/ci-tests.yaml
```diff
@@ -49,6 +49,46 @@ jobs:
...
         uses: codecov/codecov-action@v4
 
+  test_slow:
+    name: "Test Slow"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        python-version: ["3.12"]
```
juanitorduz, line 59: I suggest we just use the latest version for the slow tests. WDYT?

s3alfisc: Yes, sounds good to me!

s3alfisc (approved):

juanitorduz:
ok! Apologies for the random commits above, I was not getting why some slow tests where being shipped but I just realized there are kip statements inside the tests like 

```python
        pytest.skip(
            "Cluster adjustment only works with cluster inference. Nothing to test here."
        )
```
😅

pyproject.toml
```diff
@@ -43,19 +46,20 @@ watermark = "^2.4.3"
...
+  "--strict-markers",
+  "--strict-config",
+  "--cov=pyfixest",
+  "--cov-report=term-missing",
+  "--color=yes",
 ]
 markers = [
-    "extended: mark test to be part of the extended test suite"
+  "slow: mark test to be part of the slow test suite",
+  "extended: mark test to be part of the extended test suite",
```
juanitorduz, line 58: In the next iteration, we can split further by adding more markets if necessary.

s3alfisc:
Yep, this here 

```
    if "f3" not in fml:
pytest.skip("No f3 in formula. Nothing to test here.")
```
was the statement that deactivated all IV tests ...

Issue #606, opened 2024-09-03
CI: Add workflow to automatically render all Example Notebooks


s3alfisc:
To avoid any breaking API changes, we would like to add a github action to render all example notebooks. 

In more detail, we'd like to render the following notebooks: 

- [compare_fixest_pyfixest.ipynb](https://github.com/py-econometrics/pyfixest/blob/master/docs/compare-fixest-pyfixest.ipynb)
- [Difference-in-Differences Estimation](https://github.com/py-econometrics/pyfixest/blob/master/docs/difference-in-differences.ipynb)
- [Marginaleffects](https://github.com/py-econometrics/pyfixest/blob/master/docs/marginaleffects.ipynb)
- [Quickstart](https://github.com/py-econometrics/pyfixest/blob/master/docs/quickstart.ipynb)
- [Replicating the Effect](https://github.com/py-econometrics/pyfixest/blob/master/docs/replicating-the-effect.ipynb)
- [Regression Tables / Etable](https://github.com/py-econometrics/pyfixest/blob/master/docs/stargazer.ipynb)

@juanitorduz fyi =)

PR #613, opened 2024-09-04
test run example notebooks

juanitorduz:
Closes https://github.com/py-econometrics/pyfixest/issues/606


scripts/run_notebooks.py
```diff
@@ -0,0 +1,55 @@
...
+
+KERNEL_NAME: str = "python3"
+DOCS = Path("docs")
+NOTEBOOKS: list[Path] = [
+    # DOCS / "compare-fixest-pyfixest.ipynb",  # needs R
+    DOCS / "difference-in-differences.ipynb",
+    DOCS / "marginaleffects.ipynb",
+    DOCS / "quickstart.ipynb",
+    DOCS / "replicating-the-effect.ipynb",
+    DOCS / "stargazer.ipynb",  # failing notebook
```
juanitorduz, line 21: @s3alfisc  in Cell[22] there is an error in the notebook (checked locally)

```python
NotImplementedError: <class 'pyfixest.estimation.feols_.Feols'>
```

See https://github.com/py-econometrics/pyfixest/actions/runs/10707033524/job/29686144381

scripts/run_notebooks.py
```diff
@@ -0,0 +1,55 @@
...
+from pathlib import Path
+
+import papermill
+from joblib import Parallel, delayed
+from tqdm import tqdm
+
+KERNEL_NAME: str = "python3"
+DOCS = Path("docs")
+NOTEBOOKS: list[Path] = [
+    # DOCS / "compare-fixest-pyfixest.ipynb",  # needs R
```
juanitorduz, line 16: Let me work on this notebook on a different PR as it requires some investigation.

juanitorduz:
@s3alfisc this is ready for review. This set up the tests. I commented two notebooks:

- One needs investigation (I need to install R locally hehe)
- The other notebook is failing because of `pyfixest` I think https://github.com/py-econometrics/pyfixest/pull/613#discussion_r1744227154

scripts/run_notebooks.py
s3alfisc: The error likely stems from the fact that Stargazer's default does not have pyfixest support - makes sense to handle this in another pr!

juanitorduz:
ok! so any suggested changes or are we good to go @s3alfisc ?  :)

s3alfisc (approved):
Looks good to me, will merge now. Sorry that I did not finish the PR review yesterday, I got interrupted and then did not manage to get back to it. Btw I think you are successfully installing R via r2u, so the notebook depending on rpy2/fixest should execute properly =) thank you!

PR #614, opened 2024-09-06
Add dtables and make_tables functions. Small corrections in etable. Update of ipynb notebooks.

dsliwka:
Added two functions to display tables: 
- pf.dtable shows descriptive statistics. 
- pf.make_table displays dataframes as great_tables or latex booktab. 
- Now use make_table also for the "gt" and "tex" output in etable to allow unified layouts. 
- Output type "df" now gives back an unstyled df. 

Added a notebook "table-layout.ipynb" to which I moved the etable examples from the stargazer notebook and added a documentation of dtable and make_table as well as some hints on how to generate and use table styles for "gt" output. This is not yet linked to in the _quarto.yaml.
 
Further small changes:
- There was a small bug as etable accepts a FixestMulti object but then produces an error message. FixestMulti is now converted into a list of models so that etable now accepts FixestMulti objects without requiring the user to convert them into a list before calling etable.
- etable now moves the "Intercept" row to the end of the table as this seems to be standard practice in papers (so that the reader focuses on the first explanatory variable which is typically the variable of interest).
- Updated the ipynb notebooks to new etables functionality.


s3alfisc:
@pre-commit.ci autofix

s3alfisc:
@pre-commit.ci autofix

s3alfisc:
pre-commit.ci autofix

s3alfisc:
Hi @dsliwka , I just figured out how to run the auto-linter! There are a few style issues that `ruff` complains about + the `mypy` static type checker is unhappy with a few things as well. Do you want to try to fix this yourself? Else I will give this a go =) We can also join forces! 

For example, there are a few errors like this one: 

> pyfixest\report\summarize.py:458: error: Argument "notes" to "make_table" has incompatible type "str | None"; expected "str"  [arg-type]

which can be fixed by changing 

```python 
    notes: str = None,
```
to 
```python
    notes: str = Optional[None],
```
.

s3alfisc:
Slight side point - I was thinking about potentially writing some tests for the functionality in the `report` module. Do you have an idea on how to potentially tackle this? My initial though was to test that the input data frame to `make_table` looks the way it is supposed to.

s3alfisc:
pre-commit.ci autofix

PR #619, opened 2024-09-11
Add Support for (in-memory) Regression compression a la `duckreg`

s3alfisc:
Supersedes #574 .

- Adds a class for compressed regression, `FeolsCompressed`.
- Adds `use_compression` argument to `pf.feols()` that supports up to two-way fixed effects via the Mundlak transform, which is run automatically when users opt into fixed effects syntax. 
- Supported inference: iid, hetero. When Mundlak is used, only wild cluster bootstrap inference is supported. Cluster variables need to be part of the model. 
- No prediction method is supported yet. 

```python
%load_ext autoreload
%autoreload 2

import pyfixest as pf
from pyfixest.estimation.estimation import feols
import pytest
import pandas as pd
import numpy as np

data = pf.get_data()

fit = pf.feols("Y ~ X1 + X2| f1", data=data, vcov = "hetero", ssc = pf.ssc(adj = False, cluster_adj = False))
fit_c = pf.feols("Y ~ X1 + X2 + C(f1)", data = data, use_compression=True, vcov = "hetero", ssc = pf.ssc(adj = False, cluster_adj = False))

pf.etable([fit, fit_c], keep = ["X1", "X2"], digits = 6)
```
image

FYI @apoorvalal


s3alfisc:
Implementation: 

For compressed estimation, fixed effects are always handled via the Mundlak transform. Up to two-way fixed effects are supported. There are no checks if the input data is a panel when using two fixed effects. Maybe we should have one? 

Unit tests are implemented and pass. 

Currently not supported: if a cluster variable is not part of the model formula.

s3alfisc:
@pre-commit.ci autofix

s3alfisc:
Something appears off with Mundlak transform iid & hetero SEs.

s3alfisc:
@pre-commit.ci autofix

s3alfisc:
pre-commit.ci autofix

PR #625, opened 2024-09-16
Improve CI Run-Time: Test Smarter

s3alfisc:
Multiple Changes to CI Tests: 

- in `test_vs_fixest.py`, only test arguments `adj`, `cluster_adj`, `dropna=True` for `feols`, not for fepois & feiv. 
- move the `test_ritest.py` , `test_multcomp.py, `test_plots.py` to the extended marker
- runs all extended tests only for commits to master and if `extended-tests` label is set


s3alfisc:
tests-plots

PR #637, opened 2024-09-27
Compact tables in dtable & bug fix for FE display in etable

dsliwka:
Added a feature in dtable to display compact tables with mean and std in one cell
Bug fix in etable removing "0" line when no fixed effects in models


s3alfisc:
Looks great, thank you Dirk!

PR #654, opened 2024-10-14
Adjusting Figsize of Outcome Plot

rafimikail:


rafimikail:
Hi @s3alfisc , this will be the PR to readjust the size of outcome plot also maybe to check whether all the dependencies in pixi docs env is going to be installed automatically, i will let you know once its successfully tested locally

s3alfisc:
Strangely all docs build successfully, with the status win lock file. Does not really match your experience, right?

PR #665, opened 2024-10-22
Bug fix to display default table note in etable

dsliwka:
Correct a bug in etable causing that the default table notes are not displayed when the user specifies no custom notes.


s3alfisc:
Thanks @dsliwka!

PR #669, opened 2024-10-23
Gelbach Decomposition

s3alfisc:
Initial implementation of Gelbach's decomposition: 

```python
%load_ext autoreload
%autoreload 2

import pyfixest as pf

data = pf.get_data(N = 100_000)
fit = pf.feols("Y ~ X1 + X2 | f1", data = data)

fit.decompose(param = "X1")
```
@apoorvalal , any thoughts on naming the method? Should it be `decompose` - which would allow to implement other decomposition methods around it - or `mediate`? In labour-land, the method is known as "Gelbach decomposition" while I think in the causal word, it would rather be know as mediation analysis?


s3alfisc:
Added an aggregation argument. If `False`, impacts of individual mediators are returned: 

```python
%load_ext autoreload
%autoreload 2

import pyfixest as pf

data = pf.get_data(N = 10_000)
fit = pf.feols("Y ~ X1 + X2 + f1 ", data = data)

fit.decompose(param = "X1", reps = 10, agg = False)

#	Estimate	CI Lower (0.025)	CI Upper (0.975)
# Total Effect:	5.603648e-01	5.580088e-01	5.837112e-01
# Mediated Effect: Intercept	4.490424e-05	4.401803e-04	1.366609e-03
# Mediated Effect: X2	5.603938e-01	5.586394e-01	5.940223e-01
# Mediated Effect: f1	-7.392033e-05	-1.124425e-02	-9.606237e-04
# Direct Effect: X1	-1.110223e-16	-1.171285e-15	1.243450e-15
```

pyfixest/estimation/decomposition.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def __init__(
+        self,
+        agg: bool,
+        param: str,
+        coefnames: list[str],
+        nthreads: int = -1,
+        cluster_df: Optional[pd.Series] = None,
+    ):
+        self.cluster_dict: Optional[dict[Any, Any]]
```
apoorvalal, line 25: ~17 lines of attaching objects to the class here (this is true for almost all estimator classes). Any downsides to making estimators [dataclasses](https://docs.python.org/3/library/dataclasses.html) ?

s3alfisc: So in principle, the idea would be to change to something like this: 

```python
from dataclasses import dataclass, field, InitVar
from typing import Any, Optional, List, Dict
import numpy as np
import pandas as pd

@dataclass
class GelbachDecomposition:
    """
    Linear Mediation Model.

    Initial implementation by Apoorva Lal at
    https://gist.github.com/apoorvalal/e7dc9f3e52dcd9d51854b28b3e8a7ba4.
    """
    
    agg: bool
    param: str
    coefnames: List[str]
    nthreads: int = -1
    cluster_df: Optional[pd.Series] = None
    
    # Define attributes initialized post-creation
    cluster_dict: Optional[Dict[Any, Any]] = field(init=False, default=None)
    unique_clusters: Optional[np.ndarray] = field(init=False, default=None)
    mask: np.ndarray = field(init=False)
    mediator_names: List[str] = field(init=False)
    X_dict: Dict[Any, Any] = field(init=False, default_factory=dict)
    X1_dict: Dict[Any, Any] = field(init=False, default_factory=dict)
    X2_dict: Dict[Any, Any] = field(init=False, default_factory=dict)
    Y_dict: Dict[Any, Any] = field(init=False, default_factory=dict)

    def __post_init__(self):
        # Set up mask to exclude 'param' from coefnames
        param_idx = self.coefnames.index(self.param)
        self.mask = np.ones(len(self.coefnames), dtype=bool)
        self.mask[param_idx] = False

        # Get mediator names excluding 'param'
        self.mediator_names = [name for name in self.coefnames if self.param not in name]

        # Handle clustering setup if cluster_df is provided
        if self.cluster_df is not None:
            self.unique_clusters = self.cluster_df.unique()
            self.cluster_dict = {
                cluster: self.cluster_df[self.cluster_df == cluster].index
                for cluster in self.unique_clusters
            }
        else:
            self.unique_clusters = None
            self.cluster_dict = None

        # Additional initialization for dictionaries can be added as needed

```
The main advantage is that there is a separation between `__init__` and processing of inputs in `__post__init__` which leads to cleaner code (which I like) but it does not really help to economize on lines of code, right? 

Generally I like it!

s3alfisc:
I have slightly adjusted your code @apoorvalal as I don't think it was fully in line with Gelbach's method, but maybe I am missing something?

In Gelbach's paper, we start out with a linear model 

$$
Y = beta_1 X_1 + beta_2 X_2 + e
$$ 

The algorithm then states to 
- obtain $beta_2$ by fitting the **full** regression model 
- obtain Gamma by fitting $X_2$ on $X_1$ for each covariate K in $X_2$
- last, we can compute $delta_k = Gamma_k * beta_{2,k}$ and $delta = \sum delta_k$. 

[Reference ](https://typeset.io/pdf/when-do-covariates-matter-and-which-ones-and-how-much-5gt21vh0t7.pdf) here, see the attached screenshot. 
image

Whereas you implemented something slightly different: 
```python
            self.beta_tilde = np.linalg.lstsq(X, y, rcond=1)[0]
            self.delta_tilde = np.linalg.lstsq(X, W, rcond=1)[0]
            self.gamma_tilde = np.linalg.lstsq(W, y, rcond=1)[0]
            self.total_effect, self.mediated_effect = self.beta_tilde, self.delta_tilde @ self.gamma_tilde
            self.direct_effect = self.total_effect - self.mediated_effect
```
- $X$ here seem to be all treatment variables, equivalent to $X_1$ in Gelbach's notation
- $W$ are the mediators -> $X_2$
- beta_tilde hence computes beta in the **short** regression `Y on X_1`
- Most importantly, gamma_tilde computed as above <> beta_2 from the long regression

Am I missing something?

pyfixest/estimation/decomposition.py
s3alfisc: Pushed something like the above.

apoorvalal:
oh yeah that's a mistake now that I look at it; the long regression is never estimated. The regression that estimates `gamma` should actually be regressing `Y` on `np.c_[W, X]` (which would then give correct mediated and direct effects) and not just `Y` on `W`. The total and first stage effects (`beta` and `delta`) are fine. 

Then, the equivalence should hold exactly: 
- Jonah's $\beta_1^{\text{base}}$ is my `beta` (total effect)
- Jonah's $\Gamma$ (projection of $X_2$ on $X_1$ is my `delta` 
- Jonah's $\beta_2$ (coef on mediator in long regression) is my `gamma` 

so when he writes $\delta = \Gamma \beta_2$, that is exactly analogous to the NIE from mediation analysis. 


relevant reference from Peng's book [potentially confusing because he has different notation too] ; [code](https://github.com/apoorvalal/ding_causalInference_python/blob/master/Chapter27mediationAnalysis.ipynb)
image

s3alfisc:
Now reproducing results from [b1x2 docs](http://fmwww.bc.edu/repec/bocode/b/b1x2.sthlp): 

Stata: 

image


Python: 

```python
%load_ext autoreload
%autoreload 2

import pandas as pd
import numpy as np
import pyfixest as pf

def gelbach_test_data(): 

    # Set a random seed for reproducibility (optional)
    np.random.seed(0)

    # Create a DataFrame with 1000 observations
    df = pd.DataFrame(index=range(1000))

    # Generate x1 as a standard normal random variable
    df['x1'] = np.random.normal(size=1000)

    # Generate x21 = x1 * 1 + standard normal random variable
    df['x21'] = df['x1'] * 1 + np.random.normal(size=1000)

    # Generate x22 = x1 * 0.25 + x21 * 0.75 + standard normal random variable
    df['x22'] = df['x1'] * 0.25 + df['x21'] * 0.75 + np.random.normal(size=1000)

    # Generate x23 = x1 * 0.4 + x21 * 0.6 + x22 * 0.4 + standard normal random variable
    df['x23'] = (df['x1'] * 0.4 +
                df['x21'] * 0.6 +
                df['x22'] * 0.4 +
                np.random.normal(size=1000))

    # Generate y = x1 * 1 + x21 * 2 + x22 * 0.5 + x23 * 0.75 + standard normal random variable
    df['y'] = (df['x1'] * 1 +
            df['x21'] * 2 +
            df['x22'] * 0.5 +
            df['x23'] * 0.75 +
            np.random.normal(size=1000))

    return df

data = gelbach_test_data()
fit = pf.feols("y ~ x1 + x21 + x22 + x23", data=data)
res = fit.decompose(param = "x1", combine_covariates={"g1": ["x21", "x22"], "g2": ["x23"]})
pf.make_table(res)
```
image

To do's: 

- [ ] tests
- [ ] linter & type checking
- [ ] optional for this PR: aggregation, then decomposition (faster for hdfe)

pyfixest/estimation/decomposition.py
```diff
@@ -0,0 +1,269 @@
...
+
+            direct_effect = np.linalg.lstsq(X1, Y, rcond=None)[0].flatten()[
+                self.param_in_X1_idx
+            ]
+            direct_effect = np.array([direct_effect])
+
+            gamma = np.linalg.lstsq(X1[:, 1:], X2, rcond=None)[0].flatten()
+            beta_full = np.linalg.lstsq(X, Y, rcond=None)[0].flatten()
+            beta2 = beta_full[self.mask].flatten()
+            delta = gamma * beta2
```
apoorvalal, line 165: what's the difference between the mediation setup and gelbach's method setup? I think you run the same regressions?

s3alfisc: Yes I think both are identical (though I should take another look at the Ding chapter you mentioned earlier). I was thinking about adding yet another method, `mediate()`, which just runs `decompose()` under the hood? Might help to "un-confuse" users despite being the same thing? I.e. I would add 

```python 
def mediate(params): 
    self.decompose(params)
```
?

apoorvalal: Sure that seems fine (although strictly speaking decompose is a subset of mediate :) ). Just wanted to make sure i wasn't missing something subtle.

apoorvalal:
How do you want to handle FEs here? One idea is to dummy-code it like any other covariate and use sparse solvers (since we're bootstrapping the whole thing, the vcov isn't important)

s3alfisc:
> How do you want to handle FEs here? One idea is to dummy-code it like any other covariate and use sparse solvers (since we're bootstrapping the whole thing, the vcov isn't important)

Yeah that's a good suggestion. Basically we could do as in the [wildboottest method](https://github.com/py-econometrics/pyfixest/blob/d0c75d37ee5a739799ec34c887e730fd24554e01/pyfixest/estimation/feols_.py#L1205). 

I'll make sure to match stata first and then I'll get to that - currently I am producing annoyingly small differences at the 4th digit behind the comma 🤔 

```
FAILED tests/test_decomposition.py::test_against_stata - AssertionError: Failed for g1 with values 2.432692 and [2.43237097]
```

pyfixest/estimation/decomposition.py
```diff
@@ -126,14 +125,22 @@ def fit(self, X: np.ndarray, Y: np.ndarray, store: bool = True):
...
+                for g in self.unique_clusters:
+                    cluster_idx = np.where(self.cluster_df == g)[0]
+                    self.X_dict[g] = self.X[cluster_idx]
+                    self.Y_dict[g] = self.Y[cluster_idx]
+
             return self.contribution_dict
 
         else:
-            X1 = np.concatenate([np.ones((self.N, 1)), X[:, ~self.mask]], axis=1)
+            X1 = np.concatenate([np.ones((X.shape[0], 1)), X[:, ~self.mask]], axis=1)
```
apoorvalal, line 143: was this the source of the error? i don't understand why the bootstrap loop would require a different data matrix from the point estimation flow

s3alfisc: No, that wasn't it - I just pushed the bug fix with [accf066](https://github.com/py-econometrics/pyfixest/pull/669/commits/accf066c6999606c47d41c18c2320f1aaf07c328). Basically dropped an intercept when I shouldn't have.

s3alfisc: To save you some time the issue was to do 

```python
            delta = np.linalg.lstsq(X1[:, 1:], Hg, rcond=None)[0].flatten()
```
instead of 
```python
            delta = np.linalg.lstsq(X1, Hg, rcond=None)[0].flatten()
```

s3alfisc: > i don't understand why the bootstrap loop would require a different data matrix from the point estimation flow
I actually think this is needed, but will take another look tomorrow. Too late in the evening over here to think this through ...

s3alfisc: Actually you're right and adjusted with [43b8e79](https://github.com/py-econometrics/pyfixest/pull/669/commits/43b8e79ae46717b1b2b674688fc9a023ca63c93b). Thanks!

s3alfisc: I actually had to revert it - the reason is that bootstrapped X and Y are passed to fit and X1 and X2 need to be created from the bootstrapped X1.

s3alfisc:
I think this should more or less cover everything @apoorvalal. Maybe the only thing left to do is to write a nice vignette / case study, but that can happen in a separate PR.

pyfixest/estimation/decomposition.py
```diff
@@ -0,0 +1,390 @@
...
+                    raise ValueError(f"{overlap} is in both {key1} and {key2}.")
+
+    def fit(self, X: np.ndarray, Y: np.ndarray, store: bool = True):
+        "Fit Linear Mediation Model."
+        if store:
+            self.X = X
+            self.N = X.shape[0]
+
+            self.X1 = self.X[:, ~self.mask]
+            self.X1 = np.concatenate([np.ones((self.N, 1)), self.X1], axis=1)
```
apoorvalal, line 102: q (for my own curiosity): any benefits to `np.concatenate` over `np.vstack` or even more concise `np.c_[arr1, arr2]` ? I use the latter

pyfixest/estimation/decomposition.py
```diff
@@ -0,0 +1,390 @@
...
+            return contribution_dict
+
+    def bootstrap(self, rng: np.random.Generator, B: int = 1_000, alpha: float = 0.05):
+        "Bootstrap Confidence Intervals for Total, Mediated and Direct Effects."
+        self.alpha = alpha
+        self.B = B
+
+        if self.nthreads != 1:
+            try:
+                from joblib import Parallel, delayed
```
apoorvalal, line 161: add this as a dependency? joblib is pretty close to base (you can also use [multiprocessing](https://docs.python.org/3/library/multiprocessing.html) from the standard library although it is a bit more clunky).

apoorvalal (approved):
LGTM. Nice work.

pyfixest/estimation/decomposition.py
s3alfisc: Not that I am aware of, just habit =)

pyfixest/estimation/decomposition.py
s3alfisc: Yep, will do, a dependency with "no mandatory dependencies besides Python" sounds pretty reasonable to add =)

pyfixest/estimation/feols_.py
```diff
@@ -1478,6 +1467,165 @@ def ccv(
...
+        Compares the short model depvar on {param} with the long model
+        specified in the original feols() call. For details, take a look at
+        "When do covariates matter?" by Gelbach (2016, JoLe). You can find
+        an ungated version of the paper on SSRN under the following link:
+        https://papers.ssrn.com/sol3/papers.cfm?abstract_id=1425737 .
+
+        Parameters
+        ----------
+        param : str
+            The name of the parameter to decompose.
```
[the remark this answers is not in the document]
s3alfisc: Yes, thanks, I'll update this. I was generally planning to take a second look at the documentation & to even write a standalone vignette after getting this merge through.

Issue #468, opened 2024-05-30
Fepois: Add support for `predict` with `newdata` argument


s3alfisc:
## Context

Currently, the `newdata` argument is disabled for `Fepois.predict`. Unfortunately, I do not recall why... 😅 

## To do

- Enable the `newdata` argument for `Fepois.predict(). 
- Clean up ValueError when trying to run Fepois.predict() with `newdata` in `tests/test_errors.py`.
- Add unit tests to `tests/test_vs_r::test_single_fit`.

leostimpfle:
Having had a look at #467 I think both this issue and #467 can be resolved in a single update.

I think it would be easiest to generalise the `predict` method of `feols` to always produce the linear predictor (`link`) and only handle the conversion to the non-linear predictor (`response`) in `fepois`.

s3alfisc:
Sounds like a good plan! I agree, it makes sense to only define the `response` prediction in `Fepois`. Should I then assign this issue to you as well?

leostimpfle:
Yes, happy to take it on!

s3alfisc:
This can be close right @leostimpfle ?

s3alfisc:
Ah no, actually not. Still throws an error. But in principle we should be able to support it now I think

leostimpfle:
Yes, it should work without fixed effects. After your recent bugfix in the calculation of fixed effects, it will hopefully straightforward to support predict with both newdata and fixed effects. I should have some time this weekend to look into it

Issue #467, opened 2024-05-30
Fepois: Add support for `predict` with fixed effects


s3alfisc:
## Context

Currently, the `predict` method is disabled for Poisson Regression with fixed effects. Unfortunately, I cannot recall why... 😅 

## To do

- Enable `predict()` for Poisson regression with fixed effects. 
- Clean up ValueError when trying to run Fepois.predict() on a fixed effect model in `tests/test_errors.py`.
- Add unit tests to `tests/test_vs_r::test_single_fit`.

leostimpfle:
While I haven't completed implementation of all separation checks yet, I would like to pre-register my interest for working on this issue 🙂

s3alfisc:
More than happy to already assign this to you Leo! =)

PR #670, opened 2024-10-25
Predict

leostimpfle:
This PR is a small preliminary step for issues #467 and #468. It has two updates so far:

- make `_Y_hat_link` and `_Y_hat_response` consistent between `Feols`, `Feiv`, and `Fepois`: both attributes are now the predictors based on the regressors before demeaning (previously, `Fepois` used the regressors before demeaning while `Feols` and `Feiv` used the regressors after demeaning)
- update `Feols.predict` to use `_Y_hat_link` and `_Y_hat_response` if `newdata=None` and update `Fepois.predict` to simply call `super().predict` (which is possible because of the first step)

I will continue to update this PR to implement `Fepois.predict` with `newdata` and fixed effects.

### DIFF
--- a/pyfixest/estimation/feols_.py
+++ b/pyfixest/estimation/feols_.py
@@ -2,7 +2,7 @@ import functools
 import gc
 import warnings
 from importlib import import_module
-from typing import Optional, Union, Literal, get_args
+from typing import Literal, Optional, Union, get_args
 
 import numba as nb
 import numpy as np
@@ -452,6 +452,11 @@ class Feols:
         else:
             raise ValueError(f"Solver {solver} not supported.")
 
+    def _get_residuals_and_predictors(self) -> None:
+        self._u_hat = self._Y.flatten() - (self._X @ self._beta_hat).flatten()
+        self._Y_hat_link = self._Y_untransformed.values.flatten() - self.resid()
+        self._Y_hat_response = self._Y_hat_link
+
     def get_fit(self) -> None:
         """
         Fit an OLS model.
@@ -473,9 +478,7 @@ class Feols:
 
             self._beta_hat = self.solve_ols(self._tZX, self._tZy, _solver)
 
-            self._u_hat = self._Y.flatten() -  (self._X @ self._beta_hat).flatten()
-            self._Y_hat_link = self._Y_untransformed.values.flatten() - self.resid()
-            self._Y_hat_response = self._Y_hat_link
+            self._get_residuals_and_predictors()
 
             self._scores = _X * self._u_hat[:, None]
             self._hessian = self._tZX.copy()
--- a/pyfixest/estimation/feols_.py
+++ b/pyfixest/estimation/feols_.py
@@ -452,8 +452,7 @@ class Feols:
         else:
             raise ValueError(f"Solver {solver} not supported.")
 
-    def _get_residuals_and_predictors(self) -> None:
-        self._u_hat = self._Y.flatten() - (self._X @ self._beta_hat).flatten()
+    def _get_predictors(self) -> None:
         self._Y_hat_link = self._Y_untransformed.values.flatten() - self.resid()
         self._Y_hat_response = self._Y_hat_link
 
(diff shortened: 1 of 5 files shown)


pyfixest/estimation/feols_.py
```diff
@@ -40,6 +40,7 @@
 )
 from pyfixest.utils.utils import get_ssc, simultaneous_crit_val
 
+prediction_type = Literal["response", "link"]
```
s3alfisc, line 43: Very nice, TIL! We should probably have this everywhere in the code base

pyfixest/estimation/fepois_.py
```diff
@@ -387,32 +387,20 @@ def predict(
...
+        if self._has_fixef:
             raise NotImplementedError(
                 "Prediction with fixed effects is not yet implemented for Poisson regression."
             )
         if newdata is not None:
             raise NotImplementedError(
                 "Prediction with function argument `newdata` is not yet implemented for Poisson regression."
             )
-
-        # y_hat = super().predict(newdata=newdata, type=type, atol=atol, btol=btol)
```
s3alfisc, line 409: Where in the new version are we conducting the exponentiation for for the Poisson prediction for type = "response"?

s3alfisc: Ah, I see, we do it in the fit method already.

s3alfisc (approved):
Looks good, thanks @leostimpfle! If you like, you can merge this already; or you set up new PRs and we merge the new ones into this one, step by step?

s3alfisc:
pre-commit.ci autofix

pyfixest/estimation/feols_compressed_.py
```diff
@@ -184,6 +184,8 @@ def prepare_model_matrix(self):
         # overwrite Y, X, _data
         self._data_long = data_long_mundlak if self._use_mundlak else data_long
         self._Yd = compressed_dict.Y.to_pandas()
+        # store compressed dependent variable before demeaning
+        self._Y_untransformed = self._Yd.copy()
```
s3alfisc, line 188: This is only valid because the compression algo does not demean, right?

pyfixest/estimation/feols_.py
leostimpfle: Was thinking the same! Maybe even have a separate module `literals.py` to store all literals centrally in the repository?

pyfixest/estimation/feols_compressed_.py
leostimpfle: My idea was that `prepare_model_matrix` (where `_Y_untransformed` is set) will always be called before demean. This is also the case for `feols`: https://github.com/py-econometrics/pyfixest/blob/8435537ade9ce3e02702c4f63603e9801cc46afe/pyfixest/estimation/FixestMulti_.py#L281-L282

The reason that I had to add this line for `feols_compressed` is that the call to `super().prepare_model_matrix` stores `_Y_untransformed` before compression whereas we need the dependent variable before demeaning but after compression for prediction.

leostimpfle:
> Looks good, thanks @leostimpfle! If you like, you can merge this already; or you set up new PRs and we merge the new ones into this one, step by step?

I think it would be good to merge this first step already to reduce the risk of merge conflicts later. For me the main question is if the update to `FeolsCompressed._Y_untransformed` is [ok or not ](https://github.com/py-econometrics/pyfixest/pull/670#discussion_r1817257140). I'll already merge this branch once we're happy with this point as well.

pyfixest/estimation/feols_compressed_.py
s3alfisc: I think this should be ok. We need to assign Y_untransformed via the "long" / "uncompressed" dependent variable because we want to make predictions based on the uncompressed variable. Hence if we call `.predict()` on the compressed matrix, it will use Y_untransformed when `newdata = None`.

pyfixest/estimation/feols_.py
s3alfisc: Opened a PR to do this =)

s3alfisc:
Yes, the changes in the `FeolsCompressed` class are fine. I think the only reason to store the "long" Y_untransformed is to enable predictions on the "long" / "uncompressed" data set. But predictions are currently disabled for compressed regression - they trigger a `NotImplementedError`. So fine for me to go ahead and merge this!

Issue #671, opened 2024-10-26
Create literals and move all literals to `literals.py` file


s3alfisc:
It would be nice if we consequently used string literals for type checks as in [here](https://github.com/py-econometrics/pyfixest/blob/bf6fa56e3b2bbf5f6a6a18b38069da0f13942124/pyfixest/estimation/feols_.py#L43). 

To do so, we would set up a dedicated file, `literals.py`, and move all "stringly-typed" function arguments to the file. 

We could start with the arguments to the [feols](https://github.com/py-econometrics/pyfixest/blob/8435537ade9ce3e02702c4f63603e9801cc46afe/pyfixest/estimation/estimation.py#L13) and `fepois` APIs, for which we have the following stringly-typed arguments: 

-  vcov : Union[str, dict[str, str]] -> here "str" part could be either "iid", "hetero", "hc1", "hc2", or "hc3"
- weights_type: either "aweights" or "fweights"
- fixef_rm: "singleton" or "none"
- solver: "np.linalg.lstsq" or "np.linalg.solve"

cc @leostimpfle 

Maybe of interest to @marcandre259 or @rafimikail ?

marcandre259:
I've starting working on #557 with the weights for did, it's an issue that's also interesting for my work, so motivating to look at. 

If @rafimikail or another contributor is actively working on #557 however, or once I get something done there, I would be happy to pick up this issue.

s3alfisc:
Ah, very cool! I don't think Rafi is currently working on #557 so I don't think there are any double efforts at the moment =)

PR #680, opened 2024-10-27
Adding literals to feols and fepois api's

marcandre259:
Reference to #671. Pretty literal interpretation. Added a small validate_literal_argument function, taken from the predict method of Feols (assumption is it may be used elsewhere also).

### DIFF
--- a/pyfixest/estimation/literals.py
+++ b/pyfixest/estimation/literals.py
@@ -7,7 +7,7 @@ fixef_rm_options = Literal["singleton", "none"]
 solver_options = Literal["np.linalg.solve", "np.linalg.lstsq"]
 
 
-def validate_literal_argument(arg: Any, literal: Literal) -> None:
+def validate_literal_argument(arg: Any, literal: Literal[str]) -> None:
     """
     Validate if the given argument matches one of the allowed literal types.
 
@@ -19,7 +19,7 @@ def validate_literal_argument(arg: Any, literal: Literal) -> None:
     ----------
     arg : Any
         The argument to validate.
-    literal : Literal[ValidTypes]
+    literal : Literal[str]
         A Literal type that defines the allowed values for `arg`.
 
     Raises
--- a/pyfixest/estimation/literals.py
+++ b/pyfixest/estimation/literals.py
@@ -8,6 +8,25 @@ solver_options = Literal["np.linalg.solve", "np.linalg.lstsq"]
 
 
 def validate_literal_argument(arg: Any, literal: Literal) -> None:
+    """
+    Validate if the given argument matches one of the allowed literal types.
+
+    This function checks whether the provided `arg` is among the valid types
+    returned by `get_args(literal)`. If not, it raises a ValueError with an
+    appropriate error message.
+
+    Parameters
+    ----------
+    arg : Any
+        The argument to validate.
+    literal : Literal[ValidTypes]
+        A Literal type that defines the allowed values for `arg`.
+
+    Raises
+    ------
+    ValueError
+        If `arg` is not one of the valid types defined by `literal`.
+    """
     valid_types = get_args(literal)
 
     if arg not in valid_types:
--- a/pyfixest/estimation/literals.py
+++ b/pyfixest/estimation/literals.py
@@ -1,4 +1,4 @@
-from typing import Any, Literal, _LiteralGenericAlias, get_args
+from typing import Literal, get_args, Any
 
 PredictionType = Literal["response", "link"]
 VcovTypeOptions = Literal["iid", "hetero", "HC1", "HC2", "HC3"]
@@ -7,7 +7,7 @@ FixedRmOptions = Literal["singleton", "none"]
 SolverOptions = Literal["np.linalg.solve", "np.linalg.lstsq"]
 
 
-def _validate_literal_argument(arg: Any, literal: _LiteralGenericAlias) -> None:
+def _validate_literal_argument(arg: Any, literal: Any) -> None:
     """
     Validate if the given argument matches one of the allowed literal types.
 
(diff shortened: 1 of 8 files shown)


marcandre259:
pre-commit.ci autofix

s3alfisc (approved):
Looks very good to me. Will leave this open until tomorrow to see if @leostimpfle has some thoughs, and then merge after we get his feedback. Thanks @marcandre259!

pyfixest/estimation/literals.py
```diff
@@ -0,0 +1,36 @@
...
+    Raises
+    ------
+    ValueError
+        If `arg` is not one of the valid types defined by `literal`.
+    """
+    valid_types = get_args(literal)
+
+    if arg not in valid_types:
+        raise ValueError(
+            f"Invalid prediction type. Expecting one of {valid_types}.\
```
leostimpfle, line 34: Suggest we change this to `f"Invalid argument. Expecting one of {valid_types}."`

pyfixest/estimation/literals.py
```diff
@@ -0,0 +1,36 @@
...
+        The argument to validate.
+    literal : Any
+        A Literal type that defines the allowed values for `arg`.
+
+    Raises
+    ------
+    ValueError
+        If `arg` is not one of the valid types defined by `literal`.
+    """
+    valid_types = get_args(literal)
```
leostimpfle, line 30: I'm wondering if we have to catch the case where `valid_types` is empty? This is because the function argument `literal` can be `Any`. If `literal` is a type without type arguments, e.g., `get_args(int)` returns an empty tuple.

pyfixest/estimation/literals.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any, Literal, get_args
+
+prediction_type = Literal["response", "link"]
+vcov_type_options = Literal["iid", "hetero", "HC1", "HC2", "HC3"]
+weights_type_options = Literal["aweights", "fweights"]
+fixef_rm_options = Literal["singleton", "none"]
+solver_options = Literal["np.linalg.solve", "np.linalg.lstsq"]
+
+
+def validate_literal_argument(arg: Any, literal: Any) -> None:
```
leostimpfle, line 10: Can we restrict the type of `literal`? I think we only want to call this function on literals?

pyfixest/estimation/literals.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any, Literal, get_args
+
+prediction_type = Literal["response", "link"]
```
leostimpfle, line 3: I've just checked naming conventions of type variables and PEP suggests that 

> Names of type variables introduced in [PEP 484](https://peps.python.org/pep-0484/) should normally use CapWords (https://peps.python.org/pep-0008/#type-variable-names)

Shall we adopt this and rename the literal types accordingly (e.g., `prediction_type` to `PredictionType`)?

leostimpfle (commented):
Thanks @marcandre259 ! Looks good to me overall, just have a few comments

pyfixest/estimation/literals.py
marcandre259: Hi @leostimpfle, thanks for reviewing. 

I can look into this. I tried using Literal as a type, but at least when linting, a specification of the options is required.

pyfixest/estimation/literals.py
s3alfisc: I guess generally it is good to follow style suggestion (though my general opinion is that it's fine to deviate now and then). I guess it won't hurt to follow them here?

pyfixest/estimation/literals.py
leostimpfle: My preference would be that the function raises an error if `valid_types` is empty so that one should only call the validation function on a non-empty literal type.

This would also catch the case where `literal` is not a `typing.Literal` object (previous comment) so that we don't have to worry about more specific type hints?

marcandre259: I now raise an error if the literal argument is not _LiteralGenericAlias. This seems to match the use case e.g. when I define EmptyType: Literal and feed it as the literal argument to _validate_literal_type

pyfixest/estimation/literals.py
marcandre259: Changed as proposed

pyfixest/estimation/literals.py
marcandre259: Found the _LiteralGenericAlias type that seems like a good match.

marcandre259:
pre-commit.ci autofix

s3alfisc:
pre-commit.ci autofix

s3alfisc:
It's a little bit annoying, but whenever the autofix corrects code, one has to run it again to pass the checks (because the previous did not pass as it triggered the fix) @marcandre259

pyfixest/estimation/literals.py
s3alfisc: Yep, this looks good to me!

s3alfisc:
Looks like `mypy` is complaining about the use of the `_LiteralGenericAlias`:

```python
pyfixest/estimation/literals.py:1: error: Module "typing" has no attribute "_LiteralGenericAlias"  [attr-defined]
```
Maybe we can simply check for a literal type via smth like 

```python
    if get_origin(literal) is not Literal:
        raise TypeError("Expected a Literal type as the second argument")
```
? 

I would not overthink this as it's in the end behavior we would (hopefully) catch in a review process?

s3alfisc:
Another thing I oversaw in my initial review - `feols` and `fepois` check input types in https://github.com/py-econometrics/pyfixest/blob/a3ed27aefef01e80ade7ebf117d5999239bbb62b/pyfixest/estimation/estimation.py#L617

Maybe we should update them to use the Literal validation function whenever relevant?

marcandre259:
> Looks like `mypy` is complaining about the use of the `_LiteralGenericAlias`:
> 
> ```python
> pyfixest/estimation/literals.py:1: error: Module "typing" has no attribute "_LiteralGenericAlias"  [attr-defined]
> ```
> 
> Maybe we can simply check for a literal type via smth like
> 
> ```python
>     if get_origin(literal) is not Literal:
>         raise TypeError("Expected a Literal type as the second argument")
> ```
> 
> ?
> 
> I would not overthink this as it's in the end behavior we would (hopefully) catch in a review process?

Hi @s3alfisc, 

Indeed, I'll modify it with a similar check. I did notice that Literal was not really a type without being specific Literal["..."], at least when trying out type(literal) is Literal. But maybe it works with get_origin then?

marcandre259:
pre-commit.ci autofix

s3alfisc:
Thanks @marcandre259 ! This looks good to me know =)

s3alfisc:
Just fyi @marcandre259 , the test failures in the main branch stem back from changes to the actions wfl I merged yesterday and are completely unrelated to this PR =)

Issue #557, opened 2024-07-21
Support for weights as an optional parameter for did2s?


AnonTendim:
Similar to what is in R fixest, will you consider adding support for weights: Optional variable to run a weighted first- and second-stage regressions in did2s? Thanks!

s3alfisc:
@rafimikail would this be something of interest for you to pick up?

rafimikail:
Hey @s3alfisc , Sure, i will take on this task!

s3alfisc:
Kyle's R journal paper is a great starting point to get familiar with the method: [link](https://journal.r-project.org/articles/RJ-2022-048/). 

Btw, thanks for the suggestion, @AnonTendim, very much appreciated!

rafimikail:
I will check that one out, thanks for giving the reference :D

marcandre259:
Apparently, there is a difference between pyfixest and fixest (R) when using the predict method with the new data argument. 

I'm still digging in this issue, but the reason behind it might be obvious. I'm getting different point estimates when including weights in pyfixest did2s and I figure this might be one of the reason. 

It seems like feols then gives the predictions of the unweighted regression. For example:

`
df_castle = pd.read_stata("https://github.com/scunning1975/mixtape/raw/master/castle.dta")
df_castle = df_castle.astype({"post": bool})

fit_untrt = pf.feols("l_homicide ~ 1 | state + year", data=df_untreated)
fit_untrt_wgt = pf.feols("l_homicide ~ 1 | state + year", data=df_untreated, weights="popwt")

yhat = fit_untrt.predict(newdata=df_castle)
yhat_wgt = fit_untrt_wgt.predict(newdata=df_castle)

# Getting first stage predictions
assert np.all(np.isclose(yhat[:5], yhat_wgt[:5]))

yhat = fit_untrt.predict()
yhat_wgt = fit_untrt_wgt.predict()

assert np.logical_not(np.all((np.isclose(yhat, yhat_wgt))))
`

s3alfisc:
Hi @marcandre259 , thanks for raising this! 

If I understand you correctly, we would expect that with WLS, `predict()` would **always** produce different values than OLS, unless all weights are identical. .

The behavior of the predict() method for `newdata = None` is controlled here: https://github.com/py-econometrics/pyfixest/blob/fd8e1436efd14cc98071eed51b91f98624cd1b6c/pyfixest/estimation/feols_.py#L1606. 

[...]

PR #692, opened 2024-11-02
Add the option to use analytical weights in 2-step diff-in-diff estimator

marcandre259:
Refers to #557 

Check that results are close to R's did2s effect's standard error and coefficient.


marcandre259:
pre-commit.ci autofix

pyfixest/did/did2s.py
```diff
@@ -82,13 +87,17 @@ def __init__(
         self._first_u = np.array([])
         self._second_u = np.array([])
 
+        # column name with weights. None by default
+        self._weights = weights
```
s3alfisc, line 91: For consistency with the `Feols` class, could we call this `self._weights_name`?

s3alfisc: Actually, not really a must have 🤔

s3alfisc:
Took a first look and this looks great! Will take a second look and then likely merge tomorrow morning. Is the test data set published under a license that allows us to re-publish it?

pyfixest/did/did2s.py
marcandre259: Hey, definitely good to add consistency. I'm going to add it.

marcandre259:
> Took a first look and this looks great! Thanks @marcandre259! Will take a second look and then likely merge tomorrow morning. Is the test data set published under a license that allows us to re-publish it?

I'm not aware of a license associated with the data. Here is some background info that may be useful in deciding whether it can be hosted for testing. 

It is data published by the American government (.gov) and cleaned by the authors below. It's referenced in Scott Cunningham's the mixtape and taken from Cheng Cheng and Mark Hoestra (2013). He says:
"The data they used is somewhat standard in crime studies. They used the FBI Uniform Crime Reports Summary Part I files from 2000 to 2010"

The data comes from:
DOES STRENGTHENING SELF-DEFENSE LAW DETER CRIME OR ESCALATE
VIOLENCE? EVIDENCE FROM CASTLE DOCTRINE
By Cheng Cheng and Mark Hoestra (2013)

Who point out the following:
Outcome data come from the FBI Uniform Crime Reports (UCR) and cover all 50
states from 2000 – 2010. 10 Specifically, we use homicide, burglary, robbery, and
aggravated assault data from the official UCR data published online by the FBI. 11 In
addition, for the other variables not available from the online UCR, we use data from the
FBI’s Master files (Return A and Supplemental Homicide Report).

s3alfisc:
@marcandre259 I dropped the new test data set (even though it is likely ok to include it, for example it is distributed via the R `did2s` package) and refactored the did2s tests - they now all run via the `df_het` data set. Will merge after the CI passes. Thanks for implementing this!

s3alfisc (approved):

PR #694, opened 2024-11-03
Add bugbear to pre-commit and some code improvements

juanitorduz:
Bugbear (https://docs.astral.sh/ruff/rules/#flake8-bugbear-b) is a great tool for detecting places where bugs can appear. This PR suggests adding this to the code base. On the way I found many places where we had mutable defaults in functions, which are very dangerous https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil

### DIFF
--- a/pyfixest/estimation/estimation.py
+++ b/pyfixest/estimation/estimation.py
@@ -13,7 +13,7 @@ from pyfixest.estimation.literals import (
     WeightsTypeOptions,
 )
 from pyfixest.utils.dev_utils import DataFrameType
-from pyfixest.utils.utils import ssc
+from pyfixest.utils.utils import ssc as ssc_func
 
 
 def feols(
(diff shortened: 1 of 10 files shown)


pyfixest/estimation/estimation.py
```diff
@@ -13,15 +13,15 @@
     WeightsTypeOptions,
 )
 from pyfixest.utils.dev_utils import DataFrameType
-from pyfixest.utils.utils import ssc
+from pyfixest.utils.utils import ssc as ssc_func
```
juanitorduz, line 16: This was a particularly tricky bug as `SCC` was both a global function and a function parameter.

s3alfisc:
Looks great, thank you Juan! I keep on being amazed by the Python tooling that is out there =) Everything looks good based on a first casual glance, though I will take a second look tomorrow before merging (I do not really trust myself after 11:00 😄 ).

juanitorduz:
Sure! Take a detailed review, as some of the changes were made by the pre-commit hook (no rush to merge).

I'll continue adding more guard rails to make the code even better than it is 💪

s3alfisc (approved):

s3alfisc:
Looks all good to me! Looking forward to more guardrails =) Thanks Juan @juanitorduz!

PR #697, opened 2024-11-05
Add RUFF rules

juanitorduz:
This is another PR (and probably the final in this sequence) to improve the code base through the linter. I suggest using the RUFF set of rules https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf, which sounds very reasonable.


s3alfisc (approved):

s3alfisc:
Looks good, thank you @juanitorduz!

Issue #533, opened 2024-07-01
narwhals = Pandas + polars + ...


juanitorduz:
Use https://github.com/narwhals-dev/narwhals to support pandas and polars!

This seems to be a very cool alternative to support various backends. See for example https://github.com/koaning/scikit-lego/pull/671

MarcoGorelli:
Hey, just wanted to stop by and say - thanks for your interest! Feel free to book some time on https://calendly.com/marcogorelli if you'd like to chat about how Narwhals could help PyFixest

s3alfisc:
Hi both (@MarcoGorelli and @juanitorduz) - I've now thought about it for 15 minutes and I think `narwhals` might be a great solution for `PyFixest`! Thanks for offering to chat @MarcoGorelli , I'll book an appointment =) 

Just some background on pyfixest and how it works with Data Frames: most of the data manipulation happens via the [formulaic](https://github.com/matthewwardrop/formulaic) library, which requires an input `pd.DataFrame`. I.e. a usual flow looks like this: 

```python
%load_ext autoreload
%autoreload 2

import polars as pl
import pandas as pd
import pyfixest as pf

from formulaic import model_matrix
import narwhals as nw

data = pl.DataFrame(pf.get_data())

def feols(data):

    if isinstance(data, pl.DataFrame):
        data = data.to_pandas()

    # model_matrix requires a pandas DataFrame and returns a pandas DataFrame
    Y, X = model_matrix("Y ~ X1", data = data, output = "pandas")

    # some more pandas manipulations
    Y.dropna(inplace = True)
    X.dropna(inplace = True)

    return Y.to_numpy(), X.to_numpy()

```
Via `narwhals`, it could look as 

```python
def feols_nw(data, use_polars = False):

    data = nw.from_native(data)

    # model_matrix requires a pandas DataFrame and returns a pandas DataFrame
    Y, X = model_matrix("Y ~ X1", data = data.to_pandas(), output = "pandas")

    if use_polars:
        # another copy? potentially costly? 
        Y = nw.from_native(Y)
        X = nw.from_native(X)

    # some more pandas manipulations
    Y.dropna(inplace = True)
    X.dropna(inplace = True)

    return Y.to_numpy(), X.to_numpy()
```

MarcoGorelli:
Hey! Thanks for your explanation - if `formulaic` requires specifically pandas input/output, and then that might be a good candidate for Narwhalification :) I'll take a look, thanks!

[...]

PR #700, opened 2024-11-09
`_narwhals_to_pandas`

vincentarelbundock:
This is proof of concept to partially solve \#533. It is a minimal change that swaps the existing `_polars_to_pandas()` by a `_narwhals_to_pandas()`.

Let me know if you want me to continue with this approach, and what you would expect in a merge-ready PR.

Here’s an example where `pyfixest` ingests a DuckDB table for both fitting and prediction:

``` python
import pyfixest as pf
import duckdb

data = pf.get_data()
data = duckdb.query("SELECT * FROM data")

type(data)

    <class 'duckdb.duckdb.DuckDBPyRelation'>
```
```python
fit = pf.feols("Y ~ X1 | f1 + f2", data=data)

fit.predict(newdata = data)[:10]
```
    [ 2.20341554         nan         nan  3.23931955 -1.44371374 -1.29643938
     -1.86119629 -1.34576434  1.38234328 -2.24958648]

```python
fit.summary()
```

    Estimation:  OLS
    Dep. var.: Y, Fixed effects: f1+f2
    Inference:  CRV1
    Observations:  997

    | Coefficient   |   Estimate |   Std. Error |   t value |   Pr(>|t|) |   2.5% |   97.5% |
    |:--------------|-----------:|-------------:|----------:|-----------:|-------:|--------:|
    | X1            |     -0.919 |        0.065 |   -14.057 |      0.000 | -1.053 |  -0.786 |
    ---
    RMSE: 1.441 R2: 0.609 R2 Within: 0.2 
    None


s3alfisc:
pre-commit.ci autofix

s3alfisc:
This looks really cool! I love that `narwhals` handles not only `polars` but also `duckdb`. Thank you =)

I actually think this is 95% ready to be merged. Maybe we could add a small API test [here](https://github.com/py-econometrics/pyfixest/blob/master/tests/test_api.py) that verifies that `pf.feols()` returns identical results, irrespective of the input data frame type. Additionally, would we have to change the type description for the input data frame type for the [pf.feols ](https://github.com/py-econometrics/pyfixest/blob/f1c8a59e86adead07df879d928c588a58206f538/pyfixest/estimation/estimation.py#L21) and pf.fepois functions? It currently uses a custom `DataFrameType` and asks the linter not to check - I would think that `narwhals` might support a type hint natively?

s3alfisc:
Took a quick look at the `narwhals` docs and the relevant type hint seems to be `narwhals.typing.IntoDataFrame`:

```python
from __future__ import annotations

import narwhals as nw
from narwhals.typing import IntoDataFrame

def func(df_native: IntoDataFrame) -> tuple[int, int]:
    df = nw.from_native(df_native, eager_only=True)
    return df.shape
```

s3alfisc:
All the test errors are struggles or `rpy2`, which happens sometimes. I'll take a look at these @vincentarelbundock =)

vincentarelbundock:
Excellent! I changed the type hint and added a small test.

Don't know what your policy is for changelog and version number while in dev. happy to make a change if you indicate or let you do it yourself in my branch before/after merge.

s3alfisc:
> Don't know what your policy is for changelog and version number while in dev. happy to make a change if you indicate or let you do it yourself in my branch before/after merge.

Usually we don't bump versions when merging from dev to main, which I suppose is bad practice? Changelog are automatically handled by a [release bot](https://github.com/py-econometrics/pyfixest/blob/master/.github/release-config.yml), which automatically updates the [github release notes](https://github.com/py-econometrics/pyfixest/releases). 

Do you have a recommendation on how to potentially improve these processes?

s3alfisc:
pre-commit.ci autofix

pyfixest/utils/dev_utils.py
```diff
@@ -1,24 +1,22 @@
...
-            data = data.to_pandas()
-        except ImportError:
-            raise ImportError(
-                """Polars is not installed. Please install Polars to use it as
-                an alternative."""
-            )
+    except ImportError:
+        raise ImportError(
+            """Unable to use Narwhals to convert the DataFrame to Pandas."""
+        )
```
MarcoGorelli, line 19: I don't think it's possible for this to return an `ImportError`, as both narwhals and pandas would already be required dependency according to this PR, I think you can just `return nw.from_native(data).to_pandas()`

I'd also suggest
- using `import narwhals.stable.v1 as nw` instead of `import narwhals as nw`
- writing this as
  ```
  nw.from_native(data, eager_or_interchange_only=True).to_pandas()
  ```

This is because defining a lazy-only layer of support (which DuckDB, PySpark, and others) is high on our priorities for 2025, and we may need to change the API slightly. But if you use `stable.v1`, then we promise that what you've written here will keep working

MarcoGorelli (commented):
thanks for trying out narwhals here, just wanted to leave a comment, hope that's ok

pyfixest/utils/dev_utils.py
vincentarelbundock: Thanks a lot for taking the time to comment @MarcoGorelli !

I made the suggested changes.

s3alfisc:
@all-contributors please add @vincentarelbundock for code 
@all-contributors please add @MarcoGorelli for review

s3alfisc (approved):

s3alfisc:
@all-contributors please add @MarcoGorelli for review

Issue #704, opened 2024-11-10
Test against R fail: `crossprod` is not a BUILTIN function


s3alfisc:
Apparently, `crossprod` has been changed a while back ([with R 4.4.0](https://stat.ethz.ch/R-manual/R-patched/doc/html/NEWS.html)), with downstream impact on pyfixest. As a result, multiple unit tests currently fail. We could fix this by moving away from r2u and installing all R dependencies from source, which would lead to pretty bad increases in CI runtime, or by installing a lower R version via r2u (which I don't think is possible). 

r2u currently installs R 4.4.2 in the CI wfls. 

Context from r-devel: [link](https://stat.ethz.ch/pipermail/r-package-devel/2023q4/009958.html)

FYI @leostimpfle @vincentarelbundock, this is currently leading to test errors in your PRs.

PR #706, opened 2024-11-11
remove duplicate error classes

### LINKED ISSUE #701: Use consistent error classes
Some of `pyfixest`'s custom error classes in `pyfixest.errors` duplicate Python's built-in exceptions, for example `NotImplementedError`:

https://github.com/py-econometrics/pyfixest/blob/f1c8a59e86adead07df879d928c588a58206f538/pyfixest/errors/__init__.py#L41

For consistency, we should always use Python's built-in exceptions and remove duplicate errors in `pyfixest.errors`, for example here:

https://github.com/py-econometrics/pyfixest/blob/f1c8a59e86adead07df879d928c588a58206f538/pyfixest/estimation/fepois_.py#L10

A list of Python's built-in errors is available [here](https://docs.python.org/3/library/exceptions.html#concrete-exceptions).

leostimpfle:
Closes #701 

Very minor changes removing the custom `NotImplementError`.

@s3alfisc I've noticed that `pyfixest.errors.__init__` includes an undefined `MultiEstNotSupportedError` in `__all__` which I have also removed. Please let me know if instead it should be defined as an additional error.

### DIFF
--- a/pyfixest/did/event_study.py
+++ b/pyfixest/did/event_study.py
@@ -4,7 +4,6 @@ import pandas as pd
 
 from pyfixest.did.did2s import DID2S
 from pyfixest.did.twfe import TWFE
-from pyfixest.errors import NotImplementedError
 
 
 def event_study(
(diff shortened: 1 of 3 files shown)


s3alfisc:
Test errors only due to #704 . Looks good to me and can be merged. Thanks!

s3alfisc (approved):

PR #707, opened 2024-11-11
Model summary if intercept-only

### LINKED ISSUE #702: Help fail `tidy` gracefully if model is intercept only
Fitting an intercept only model, e.g., `"Y ~ 1 | f1 + f2"` will make the `tidy` method fail ungracefully because of an error in `get_inference`: `ValueError: diag requires an array of at least two dimensions`

It would be nice to catch the intercept-only case at an earlier stage in `tidy` (or maybe `get_inference`?) to avoid throwing hard-to-understand errors in `get_inference`.

Minimal working example:

    import pyfixest as pf
    data = pf.get_data(N=1000, seed=0, beta_type="2", error_type="2", model="Feols" )
    feols = pf.feols("Y ~ 1 | f1 + f2", data )
    feols.coef()

leostimpfle:
Small updates to `get_inference` and `tidy` to deal with the case of a model with intercept only. My approach is to let `get_inference` raise a custom `EmptyVcovError` if `self._vcov` is empty but only raise a warning in `tidy` and return empty results. This also means that derived methods (e.g., `summary()` only raise a warning).

Closes #702


s3alfisc:
I might finally merge this tonight @leostimpfle . Sorry for the delay!

PR #714, opened 2024-11-13
polars -> narwhals

s3alfisc:
Full migration from all code parts that rely on `polars` to `narwhals`. Drops `polars` dependency.


s3alfisc:
Hi @marcogorelli ,I am trying to fully deprecate the `polars` dependency in favor of `narwhals`. I am using `pl.count()` for which `narwhals` does not seem to support an equivalent. Is that correct? If yes, is there any other way to rewrite 

```python
    agg_expressions.append(pl.count(depvars[0]).alias("count"))
``` 
to `narwhals` supported syntax?

s3alfisc:
Actually, this isn't even true, of course there is a `count` in `narwhals`: [link](https://narwhals-dev.github.io/narwhals/api-reference/expr/#narwhals.Expr.count).

s3alfisc:
This should work now =)

s3alfisc:
Unit tests are now very slow - looks like I lose a lot of performance, likely by doing something that is not recommended 🤔

s3alfisc:
I think I completely misunderstood how `narwhals` works - after all, I now realized that narwhals does not copy code from pandas to polars and runs polars under the hood (not sure where I got this from, maybe because all the R translators of dplyr do this?), but simply translates code! So the performance drop came from running all computations via pandas instead of polars 🤦

PR #721, opened 2024-11-18
fix docstring typos

RoyalTS:
Changes so minor I almost want to apologize. Stumbled on whilst going through code, seemed worth fixing, even just for aesthetic/OCD reasons :)

### DIFF
--- a/pyfixest/estimation/FormulaParser.py
+++ b/pyfixest/estimation/FormulaParser.py
@@ -230,7 +230,7 @@ class FixestFormulaParser:
 
 class FixestFormula:
     """
-    A class with information confainted in model formulas.
+    A class with information contained in model formulas.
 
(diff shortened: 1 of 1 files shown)


RoyalTS:
Feels way out of proportion to name me as a contributor for changing three characters. If this is a subtle attempt to try to guilt me into making an actual contribution, it may be working :)

s3alfisc:
Haha I feel seen 😄 though beyond this being a not so subtle attempt to incentivize / lure one-off contributors to contribute more regularly, I really like the idea to also acknowledge contributions that go beyond writing code & implementing full features. And then it's somewhat difficult to make a cutoff, so my decision rule at the moment is that "everything that is merged or leads to a code change" qualifies =)

Issue #717, opened 2024-11-15
Randomization inference, "ri" sampling_method in rwolf, gives too tight a sample of null t-statistics


marcandre259:
Possible issue I noticed while working on #698. 

The behavior was initially noticed when comparing "wild-bootstrap" to the "ri" sample_method p-values when the parameter of interest has no association with the outcome. 

With the tight null t-distribution, the resulting p-value is too small. 

To reproduce:

```python
import pyfixest as pf
import numpy as np

import matplotlib.pyplot as plt

# Get data and randomize
data = pf.get_data()

np.random.default_rng(232)
data["X1"] = np.random.choice(data["X1"], size=data.shape[0], replace=False)

fit = pf.feols("Y ~ X1", data=data)

fit.summary()
```

Estimation:  OLS
Dep. var.: Y, Fixed effects: 0
Inference:  iid
Observations:  998

| Coefficient   |   Estimate |   Std. Error |   t value |   Pr(>|t|) |   2.5% |   97.5% |
|:--------------|-----------:|-------------:|----------:|-----------:|-------:|--------:|
| Intercept     |     -0.160 |        0.119 |    -1.344 |      0.179 | -0.394 |   0.074 |
| X1            |      0.033 |        0.090 |     0.367 |      0.714 | -0.144 |   0.211 |
---
RMSE: 2.304 R2: 0.0 

```python
seed = 111
df_wild, df_t_wild = fit.wildboottest(param="X1", reps=9999, return_bootstrapped_t_stats=True, seed=seed)

rng = np.random.default_rng(232)
fit.ritest(resampvar="X1", reps=9999, type="randomization-t", store_ritest_statistics=True, rng=rng)

fig, ax = plt.subplots(nrows=1, ncols=2, figsize=(12, 4))
ax[0].hist(fit._ritest_statistics, label="RI t stats", alpha=0.4);
ax[0].axvline(x=fit._ritest_sample_stat, linestyle="--", label="Observed RI t stats", color="black")
ax[0].legend()
ax[1].hist(df_t_wild, label="Wild t stats", alpha=0.4, color="orange");
ax[1].axvline(df_wild["t value"], label="Observed Wild t stat", color="black", linestyle="--");
ax[1].legend()
```

![comparing_null_t_empirical_distributions](https://github.com/user-attachments/assets/0133afa7-d7ef-4a5b-ad92-f08380791d6b)

s3alfisc:
[...]

PR #725, opened 2024-11-23
Adding westfall-young p-value adjustment for multiple tests

marcandre259:
From closed pr #698

Takes in most of @s3alfisc review comments. 
Tests are still a weak point. 
Possible issue with Randomization Inference -> #717


marcandre259:
pre-commit.ci autofix

s3alfisc:
Thanks @marcandre259 , I'll review this evening!

s3alfisc:
Hi, this looks very good, thank you @marcandre259! I left a tiny comment that I fixed myself =) I might also add some unit tests tomorrow and think again about the issue of RI vs the bootstrap you raised in the other PR. What is your current thinking on this at the moment? Do you think it is a bug, or are we being fooled by statistics?

marcandre259:
> Hi, this looks very good, thank you @marcandre259! I left a tiny comment that I fixed myself =) I might also add some unit tests tomorrow and think again about the issue of RI vs the bootstrap you raised in the other PR. What is your current thinking on this at the moment? Do you think it is a bug, or are we being fooled by statistics? 

Thanks for looking into it @s3alfisc :) I can also add tests if needed. 

Regarding the (maybe) issue, for me what is weird is getting this super tight null distribution. I figure maybe I'm missing something. In any case, there are probably tons of randomization sampling programs out there to do comparisons, so if it's still hanging, I'll look into it this weekend.

s3alfisc:
Thanks for tackling the tests! 

But the null distribution seems to fit the analytical SEs? The tests against [ritest ](https://github.com/grantmcdermott/ritest)are [here](https://github.com/py-econometrics/pyfixest/blob/master/tests/test_ritest.py), though I only check p-values as far as I recall, not null distribution widths.

s3alfisc:
Actually there are tests for the SEs of the ritest t-stats and the lower bound of its ci implemented as well: https://github.com/py-econometrics/pyfixest/blob/ce4efd370710bf2a75b66815f3f614cbe4946fbf/tests/test_ritest.py#L104 Though if you have access to a Stata license, would of course be great to check against Stata ritest as well!

s3alfisc:
Now that I've found the ritest bug and fixed it in #730, I will take another look at this & implement some tests tomorrow evening. 👍

s3alfisc:
pre-commit.ci autofix

s3alfisc (approved):

s3alfisc:
And merged. Thanks so much @marcandre259!

Issue #720, opened 2024-11-18
Coefplot: Plotting the same regression equation using different dataframes


escherpf:
I am trying to use <coefplot> for a list of models that fit the same equation (with the same variable names) but using different data frames. For example, I would like to plot one (or two) variables for each of those models to get something like this:

![example_coefplot](https://github.com/user-attachments/assets/7e923b1a-4186-4b1e-aec6-925f36ba26a4)

However, as far as I can tell, this can only be achieved if the model specification is not exactly identical across models. If it is, <coef_plot> generates what seems like a strange repeating pattern:

![example_coefplot2](https://github.com/user-attachments/assets/cd62a1a5-bea9-4be1-bfcb-eefc48df1043)

Hence, it seems like <coefplot> gets confused if the models in the list use the same formula.
Is it currently possible to plot the same variable from the same model specification estimated on different different frames _as _though_ they were different models? 

Thank you.

s3alfisc:
Yes - the problem arises because `coefplot` loops over the provided models and uses the model name. I see two options: 

Option 1: You could work with the `split` argument: 

```python
import pyfixest as pf 

df = pf.get_data()
fit = pf.feols("Y ~ X1", split = "f1", data = df[df.f1.isin([1,2])])
fit.coefplot(coord_flip=False, keep = "X1")
```
![image](https://github.com/user-attachments/assets/279bde91-30b8-4afe-9635-28636fe59a6f)

Alternatively, you could overwrite the `_model_name` attribute by hand: 

```python
fit1 = pf.feols("Y ~ X1", data = df[df.f1.isin([1])])
fit2 = pf.feols("Y ~ X1", data = df[df.f1.isin([2])])

fit1._model_name
# 'Y~X1'
fit2._model_name
# 'Y~X1'

fit1._model_name += ", sample f1 = 1"
fit2._model_name += ", sample f1 = 2"
pf.coefplot([fit1, fit2], coord_flip=False, keep = "X1")
```
which also produces 

![image](https://github.com/user-attachments/assets/da53173b-6882-4334-a9fe-9fbc8d7ce339)

Maybe it would be convenient to add a `model_name` argument to `pf.coefplot()` that would allow users to pass custom model names to avoid duplicates?

escherpf:
Thank you for this tip! Accessing and modifying the model name did the trick!

s3alfisc:
#Hi @escherpf I would reopen this as 
[...]

PR #748, opened 2024-12-15
Rename Regression Plot in Case of Duplicate Names

s3alfisc:
Context: see #720.

Now we get three different non-overlapping models in case the `_model_name` attribute is identical across models: 

```python
import pyfixest as pf

data = pf.get_data()
fit1 = pf.feols("Y ~ X1", data = data)
fit2 = pf.feols("Y ~ X1", data = data)
fit3 = pf.feols("Y ~ X1", data = data)
fit4 = pf.feols("Y ~ X1 + X2", data = data)

pf.coefplot([fit1, fit2, fit3, fit4])
```
image

Users are notified with a warning than an attribute is overwritten.


s3alfisc:
@all-contributors please add @escherpf for bug

PR #756, opened 2024-12-19
update did error handling of time variables

s3alfisc:


escherpf (commented):
For me the problem wasn't so much the way errors were handled as much as did supporting variables to be able to work more easily with year-month variables, but it seems that is a bigger lift.  A workaround for the user would be to recast the year-month variable into a relative time (i.e., an integer), which is essentially what is done with did2s.

pyfixest/did/did.py
```diff
@@ -17,14 +18,12 @@ class DID(ABC):
     idname : str
         The name of the identifier variable.
     tname : str
-        Variable name for calendar period. Must be an integer in the format
-        YYYYMMDDHHMMSS, i.e. it must be possible to compare two dates via '>'.
-        Datetime variables are currently not accepted.
+        Variable name for calendar period. Must be a numeric (int or float) and
+        of the same data type as the gname variable.
     gname : str
```
escherpf, line 23: this should be int or float instead of str, no?

pyfixest/did/did.py
```diff
@@ -17,14 +18,12 @@ class DID(ABC):
     idname : str
         The name of the identifier variable.
     tname : str
```
escherpf, line 20: Should be int or float rather than str, no?

Issue #731, opened 2024-12-04
`predict` bug with mismatched data types


ivanhigueram:
Hello there, 

I am trying to create predictions with a dataset outside my model data. I found that if there's any type mismatch in the `newdata` compared with the `data` we used to estimate the model, the `predict()` method will return an array of `nan`: 

Here's a reproducible example:

```python
import pyfixest as pf

data = pf.get_data()

# Run model
model = pf.feols("Y ~ X1 | f1", data=data)

# Allow me to fill the nans just to make my points :)
data["f1"] = data["f1"].fillna(0).astype(int)
model.predict(newdata=data.iloc[0:100])

>>> array([nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, ...])
```

This error is coming from the definition of `df_fe` in the `predict()` and the `_apply_fixef_numpy` as the dictionary keys will be saved with a `10.0` rather than `10`. 
https://github.com/py-econometrics/pyfixest/blob/10a8fb8417e5c7581bf6ae5824f1aa6221f1d261/pyfixest/estimation/feols_.py#L1832

Not sure if this constitutes a bug on itself, or if this is a skill issue, but it would be nice to get a warning maybe? This is no problem for unit FEs if the ID is a `str`, but in dates, sometimes we get `2010.0` rather than `2010` when we do time operations and numpy preserves the data type. 

I am running the `'0.26.2'` version in Python `3.10`.

s3alfisc:
Thanks for reporting this Ivan @ivanhigueram! I have to think about this one for a bit. My intuition would be that a keys of 10.0 and 10 should be treated equally? For sure I think adding a warning would be a good start!

s3alfisc:
If you had to choose, what would be your preferred behavior?

ivanhigueram:
I'd say a warning would be enough. Is definitely easier to do a `astype()` in the `newdata` than trying to fix this in your code base.

s3alfisc:
Yes I was also afraid that handling things in pyfixest would be a lot of work 😅 I'll add a warning then (or would you be up to open a PR? but of course no pressure 😄 ). 

I think it could be as simple as adding 

```python
        if self._has_fixef: 
            fixef = self._fixef.split("+")

            mismatched_fixef_types = [x for x in fixef if newdata[x].dtypes != self._data[x].dtypes]
            if mismatched_fixef_types:
[...]

PR #758, opened 2024-12-22
GLM Support

s3alfisc:
This PR 

- Adds an abstract `Feglm` class from which different type of GLM families can inherit.
- Adds a `pf.feglm(, family = "logit")` front end to call these functions. 
- The implementation closely follows this paper by Stamann: https://arxiv.org/pdf/1707.01815. Equations and algo are implemented "verbatim" - should be easy to match code to paper
- Families to support for now: (unconditional) logit, probit, an alternative Poisson implementation. 

Other PR
- Later: anything else of interest. Clogit (different estimator). Negbin? 
- Incidental Param Problem Bias Corrections

TODO: 

- [x] Implement abstract GLM base class (inheriting from Fepois for now, later this should be reversed) with IWLS algo and step halfing
- [ ] Implement Logit class based on Stamman (not yet converging, bug somewhere)
- [x] store IWLS parameters so that they can be used in post estimation classes (i.e. for computing vcov's etc). See `Fepois.get_fit()` method for details. 
- [ ] tests against `fixest::feglm()`

Goal: 

```python
import pyfixest as pf
data = pf.get_data(model ="Fepois")
data["Y"] = np.where(data.Y > 0, 1, 0)
fit = pf.feglm("Y ~ X1 | f1", data = data, family = "logit")
```

Should all be doable before the year ends (hopefully!)

Fyi @leostimpfle @apoorvalal @gbekes.

PR #761, opened 2024-12-26
Add warning when predict has a type mismatch

ivanhigueram:
As discussed in (#731), when predict's `newdata` has a different `dtype` in the FEs, the function will yield missing values only, this because predict relies on a dictionary of FEs where keys are transformed to strings and will fail to get the correct value (i.e. 10 vs "10.0").

This commit adds a warning in the `predict` method and a simple unit test that checks for this case for the user.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
         args: ["--verbose"]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.8.3
+    rev: v0.8.4
     hooks:
       - id: ruff
         args: ["--fix", "--output-format=full"]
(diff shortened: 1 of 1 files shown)


s3alfisc (approved):
This looks great, thank you @ivanhigueram! 

@all-contributors please add @ivanhigueram for code =)

s3alfisc:
pre-commit.ci autofix

s3alfisc:
We get new mypy errors - I just fixed those in my local branch of #758 =D Will push a patch!

s3alfisc:
pre-commit.ci autofix

s3alfisc:
https://github.com/all-contributors please add @ivanhigueram for code

s3alfisc:
Oh wow changes are clearly not working, I'll tackle this tomorrow

Issue #380, opened 2024-04-25
Implement Core Algorithms in JAX


s3alfisc:
The other day @janosg suggested that implementing some of the core algorithms in [JAX ](https://jax.readthedocs.io/en/latest/)might provide exceptional performance improvements. 

`pyfixest's` alternating projections algorithm is implemented [here](https://github.com/s3alfisc/pyfixest/blob/867643d9e69072d90aa0190b5603d66cfb56560f/pyfixest/estimation/demean_.py#L136). 

A vectorized `numpy` version of the algorithm is implemented in the [pyhdfe library](https://github.com/jeffgortmaker/pyhdfe/blob/25d79b090f6b68d1c4879bd25d99da27b7a7fb54/pyhdfe/algorithms.py#L335). 

A very simple implementation of the MAP algorithm is sketched out in R in the [lfe documentation ](https://cran.r-project.org/web/packages/lfe/vignettes/lfehow.pdf)(at the bottom of page 5): 

```R
demean <- function(v, fl) {
  Pv <- v
  oldv <- v - 1
  while (sqrt(sum((Pv - oldv)**2)) >= 1e-7) {
    oldv <- Pv
    for (f in fl) Pv <- Pv - ave(Pv, f)
  }
  Pv
}
```
Also tagging @jeffgortmaker, as I recall you suggesting using JAX to solve the MAP fixed point problem at some point =)

janosg:
No guarantees but could be worth a try. The important thing would be to use [lax control flow operators](https://jax.readthedocs.io/en/latest/jax.lax.html#control-flow-operators) instead of Python loops and vectorize as much as possible.

juanitorduz:
Which could be a good candidate algorithm to start with? I might also wanna try it :)

jeffgortmaker:
Probably just simple iterative de-meaning. Here's something close that I threw together for my own research:

```python
def jax_residualize(matrix, ids, iterations=100):
    """Iteratively residualize a matrix within groups. With more than one set of groups, simply de-mean up to a fixed
    number of iterations to allow for reverse mode differentiation (neither custom roots nor returning the same thing
    after a tolerance was reached seemed to be working).
    """
    import jax
    import jax.numpy as jnp

    def demean(values, groups):
        """Demean within groups."""
        group_sums = jax.ops.segment_sum(values, groups, groups.size)
        group_counts = jax.ops.segment_sum(jnp.ones_like(values), groups, groups.size)
[...]

PR #767, opened 2024-12-30
JAX demean implementation via Claude 3.5

juanitorduz:
Related to https://github.com/py-econometrics/pyfixest/issues/380

JAX implementation of the demean function + tests

### DIFF
--- a/pyfixest/estimation/demean_.py
+++ b/pyfixest/estimation/demean_.py
@@ -1,8 +1,12 @@
+from functools import partial
 from typing import Any, Optional
 
+import jax
+import jax.numpy as jnp
 import numba as nb
 import numpy as np
 import pandas as pd
+from jax import config, lax
 
 
 def demean_model(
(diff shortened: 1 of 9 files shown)


juanitorduz:
Any hints on the failing tests? They seem unrelated to the PR, right?

pyfixest/estimation/demean_.py
```diff
@@ -298,3 +302,84 @@ def demean(
...
+                wts = wts.at[idx].add(weights[i])
+                return sums, wts
+
+            # Use fori_loop to match Numba's iteration order
+            group_sums, group_weights = lax.fori_loop(
+                0, x_curr.shape[0], _accumulate, (group_sums, group_weights)
+            )
+
+            # Compute weighted means exactly as in Numba
+            group_means = group_sums / (group_weights[:, None] + 1e-15)
```
s3alfisc, line 344: Here I would drop the `1e-15` , I see no reason to add it? In my experience LLMs are very prone to bake in some guardrails to avoid numerical problems that I never see in computational stats, maybe this is a ML thing?

juanitorduz: changed!

juanitorduz:
ok @s3alfisc your numba code is very hard to beat hehe

image

I am unsure jax can beat it to be honest... shall we close this one and keep it for future reference? At least we know is not entirely trivial 

image

juanitorduz:
maybe this would pay of on GPU?

juanitorduz:
ok in https://github.com/py-econometrics/pyfixest/pull/767/commits/084a55da04ae764c9a18238dc8b055d2c066686d I changed the jit strategy and seems faster

image

s3alfisc:
Nice! In my experience, making code faster is a super dangerous rabbit hole, once I start I have a hard time to abandon it ... 😄 

Generally I think it would be great to get to test this on a GPU - if it is indeed faster than pure `numba`, I think we should consider to add JAX as an optional dependency and allow users to run the demeaning through it? 

Another option is to simply ask for help in the [JAX forum](https://github.com/jax-ml/jax/discussions)? The numba code only got very fast because @styfenschaer did his magic after I [asked in the numba forum](https://numba.discourse.group/t/optimization-of-numba-implementation-of-an-alternating-projection-algorithm/2193).

juanitorduz:
agree! I think https://github.com/py-econometrics/pyfixest/pull/767/commits/7d85386b02f967a79181971078dd6411dd0d4e97 is a good place to start where I try to jit as much as possible which was the missing piece on the first approach.

image

It would be really nice to have numba as default and use JAX  as an optional dependency IF it's faster on GPU. I will try to test in Google Colab

juanitorduz:
For `N = 1_000_000` seems are most comparable 

image

juanitorduz:
@s3alfisc @jeffgortmaker The results on GPU look very promising! (`N = 100_000`)

image

[Here](https://colab.research.google.com/drive/1vvL73B6Lwicm7agzBvwRugPkbHSYEApg#scrollTo=TLDCdD-XG3D8) is the Google colab notebook (its essentially copy-paste the code from this branch and run the `tests_demean`)

s3alfisc:
One other part of the code base that can be relatively easily ported to JAX is the OLS solver here: https://github.com/py-econometrics/pyfixest/blob/170dd98dc27882d84237cbe7527d7ad22982a0f7/pyfixest/estimation/feols_.py#L441 

Basically we could add a solver option "JAX" and then simply update to 

```python
    
    from jax.numpy.linalg import lstsq as jax_lstsq
    
    def solve_ols(tZX: np.ndarray, tZY: np.ndarray, solver: str):
        if solver == "np.linalg.lstsq":
            return np.linalg.lstsq(tZX, tZY, rcond=None)[0].flatten()
        elif solver == "np.linalg.solve":
            return np.linalg.solve(tZX, tZY).flatten()
        elif solver == "scipy.sparse.linalg.lsqr":
            return lsqr(tZX, tZY)[0].flatten()
       elif solver == "jax": 
            return jax_lstsq(tZX, tZY, rcond=None)[0].flatten()
        else:
            raise ValueError(f"Solver {solver} not supported.")
```
and put the function in a standalone script to avoid namespace clashes?

s3alfisc:
I guess the CI fails because JAX is no longer available in the dev environment (which inherits from default)?

juanitorduz:
yes! I think we can port many components to jax. Shal we do it on a different PR? This one makes JAX an optional dependency. We can then create another PR where we re-write core algorithms and decide whether we want to make JAX a core dependency. WDYT?

s3alfisc:
Yes, let's donut like that! =)

juanitorduz:
@s3alfisc anything else this PR is missing? It could wait if you wanna do more thorough benchmarks 😄

s3alfisc:
Generally it looks good - I'll do a proper review of the algo implementation later today. Should we make it callable via `feols / fepois` already in this PR, or should we do this in a separate PR, maybe after the GLMs are merged? I'd also like to update the [benchmarks in here](https://github.com/py-econometrics/pyfixest/blob/master/benchmarks/run_benchmarks.ipynb) then with JAX CPU & JAX GPU demeaning.

juanitorduz:
Sure! No rush :)

s3alfisc (approved):

s3alfisc:
Merged! Thank you so much for this Juan, I am super excited!

Issue #769, opened 2024-12-31
Enable formulaic to parse custom data transforms from context


adamvig96:
Add the option to provide the extra context via the feols / fepois APIs for formulaic's [context capture method](https://github.com/matthewwardrop/formulaic/blob/ddb3c11dbb8cb661234142a753d7bea1f7a644b5/formulaic/utils/context.py#L7).

s3alfisc:
Can be closed by merging #770 =)

PR #770, opened 2024-12-31
Add context capture to formula materialisation

adamvig96:
Context: https://github.com/py-econometrics/pyfixest/issues/769

### DIFF
--- a/pyfixest/estimation/FixestMulti_.py
+++ b/pyfixest/estimation/FixestMulti_.py
@@ -1,6 +1,6 @@
 import functools
 from importlib import import_module
-from typing import Optional, Union
+from typing import Optional, Union, Mapping, Any
 
 import pandas as pd
 
@@ -10,6 +10,7 @@ from pyfixest.estimation.feols_compressed_ import FeolsCompressed
 from pyfixest.estimation.fepois_ import Fepois
 from pyfixest.estimation.FormulaParser import FixestFormulaParser
 from pyfixest.utils.dev_utils import DataFrameType, _narwhals_to_pandas
+from pyfixest.utils.utils import capture_context
 
 
 class FixestMulti:
@@ -29,6 +30,7 @@ class FixestMulti:
         split: Optional[str],
         fsplit: Optional[str],
         separation_check: Optional[list[str]] = None,
+        context: Union[int, Mapping[str, Any]] = 0,
     ) -> None:
         """
         Initialize a class for multiple fixed effect estimations.
@@ -61,6 +63,11 @@ class FixestMulti:
         separation_check: list[str], optional
             Only used in "fepois". Methods to identify and drop separated observations.
             Either "fe" or "ir". Executes both by default.
+        context : int or Mapping[str, Any]
+            A dictionary containing additional context variables to be used by
+            formulaic during the creation of the model matrix. This can include
+            custom factorization functions, transformations, or any other
+            variables that need to be available in the formula environment.
 
         Returns
         -------
@@ -75,6 +82,7 @@ class FixestMulti:
         self._reps = reps if use_compression else None
         self._seed = seed if use_compression else None
         self._separation_check = separation_check
+        self._context = capture_context(context)
 
         self._run_split = split is not None or fsplit is not None
         self._run_full = not (split and not fsplit)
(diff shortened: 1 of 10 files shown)


s3alfisc:
pre-commit.ci autofix

adamvig96:
@s3alfisc I added a unit test. 

Would you like to add some integration tests too? if so, I'd be happy to have some guidance on where to put those. Thanks!

s3alfisc:
Hi @adamvig96 , looking very good so far, thank you! Could you add integration into [test_api.py](https://github.com/py-econometrics/pyfixest/blob/af99811f325340a29e52c79624012062f9ad6523/tests/test_api.py)? I guess that testing equality of coefficients and SEs of creating splines via the new capture arg + formulaic vs "by hand" should do the job?

adamvig96:
pre-commit.ci autofix

adamvig96:
Hi @s3alfisc! I added an integration test, wdyt is this enough? Thanks! :)

s3alfisc:
Yes, I think so, thank you! Will review and merge later today. One thing that wasn't clear to me - why is the argument set to 0 be default and set to 2 in the code if no mapping is provided?

adamvig96:
I set it to 0 because it was the default in formulaic, but formulaic always adds one when calling their capture_context function (see [this](https://github.com/matthewwardrop/formulaic/blob/ddb3c11dbb8cb661234142a753d7bea1f7a644b5/formulaic/sugar.py#L57) for example). I wanted to avoid that repetition of checking whether we got an int (and the add +1) or a mapping, that's why I created the wrapper function in utils.

We should probably set it to 1 by default and adding +1 in our wrapper function, maybe that's more readable?

s3alfisc:
Hi @adamvig96 , sorry for the slight delay - I took a look at the formulaic code base and it looks like the `get_model_matrix` method expects a mapping or None: only what is passed as context to the method will be available for creating the design matrix. So I think that we could drop the `context()` function and explicitly pass the context argument to the method? 

https://github.com/matthewwardrop/formulaic/blob/ddb3c11dbb8cb661234142a753d7bea1f7a644b5/formulaic/model_spec.py#L684

s3alfisc:
pre-commit.ci autofix

adamvig96:
Hi @s3alfisc, thanks for taking a look! If I understand your proposal correctly, that would mean that we only call the `capture_context` function at the feols/fepois api level – we need to call the function somewhere to get the default context, i think – and from there on, the context is just passed through until it reaches formulaic?

s3alfisc:
I actually just realized that I misunderstood how everything works, and what I suggested above doesn't make sense / work. Sorry!

s3alfisc:
By setting the defaults to `context = 0`, we basically always pass the entire context of the `feols` call to `get_model_matrix`, right? Do you think there is any downside to that, i.e. could there be side effects? For example, could it happen that users accidentally pass something that overwrites internal behavior? (Currently I don't think so). Maybe it would be safer to set the default to `None` and to explicitly ask users to either provide the context as a mapping, or to use the "global" context? Something along these lines: 
```python
def feols(context=None):
    if context is None:
        context = {}
    elif isinstance(context, int):
        context = capture_context(context)
```

adamvig96:
Yes, you're right that if the user overwrites some default functions, then formulaic will use those, instead [the defaults](https://github.com/matthewwardrop/formulaic/blob/ddb3c11dbb8cb661234142a753d7bea1f7a644b5/formulaic/transforms/__init__.py#L32) Btw this is how statsmodels work too.

I'll add None as default and the condition tomorrow!

adamvig96:
pre-commit.ci autofix

adamvig96:
Hi @s3alfisc, I made the context parameter optional in feols and fepois.

s3alfisc:
pre-commit.ci autofix

s3alfisc:
pre-commit.ci autofix

s3alfisc:
@all-contributors please add @adamvig96 for code

s3alfisc:
pre-commit.ci autofix

s3alfisc:
pre-commit.ci autofix

s3alfisc:
pre-commit.ci autofix

s3alfisc:
Sorry for all the noisy @adamvig96 , I promise I will not code in the github browser going forward ... 🤕

s3alfisc:
And merged! Thank you Adam @adamvig96 =)

adamvig96:
@s3alfisc, thanks for the support! 💯

PR #772, opened 2025-01-01
Add JAX OLS Solver

juanitorduz:
See https://github.com/py-econometrics/pyfixest/pull/767#issuecomment-2566959514

Allow to use [`jax.numpy.linalg.lstsq`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.lstsq.html)

### DIFF
--- a/pyfixest/estimation/solvers.py
+++ b/pyfixest/estimation/solvers.py
@@ -1,0 +1,35 @@
+import numpy as np
+from scipy.sparse.linalg import lsqr
+
+
+def solve_ols(tZX: np.ndarray, tZY: np.ndarray, solver: str) -> np.ndarray:
+    """
+    Solve the ordinary least squares problem using the specified solver.
+
+    Parameters
+    ----------
+    tZX (array-like): Z'X.
+    tZY (array-like): Z'Y.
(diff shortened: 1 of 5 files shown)


pyfixest/estimation/solvers.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    if solver == "np.linalg.lstsq":
+        return np.linalg.lstsq(tZX, tZY, rcond=None)[0].flatten()
+    elif solver == "np.linalg.solve":
+        return np.linalg.solve(tZX, tZY).flatten()
+    elif solver == "scipy.sparse.linalg.lsqr":
+        return lsqr(tZX, tZY)[0].flatten()
+    elif solver == "jax":
+        import jax.numpy as jnp
+
```
juanitorduz, line 32: We can still keep this an optional dependency for now. If we keep adding more jax code we can think on adding it as a core dependency. I do no think this should be any issue

s3alfisc: Yes, agreed 👍

s3alfisc (approved):

PR #775, opened 2025-01-02
Add more tests

juanitorduz:
I think we can increase the code coverage. This is a first initial step. For this purpose LLMs can provide a great starting point :)

### DIFF
--- a/tests/test_dgps.py
+++ b/tests/test_dgps.py
@@ -1,0 +1,177 @@
+import numpy as np
+import pandas as pd
+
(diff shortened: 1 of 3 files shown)


s3alfisc (approved):
That's great, thanks! Nice reminder for me to be a little bit more thorough with codecov =) Generally the test coverage is actually a little bit higher than suggested - codecov only picks up tests that the CI runs, but I had to disable others (for example, the ritest tests just take way to long to finish).

Issue #45, opened 2023-04-07
Make sure to drop singleton fixed effects from the clustering matrix


s3alfisc:
Link to original issue: [link ](https://github.com/s3alfisc/pyfixest/issues/41)

Issue #24, opened 2023-03-14
Improve performance


s3alfisc:
- [ ]  See [here](https://stackoverflow.com/questions/55367024/fastest-way-of-solving-linear-least-squares) for faster solving of the least squares problem. 
- [ ]  faster demeaning algo via numba / c++
- [ ] clean up code, drop redundancies, etc
- [x]  for a given column in the design matrix, only project out fe's when not already done (for multiple estimation)

PR #784, opened 2025-01-09
Jax benchmarks

s3alfisc:
Add JAX benchmarking notebook.


benchmarks/gpu_pyfixest_errors.ipynb
```diff
@@ -0,0 +1,1478 @@
+{
```
juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #24.        dum_all[:, 0] = np.random.choice(nb_dum[0], n, replace=True)
</pre><p>You could use a rng (<a href="https://numpy.org/doc/stable/reference/random/generator.html" target="_blank">https://numpy.org/doc/stable/reference/random/generator.html</a>) to make the examples reproducible </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/py-econometrics/pyfixest/pull/784/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #45.                "ln_y": np.log(y + 1),
</pre><p>there is <code>np.log1p</code> (<a href="https://numpy.org/doc/stable/reference/generated/numpy.log1p.html" target="_blank">https://numpy.org/doc/stable/reference/generated/numpy.log1p.html</a>) hehe </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/py-econometrics/pyfixest/pull/784/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

juanitorduz, line 1: <p>demeaning times ara nan?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/py-econometrics/pyfixest/pull/784/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #11.    g = sns.FacetGrid(
</pre><p>I think you could use <a href="https://seaborn.pydata.org/generated/seaborn.catplot.html" target="_blank">https://seaborn.pydata.org/generated/seaborn.catplot.html</a> directly </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/py-econometrics/pyfixest/pull/784/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

juanitorduz:
I left some minor comments.

Despite this is a dev notebook. I think it would be nice to have a title, a description of what's happening, and the motivation (and conclusions?) . Think a bout a new dev reading this in 6 months ;) (it could be you! hehe)

benchmarks/gpu_pyfixest_errors.ipynb
s3alfisc: Only when the demeaning is explicitly skipped.

s3alfisc: Done.

s3alfisc:
Thanks Juan! I think I should have adressed all your comments in https://github.com/py-econometrics/pyfixest/pull/784/commits/279167a51b4f010ae813c0f9a96eeb2afbbe73f7 =)

s3alfisc:
I was also think that we might just add this nb to the docs, it might be interesting for users to see these benchmarks?

benchmarks/gpu_benchmarks.ipynb
```diff
@@ -0,0 +1,1494 @@
+{
```
juanitorduz, line 1: <p>Did the results chage? I thought JAX on GPU was faster</p><p> </p><p>Also, the y ticks are hard to read ;)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/py-econometrics/pyfixest/pull/784/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

s3alfisc:
Hi, no, this is still all run on my Laptop, so no GPUs involved. The idea was that @apoorvalal could use this nb to run benchmarks on the GPU.

juanitorduz:
got it! thanks :) (then just the comment about the y -ticks ;) )

benchmarks/gpu_benchmarks.ipynb
s3alfisc: <p>On the y-ticks - I was hoping they'd get easier to read one the benchmarks ran for longer (though am not sure if seaborn actually supports this type of auto-scaling, but I would suppose so?)</p>

apoorvalal:
Working on it; immediately ran into some cuda issues (jax is great on goog's TPUs but much more fiddly on nvidia gpus; torch is much better on those) so had to get some sysadmins to fix. Should be able to finish running today

s3alfisc:
Awesome, I'm super curious 😀

s3alfisc:
And thanks!

apoorvalal:
Ran and added results (csv and notebook). As expected, pretty big performance gains from jax in large problems. Your results in `gpu_benchmarks.ipynb` don't actually use a GPU (`jnp.ones(10).devices()` shows `{CpuDevice(id=0)}`, i.e. the array is being created on the CPU), so that notebook should be renamed to reflect that it is Jax performance on CPU, and probably suggests that the JAX backend is only worthwhile on a powerful GPU. 

Also, I noticed that the jax demeaning algo is potentially inefficient; it doesn't have early stopping (`has_converged` never triggers a `break` in `apply_factor` or `body_fun`) so always goes over a fixed number of iterations. I can take a crack at improving it, but that might be best left for a separate PR.

apoorvalal:
also cc @iamlemec ; any thoughts on how the alternating projections step is implemented here? This isn't quite as general as your MLE implementation in fastreg (since the primary problem has a closed form solution).

s3alfisc:
Wow this looks super good! Thanks @apoorvalal! Basically we see a 2-3x speed increase for big problems? Do you happen to know how many CPUs numba had available? 

> Also, I noticed that the jax demeaning algo is potentially inefficient; it doesn't have early stopping (has_converged never triggers a break in apply_factor or body_fun) so always goes over a fixed number of iterations. 

😅 about this. So JAX really always runs the full maximum 100_000k iterations? Then we shouldn't be surprised that we don't see any speed improvements on the CPU (but really worse performance). Seeing the glass half full I'd say even more scope for performance improvements :D

s3alfisc:
Wow, I wasn't aware that `fastreg` runs GLMs on JAX (cc @juanitorduz) - if you have any suggestions on how to optimize [demean_jax](https://github.com/py-econometrics/pyfixest/blob/master/pyfixest/estimation/demean_jax_.py), that would be super appreciated @iamlemec! I think we're all still starting out with JAX and at least I still have a lot to learn about performance optimization (and even more basic things, as I apparently missed a break statement in the while clause...).

s3alfisc:
@apoorvalal I think that the while loop breaks here https://github.com/py-econometrics/pyfixest/blob/f6ce66e19cc7dbe15fa7688e6befaa972d465155/pyfixest/estimation/demean_jax_.py#L62 if `_conf_fun` is False, or not? https://github.com/py-econometrics/pyfixest/blob/f6ce66e19cc7dbe15fa7688e6befaa972d465155/pyfixest/estimation/demean_jax_.py#L55

apoorvalal:
Oh right that makes sense; I was tripped up by the haskell-y separate conditional function.

s3alfisc:
> Oh also the instance I ran on had 12 CPU cores, so I'm assuming numba has access to all of them?
Yes, it should use all available CPUs by default 👍 

> Oh right that makes sense; I was tripped up by the haskell-y separate conditional function.
I also think that reading JAX code takes some time getting used too...

apoorvalal (approved):
looks ready to merge after some cleanup; maybe delete the old GPU benchmarks notebook and lightly edit the GPU benchmark one and integrate it into the docs page as well. 

My conclusion from the benchmarking exercise is that the JAX backend is only really useful for very large datasets with access to fairly specialised GPUs, so we should keep the numba backend as the default and leave JAX as a poweruser feature.

iamlemec:
Thanks for tagging me in @apoorvalal and @s3alfisc! Not surprisingly, this already looks really good to me. I tried my darnest to eek out some more performance gains (see this gist: [demean_jax_.py](https://gist.github.com/iamlemec/7d6eca7111052a6c5f3ac31bf3635d05)) and I maybe got 5-10%, but it's probably configuration dependent.

My main takeaways were:
- You can factor out the nested functions and just have one top-level jit, but it doesn't seem to make a huge difference performance wise
- For the call to `scan`, you can get away with slicing through `flist` directly, rather than using a range to address it
- You actually don't need the `x_prev` part of the state in the current implementation (I'm assuming that was vestigial)
- You can get away with not passing `tol` to `static_argnums` since its a float

One thing I'm confused about: why don't the large gains on the demean step pass through one-to-one to the full feols estimation? I would think it's a subset of the full op. Also, are there plans to bring JAX to other parts of the code?

apoorvalal:
jnp.linear is likely not the fastest jax-based alternative for solving least-squares systems fwiw. Lineax has some nice automatic methods to pick the right decomposition method and seems to [show pretty good improvements over scikit's regression class](https://github.com/apoorvalal/jaxonometrics/blob/master/jax_linear.ipynb) (which uses numpy/scipy solvers)

then again, given how mature pyfixest's internals are, I would be in favour of a separate experimental jax-based econometrics package that we develop and iterate on faster and potentially integrate into pyfixest further down the line.

s3alfisc:
> then again, given how mature pyfixest's internals are, I would be in favour of a separate experimental jax-based econometrics package that we develop and iterate on faster and potentially integrate into pyfixest further down the line.

I'd be up for that (might really make more sense than trying to squeeze things into pf)! If we were to go ahead with this, I'd try to involve @janosg and @timmens - did the two of you ever got around to sketch out the design doc for a JAX-first stats package you had in mind?

s3alfisc:
> But we could have a call to exchange some ideas.

Sounds like a good first step @janosg - I'd be happy to try to arrange this and invite everyone interested (please responding with a 👍if you are!)

> I think the ideal solution would be to make pyfixest more extensible 

I think this is a great suggestion - for the beginning, we could continue to port core pieces of the code to `JAX` and collect it in a sub module. We could try to write the code in a very functional style that would make it easy to be reused in different code bases, so that later on, we could decide to spin it out (or just the most useful pieces) into a standalone package?

s3alfisc:
Slightly off topic - I've [opened a PR ](https://github.com/py-econometrics/pyfixest/pull/790)to move all JAX code into a `jax` module + have implemented a POC [JAXOLS class](https://github.com/py-econometrics/pyfixest/blob/80bd6a4ce2cb67bfd89811645eda55e5ac698586/pyfixest/estimation/jax/OLSJAX.py) that does everything in JAX (demeaning, fitting, inference).

iamlemec:
Thanks for the info @s3alfisc! About the timing stats, I think I was confused about the numbers coming out of the GPU benchmark notebook. Still getting 5-10% boost and things make sense now.

I'll wait till #790 settles and then submit a PR. I'll poke around a bit too and see if there are any good targets for JAX conversion, HC* stuff does seem promising. I agree that repeated transfers back and forth might be an issue, which kind of pushes one towards a full-pipeline approach.

Would happily contribute to a JAX-first stats package too! Count me in if there's a call/discussion.

s3alfisc:
Thanks all for your responses! I'll send out an email this weekend then =) @iamlemec I'll ping you once I get to merge #790 .

PR #798, opened 2025-01-25
Add feature to automatically rename categorical variables in etable

dsliwka:
This adds a feature to etable that automatically renames categorical variables in regression tables. The user can specify a formatting template (default is "{variable}={value}"). It should work either for pandas categoricals (i.e "y ~ x + c" when c is categorical) and with categorical variables generated in the regression formula with the C() or i() operator.  

To do this, the existing rename_categoricals function is slightly extended to allow using formatting templates.

### DIFF
--- a/docs/table-layout.qmd
+++ b/docs/table-layout.qmd
@@ -15,7 +15,7 @@ To begin, we load some libraries and fit a set of regression models.
 ```{python}
 import numpy as np
 import pandas as pd
-import pylatex as pl  # for the latex table; note: not a dependency of pyfixest - needs manual installation
+#import pylatex as pl  # for the latex table; note: not a dependency of pyfixest - needs manual installation
 from great_tables import loc, style
 from IPython.display import FileLink, display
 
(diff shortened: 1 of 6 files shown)


s3alfisc:
pre-commit.ci autofix

s3alfisc:
> "i(x, ref=v)" then having blanks in category (i.e. level) names leads to an error.

Ah ok, I see! I think we should treat this as a bug of the `i()` parser and I might open a separate issue for it? 

Btw, I have also changed the default to not apply the relabeling by default - the reason is that I think there should be consistency in internal variable naming and users should explicitly ask for renaming. Else we might run into users who see variables names in etable() but struggle to understand how they might relate to what is included in `.coef()`, `.tidy()` or `.summary()`?

s3alfisc:
Merged this one for now, thanks Dirk! On the i-ref and white space issue - let's see if I will manage to fix it, if not happy to discuss bringing back the `set_first_cat` function =)

dsliwka:
Thanks!!

Issue #348, opened 2024-03-12
Feols: Implement R2 for WLS


s3alfisc:
## Context

Implement support for R2 measures for weighted least squares. 

So far, this is not implemented, as there are quite a few options out there. For details on what R's `stats` package does, see [here](https://stats.stackexchange.com/questions/439590/how-does-r-compute-r-squared-for-weighted-least-squares). Likely `fixest` follows `stats`, which we would have to confirm. 

## To do

- Research how `fixest` implements R2 for WLS. 
- Implement R2 measures for WLS in [Feols.get_performance()](https://github.com/py-econometrics/pyfixest/blob/734a21c34a56a88ba12b0fa26cd3ed0e8c24bd51/pyfixest/estimation/feols_.py#L1389). 
- Update summary and etable to return R2 measures if regression weights are employed. 
- Implement unit tests against fixest in `tests/test_vs_fixest.py::test_single_fit`.

mpalenciaolivar:
I want to work on this!

escherpf:
While this issue is being addressed (this would be an excellent improvement, btw), is there a way to hide the (blank) R2 reported in the etable output?

Thank you!

s3alfisc:
This is now implemented @escherpf, thanks to @AronNemeth's efforts =)

PR #815, opened 2025-02-24
Feols: Implement R2 for WLS

AronNemeth:
This pull request references issue #348 

I implemented the R2 measures for WLS regressions. The implementation follows what fixest does. I compared fixest results and this implementation on the test data.

Howeverer, I didn't add WLS R2 to the tests yet, so that still is outstanding. I am happy to do that as well as next step

### DIFF
--- a/pyfixest/estimation/feols_.py
+++ b/pyfixest/estimation/feols_.py
@@ -1872,11 +1872,9 @@ class Feols:
         _has_fixef = self._has_fixef
         _weights = self._weights
 
-        if _has_fixef:
-            _k_fe = np.sum(self._k_fe - 1) + 1
-            _adj_factor = (_N - _k_fe) / (_N - _k - _k_fe)
-        else:
-            _adj_factor = (_N - _has_intercept) / (_N - _k)
+        _adj_factor = (_N - _has_intercept) / (_N - _k)
+        _k_fe = np.sum(self._k_fe - 1) + 1
+        _adj_factor_fe = (_N - _k_fe) / (_N - _k - _k_fe)
 
         ssu = np.sum(_u_hat**2)
         ssy = np.sum(_weights * (_Y - np.average(_Y, weights=_weights)) ** 2)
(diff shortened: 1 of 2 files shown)


s3alfisc:
For tests, you would more or less un-comment the if clause in this line https://github.com/AronNemeth/pyfixest/blob/b18fd8de6ef36565ea9e6950967b687894c4c8c4/tests/test_vs_fixest.py#L328 - then the tests for R2 and R2adjusted would run for OLS as well =)

s3alfisc:
I have also added some unit tests =)

s3alfisc:
@all-contributors please add @AronNemeth for code

s3alfisc:
Ok some good news and some bad news: the tests for the regular `R2` value pass, but for the adjusted R2 fail. You can see this if you scroll to the very end of this test logs [here](https://github.com/py-econometrics/pyfixest/actions/runs/13507682483/job/37740987161?pr=815) @AronNemeth . But I am quite sure that if we adjust the adj R2 computation as mentioned above ⬆️ , the tests will also pass =)

s3alfisc:
Btw, once we merge this, it will be PyFixest's 1000s commit =)

s3alfisc:
Looks like something is still off in the calculation of adjusted R2: 
image

s3alfisc:
Indeed the error seems to lie in the adj r2 computations, as all other tests pass.

AronNemeth:
Pushed the corrected adjusted R2 calculation. This issue took more iterations than I expected, since I couldn't execute test locally. I managed to fix that problem, and for my latest commit all the local tests passed. So this implementation should be okay

s3alfisc:
Excellent, I will merge this now. Thanks so much @AronNemeth!

s3alfisc:
https://github.com/all-contributors please add @AronNemeth for code

s3alfisc:
@all-contributors please add @AronNemeth for code

PR #818, opened 2025-02-27
Modified code to make lets-plot an optional dependency.

elchorro:
This PR makes `lets-plot` into optional dependency (#253). It adds installation option which includes `lets-plot` and uses this package as a default whenever it is available.


elchorro:
@s3alfisc I am having trouble with pixi and I'm unfamiliar with this. I run the tests locally but the GH workflow fails on setup. Any suggestions?

s3alfisc:
pre-commit.ci autofix

s3alfisc:
@all-contributors please add @elchorro for code

s3alfisc:
I think this should be ready now. Thanks so much for picking this up @elchorro!

elchorro:
Thanks @s3alfisc for helping land this -- I had difficulties installing `lets-plot` in some environments, so I wanted try address this.

s3alfisc:
It's really much appreciated! We've actually experienced the same thing, but I never really prioritized it, so I am glad it's finally done =) I'll try to prepare a release of v0.29 by Sunday so that you won't have to install from github!

elchorro:
Hi @s3alfisc, when are you planning a release? Any other blocking issues I could help with to speed up?

PR #830, opened 2025-03-08
Create a minimal environment with core dependencies

s3alfisc:
Hi @elchorro , here is a first attempt at creating a minimal environment. 

It allows users to access all estimation function and methods (`feols`, `fepois, `feglm` etc) and a range of post-processing methods (`tidy`, `summary`, plotting via `plt`). 

It drops `joblib`, `great_tables`, `tqdm` and `tabulate` from the main env and moves it into a secondary environment `full` which will be installable by typing `pip install pyfixest[full]`. 

In the new "core" env
- `tqdm` is replaced by a dummy function
- `joblib` only executed when installed 
- calling `GT` produces an `ImportError`
- calling `tabulate` produces an `ImportError`

What still would have to be done: 

- [ ] We should add logic to the CI workflow to run tests in the "core environment" vs the base environment, to make sure things work well without the "optional" dependencies. 
- [ ] We have to update the readme. 
- [ ] We should decide if the "minimal" environment should be the "core" environment. I assume that it would be more user friendly to install etable etc by default and require users with other specs to specifically install the "core" env? I would be curious about @dsliwka and @juanitorduz ' thoughts on this? 
- [ ] Should we make `seaborn` an optional dependency? If no, we could replace the seaborn dependency by a matplotlib dependecy (the only piece of code that relies on seaborn is the [plotting method for randomization inference ](https://github.com/py-econometrics/pyfixest/blob/0850dad74225cbd7bdf3eb0f82333ccabab76e01/pyfixest/estimation/ritest.py#L402) which should be an easy fix. 
- [ ] We could even make `narwhals` optional, in which case we'd have to thrown an error wherever it converts non `pd.DataFrame`'s for us (happens in maybe 3 scripts). 

If you have some time to polish this PR, it would be much appreciated @elchorro ! =)

### DIFF
--- a/pyfixest/estimation/decomposition.py
+++ b/pyfixest/estimation/decomposition.py
@@ -3,10 +3,24 @@ from typing import Any, Optional
 
 import numpy as np
 import pandas as pd
-from joblib import Parallel, delayed
 from scipy.sparse import csc_matrix
 from scipy.sparse.linalg import lsqr
-from tqdm import tqdm
+
+try:
+    from tqdm import tqdm
+except ImportError:
+
+    def tqdm(iterable, *args, **kwargs):
+        "Define a dummy tqdm function."
+        return iterable
+
+
+try:
+    from joblib import Parallel, delayed
+
+    joblib_available = True
+except ImportError:
+    joblib_available = False
 
 
 @dataclass
@@ -154,9 +168,12 @@ class GelbachDecomposition:
         self.alpha = alpha
         self.B = B
 
-        _bootstrapped = Parallel(n_jobs=self.nthreads)(
-            delayed(self._bootstrap)(rng=rng) for _ in tqdm(range(B))
-        )
+        if joblib_available:
+            _bootstrapped = Parallel(n_jobs=self.nthreads)(
+                delayed(self._bootstrap)(rng=rng) for _ in tqdm(range(B))
+            )
+        else:
+            _bootstrapped = [self._bootstrap(rng=rng) for _ in tqdm(range(B))]
 
         self._bootstrapped = {
             key: np.concatenate([d[key] for d in _bootstrapped])
--- a/pyfixest/estimation/feols_compressed_.py
+++ b/pyfixest/estimation/feols_compressed_.py
@@ -6,7 +6,15 @@ from typing import Any, Literal, Optional, Union
 import narwhals as nw
 import numpy as np
 import pandas as pd
-from tqdm import tqdm
+
+try:
+    from tqdm import tqdm
+except ImportError:
+
+    def tqdm(iterable, *args, **kwargs):
+        "Define a dummy tqdm function."
+        return iterable
+
 
 from pyfixest.estimation.feols_ import Feols, PredictionErrorOptions, PredictionType
 from pyfixest.estimation.FormulaParser import FixestFormula
--- a/README.md
+++ b/README.md
@@ -57,7 +57,14 @@ You can install the release version from `PyPI` by running
 python -m pip install pyfixest
 ```
 
-or the development version from github by running
+The above will install a bare-bone version of pyfixest that will allow you to fit models but lacks post-processing functionality and
+estimation extensions. We recommend to install all dependencies by running
+
+```py
+python -m pip install pyfixest[full]
+```
+
+You can install the development version from github by running
 
 ```py
 python -m pip install git+https://github.com/py-econometrics/pyfixest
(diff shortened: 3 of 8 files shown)


s3alfisc:
I adjusted the pyproject toml so that the `default` env has a richer dependency than `minimal`. Users who run `pip install pyfixest` install gt, tabulate, etc by default; users who run `pip install pyfixest[minimal]` will get an env with only the core dependency set. 

```py
(pyfixest) PS C:\Users\alexa\Documents\pyfixest> pixi list -x                                               
Package       Version  Build               Size        Kind   Source                                        
formulaic     1.1.1                        371.3 KiB   pypi   formulaic-1.1.1-py3-none-any.whl              
great_tables  0.16.1                       6.4 MiB     pypi   great_tables-0.16.1-py3-none-any.whl
joblib        1.4.2                        886.2 KiB   pypi   joblib-1.4.2-py3-none-any.whl
narwhals      1.29.1                       1.5 MiB     pypi   narwhals-1.29.1-py3-none-any.whl
numba         0.61.0                       11.7 MiB    pypi   numba-0.61.0-cp312-cp312-win_amd64.whl        
numpy         2.1.3                        40.2 MiB    pypi   numpy-2.1.3-cp312-cp312-win_amd64.whl
pandas        2.2.3                        37.7 MiB    pypi   pandas-2.2.3-cp312-cp312-win_amd64.whl        
pyfixest      0.28.0                                   pypi    (editable)
python        3.12.8   h3f84c4b_1_cpython  15.1 MiB    conda  python
scipy         1.15.2                       115.1 MiB   pypi   scipy-1.15.2-cp312-cp312-win_amd64.whl        
seaborn       0.13.2                       1018.2 KiB  pypi   seaborn-0.13.2-py3-none-any.whl
tabulate      0.9.0                        128.4 KiB   pypi   tabulate-0.9.0-py3-none-any.whl
tqdm          4.67.1                       220.5 KiB   pypi   tqdm-4.67.1-py3-none-any.whl
(pyfixest) PS C:\Users\alexa\Documents\pyfixest> pixi shell --environment minimal
(pyfixest:minimal) PS C:\Users\alexa\Documents\pyfixest> pixi list -x
Environment: minimal
Package    Version  Build               Size        Kind   Source
formulaic  1.1.1                        371.3 KiB   pypi   formulaic-1.1.1-py3-none-any.whl
narwhals   1.29.1                       1.5 MiB     pypi   narwhals-1.29.1-py3-none-any.whl
numba      0.61.0                       11.7 MiB    pypi   numba-0.61.0-cp312-cp312-win_amd64.whl
numpy      2.1.3                        40.2 MiB    pypi   numpy-2.1.3-cp312-cp312-win_amd64.whl
pandas     2.2.3                        37.7 MiB    pypi   pandas-2.2.3-cp312-cp312-win_amd64.whl
pyfixest   0.28.0                                   pypi    (editable)
python     3.12.8   h3f84c4b_1_cpython  15.1 MiB    conda  python
scipy      1.15.2                       115.1 MiB   pypi   scipy-1.15.2-cp312-cp312-win_amd64.whl
seaborn    0.13.2                       1018.2 KiB  pypi   seaborn-0.13.2-py3-none-any.whl
```
I guess we should still check with the `pixi` team if this is "supported" behavior or a non-supported hack?

pyfixest/estimation/feols_compressed_.py
```diff
@@ -6,7 +6,15 @@
...
 import pandas as pd
-from tqdm import tqdm
+
+try:
+    from tqdm import tqdm
+except ImportError:
+
+    def tqdm(iterable, *args, **kwargs):
+        "Define a dummy tqdm function."
+        return iterable
```
juanitorduz, line 16: This is being repeated a lot ... why don't we let tqdm be a core dep? It seems very lightweight

pyfixest/estimation/decomposition.py
```diff
@@ -3,10 +3,24 @@
...
+        "Define a dummy tqdm function."
+        return iterable
+
+
+try:
+    from joblib import Parallel, delayed
+
+    joblib_available = True
+except ImportError:
+    joblib_available = False
```
juanitorduz, line 23: shall we at least raise a warning so that these import errors do not pass silently?

pyfixest/estimation/feols_compressed_.py
s3alfisc: Same with `joblib`, actually - it has [no other dependencies except for python ](https://pypi.org/project/joblib/). More or less the motivation of this PR is to make the package so lightweight that it can easily be installed in closed environments (with no access to pypi / conda). My prior would be that both joblib and tqdm should actually be installed on every system that runs Python, though I'm not 100% sure. 

I am more or less gravitating towards having a default / core env that includes joblib, tqdm, GT etc that users install with `pip install pyfixest` and then a minimal but still functional env for those who can really only work with numpy, scipy, etc. Would be your suggestion to keep tqdm in the "minimal" dependency set @juanitorduz?

pyfixest/estimation/decomposition.py
s3alfisc: done 👍

PR #839, opened 2025-03-12
Fix formula parsing bug

DTchebotarev:
This PR addresses a bug where the predict method re-parses the string formula which can result in a different shape of the model matrix. This is fixed by storing the modelspec object and then using that to create the model matrix, instead of re-parsing it from scratch. I've added a test case as well to test exactly this.

@s3alfisc please let me know what you think. I'm happy to do a rev or two if you'd like changes. Once you're happy, would love to get your help cutting a patch version (i.e. bumping to v0.28.1) so this fix can be picked up by pypi.

### DIFF
--- a/pyfixest/estimation/prediction.py
+++ b/pyfixest/estimation/prediction.py
@@ -55,16 +55,10 @@ def get_design_matrix_and_yhat(
                     "predict() with argument newdata is not supported with i() syntax."
                 )
 
-            # Use the stored model_spec instead of creating a new Formula
             if hasattr(model, "_model_spec") and model._model_spec is not None:
-                # Get the right-hand side model spec from the nested structure
-                # ModelSpecs contains fml_second_stage which contains rhs
                 rhs_spec = model._model_spec.fml_second_stage.rhs
-
-                # Use the right model spec directly with get_model_matrix
                 X = rhs_spec.get_model_matrix(newdata)
             else:
-                # Fallback to old behavior if model_spec is not available
                 xfml = model._fml.split("|")[0].split("~")[1]
                 X = Formula(xfml).get_model_matrix(newdata)
 
--- a/pyfixest/estimation/prediction.py
+++ b/pyfixest/estimation/prediction.py
@@ -56,7 +56,7 @@ def get_design_matrix_and_yhat(
                 )
 
             # Use the stored model_spec instead of creating a new Formula
-            if hasattr(model, '_model_spec') and model._model_spec is not None:
+            if hasattr(model, "_model_spec") and model._model_spec is not None:
                 # Get the right-hand side model spec from the nested structure
                 # ModelSpecs contains fml_second_stage which contains rhs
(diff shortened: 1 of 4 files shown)


DTchebotarev:
pre-commit.ci autofix

tests/test_predict_resid_fixef.py
```diff
@@ -298,6 +298,17 @@ def test_categorical_covariate_predict():
...
 
+def test_specific_categorical_prediction():
+    """Test prediction with a specific categorical case."""
+    test_df = pd.DataFrame(
+        {"y": [2, 3, 4, 5], "x": [1, 1, 2, 3], "f": ["a", "b", "a", "a"]}
+    )
+    test_model = pf.feols("y ~ x + C(f)", data=test_df)
+    prediction = test_model.predict(newdata=pd.DataFrame({"x": [1], "f": ["b"]}))
+    expected_prediction = 3
+    np.testing.assert_almost_equal(prediction[0], expected_prediction, decimal=3)
```
DTchebotarev, line 309: On master this assertion fails; the prediction returned is 2.16: https://github.com/py-econometrics/pyfixest/issues/840

s3alfisc:
This looks excellent. Thank you @DTchebotarev! I will merge after the CI run =)

s3alfisc:
@all-contributors please add @DTchebotarev for code

pyfixest/estimation/prediction.py
```diff
@@ -50,12 +50,18 @@ def get_design_matrix_and_yhat(
...
                     "predict() with argument newdata is not supported with i() syntax."
                 )
-            X = Formula(xfml).get_model_matrix(newdata)
+
+            if hasattr(model, "_model_spec") and model._model_spec is not None:
+                rhs_spec = model._model_spec.fml_second_stage.rhs
+                X = rhs_spec.get_model_matrix(newdata)
+            else:
+                xfml = model._fml.split("|")[0].split("~")[1]
+                X = Formula(xfml).get_model_matrix(newdata)
```
s3alfisc, line 63: Generally much better to work with the `model_spec` and the already deparsed info in in than to do the parsing again - we have a few of these cases in the code base and should clean them up step by step. But that's something for another PR =)

s3alfisc (approved):

pyfixest/estimation/prediction.py
```diff
@@ -50,12 +50,18 @@ def get_design_matrix_and_yhat(
         newdata = _narwhals_to_pandas(newdata).reset_index(drop=False)
 
         if not model._X_is_empty:
-            xfml = model._fml.split("|")[0].split("~")[1]
             if model._icovars is not None:
                 raise NotImplementedError(
                     "predict() with argument newdata is not supported with i() syntax."
                 )
```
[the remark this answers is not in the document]
s3alfisc: No, this doesn't work, as FixestFormula only contains `fml_second_stage` as a string and is not derived from a formulaic object, hence the `rhs` method is not supported. Also maybe something to change going forward!

Issue #821, opened 2025-03-01
Report Within R2 in `pf.etable()`


s3alfisc:
For compatibility with `r-fixest`, we would like to report **adjusted R2** in `pf.etable()`.

In R, we have 

![Image](https://github.com/user-attachments/assets/6ca843ee-d1a6-4424-a389-450d53410b59)

In Python, we have 

```python
import pyfixest as pf 
data = pf.get_data()

fit = pf.feols("Y ~ X1 | f1", data = data, weights = "weights")
pf.etable(fit)
```

![Image](https://github.com/user-attachments/assets/4e715931-84e7-471e-b791-90f14b032820)

To do so, we need to populate the [r2_within_list](https://github.com/py-econometrics/pyfixest/blob/f25dbdd1cf34f0ef6c539418a1902554583ecd4d/pyfixest/report/summarize.py#L236) [here](https://github.com/py-econometrics/pyfixest/blob/f25dbdd1cf34f0ef6c539418a1902554583ecd4d/pyfixest/report/summarize.py#L261).

We then add it to the [model_stats_df](https://github.com/py-econometrics/pyfixest/blob/f25dbdd1cf34f0ef6c539418a1902554583ecd4d/pyfixest/report/summarize.py#L293C20-L293C26) here .

And I hope this would be it! (Though not sure, would have to be tested). 

@AronNemeth would you be up for this one? Should be relatively small and self-contained.

AronNemeth:
Sure thing! I'll get to work on this

s3alfisc:
Awesome! =)

leostimpfle:
This now works in `maketables`:

https://py-econometrics.github.io/maketables/docs/ETable.html#specify-displayed-model-statistics

PR #842, opened 2025-03-13
Add R2 within to etable #821

AronNemeth:
pf.etable() now reports the within R2 for FE models

### DIFF
--- a/pyfixest/report/summarize.py
+++ b/pyfixest/report/summarize.py
@@ -233,20 +233,23 @@ def etable(
     n_coefs = []
     se_type_list = []
     r2_list = []
-    r2_within_list: list[float] = []  # noqa: F841
+    r2_within_list: list[float] = []
 
     # Define code for R2, interaction & line break depending on output type
     if type in ["gt", "html"]:
         interactionSymbol = " &#215; "
         R2code = "R<sup>2</sup>"
+        R2_within_code = "R<sup>2</sup> Within"
         lbcode = "<br>"
     elif type == "tex":
         interactionSymbol = " $\\times$ "
         R2code = "$R^2$"
+        R2_within_code = "$R^2$ Within"
         lbcode = r"\\"
     else:
         interactionSymbol = " x "
         R2code = "R2"
+        R2_within_code = "R2 Within"
         lbcode = "\n"
 
     for model in models:
@@ -262,6 +265,11 @@ def etable(
             r2_list.append(_number_formatter(model._r2, **kwargs))
         else:
             r2_list.append("-")
+
+        if not np.isnan(model._r2_within):
+            r2_within_list.append(_number_formatter(model._r2_within, **kwargs))
+        else:
+            r2_within_list.append("-")
 
         if model._vcov_type == "CRV":
             se_type_list.append("by: " + "+".join(model._clustervar))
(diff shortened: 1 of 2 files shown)


AronNemeth:
Shall I also add the adj R2 values to etable? If yes report them automatically, or only when user asks for them by passing a new parameter to pf.etable()?

s3alfisc:
Looks like R2U is down at the moment so the CI fails.

pyfixest/report/summarize.py
```diff
@@ -292,6 +300,8 @@ def etable(
         model_stats_df["S.E. type"] = se_type_list
     model_stats_df[R2code] = r2_list
     n_model_stats = model_stats_df.shape[1]
```
s3alfisc, line 302: PR looks generally good to me, currently building the docs locally. Though not 100% sure about the `if(any)` statment here. What is this check actually doing?

s3alfisc: Ah, I see now. Basically, this is to avoid that the blank "-" is displayed when there is no fixed effect? 

I would rather follow r-fixest, which prints "-" for models without fixed effects if at least one model passed to etable has a fixed effect. 

If there is no fixed effect in any model, we would print adjusted R2. 

Examples: 

```r
library(fixest)
data(mtcars)
mtcars

fit = feols(mpg ~ disp, data = mtcars)
fit2 = feols(mpg ~ disp | gear, data = mtcars)

etable(fit)
# Dependent Var.:                 mpg
# 
# Constant           29.60*** (1.230)
# disp            -0.0412*** (0.0047)
# _______________ ___________________
# S.E. type                       IID
# Observations                     32
# R2                          0.71834
# Adj. R2                     0.70895
# ---
#   Signif. codes: 0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1

etable(fit, fit2)
# fit             fit2
# Dependent Var.:                 mpg              mpg
# 
# Constant           29.60*** (1.230)                 
# disp            -0.0412*** (0.0047) -0.0408 (0.0152)
# Fixed-Effects:  ------------------- ----------------
#   gear                             No              Yes
# _______________ ___________________ ________________
# S.E. type                       IID         by: gear
# Observations                     32               32
# R2                          0.71834          0.71847
# Within R2                        --          0.50683
# ---
#   Signif. codes: 0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
```
So maybe we could simply drop the if clause mentioned above and then tackle the adjusted r2 for etable where no model has a fixef in a separate PR?

s3alfisc:
It surely looks great! 
image

s3alfisc (approved):

pyfixest/report/summarize.py
AronNemeth: Makes sense, so should I just open a different PR adding the adjusted R2 to etable without any FE models?

s3alfisc: You can decide, either we do it in this one or in a separate pr. Your call!

AronNemeth:
Seemed more straightforward to add the adj. R2 in this PR rather than opening a new one.
I added the adj R2 to etable in the following way. If none of the models have FEs, than R2 and adj R2 is shown, if at least one of the models have FE, than R2 and R2 within is shown. Models which doesn't have R2 scores both R2 and adj. R2 is shown with "-" values.

```python
fit = pf.feols("Y ~ X1", data=data, weights="weights")
fit_fe = pf.feols("Y ~ X1 | f1", data=data, weights="weights")
pf.etable([fit, fit_fe])
```
This would produce this output:
image

Models without FEs:

```python
fit1 = pf.feols("Y ~ X1", data=data, weights="weights")
fit2 = pf.feols("Y ~ X1 + X2", data=data, weights="weights")
pf.etable([fit1, fit2])
```
image

PR #843, opened 2025-03-15
Gelbach optimization

s3alfisc:
Speed up the gelbach estimation by sparsifying more operations.

### DIFF
--- a/.github/workflows/extended_tests.yaml
+++ b/.github/workflows/extended_tests.yaml
@@ -1,13 +1,20 @@
 name: Extended Tests (Optional)
 
 on:
-  workflow_dispatch:
-  schedule:
-    - cron: '35 12 * * 6'  # Runs every Saturday at 12:35 UTC (13:35 CET, 14:35 CEST in summer)
-jobs:
-  extended_tests:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


s3alfisc:
Quite happy with this, much improved runtime and no more oom errors: 

```python

import pyfixest as pf
import pandas as pd
import numpy as np
import re
import time

N = 10_000_000
data = pd.DataFrame({})
rng = np.random.default_rng(12345)
data["f1"] = rng.choice(10000, size = N)
data["f2"] = rng.choice(20, size = N)
data["X1"] = rng.normal(size = N)
data["Y"] = rng.normal(size = N)

fit = pf.feols("Y~X1  | f1 + f2", data=data)

# new sparse implementation 

tic = time.time()
res = fit.decompose(
    param = "X1", 
    only_coef=True, 
    agg_first = True, 
    combine_covariates = {
        "f1": re.compile("f1"), 
        "f2": re.compile("f2")
    }
)
toc = time.time()
print("Time elapsed: ", toc-tic)
# 38.50720739364624

# old semi-sparse implementation  
# MemoryError: Unable to allocate 747. GiB for an array with shape (10000000, 10020) and data type float64
```

PR #848, opened 2025-03-24
Add context capture to prediction with new data

adamvig96:
When a model was fitted using some functions in the formula from a given context, a prediction on new data fails as predict does not have access to the function in the context.


adamvig96:
pre-commit.ci autofix

s3alfisc (approved):
looks great, thank you so much @adamvig96 !

PR #878, opened 2025-04-22
Implement Rust Backend

s3alfisc:
- Implement a Rust Backend for PyFixest.
- Reorganize demeaner_backend code logic.


schroedk:
@s3alfisc it was nice to meet you at the PyCon some weeks ago (we met at the appliedAI booth together with @janosg). I have some time to work on this topic and I would have some questions, before starting. First of all, do you want the user to choose the backend? Or is it a question of benchmarking, i.e. if rust implementation is faster everywhere, then get rid of the numba dependency completely?

s3alfisc:
Hi @schroedk, that's super awesome and I'm really excited! The idea was to initially support rust as one of three backends - we currently support numba (which is the default) and JAX for runs on the GPU. There I would like to add rust as an option. Eventually (maybe even directly after merging this pr), I would like to make Rust the default backend, still support JAX, and deprecate numba after a while.

src/collinear.rs
```diff
@@ -0,0 +1,123 @@
...
+
+
+/// Detects collinear variables in a matrix using a modified Cholesky decomposition.
+///
+/// This algorithm identifies which columns in the input matrix can be expressed
+/// as linear combinations of other columns, within the specified tolerance.
+///
+/// # Arguments
+///
+/// * `x` - Input matrix (must be square, typically a correlation or covariance matrix)
```
s3alfisc, line 26:
```suggestion
/// * `x` - Input matrix (must be square, typically X'X in a regression model, where X is the N x k design matrix)
```

s3alfisc:
@schroedk just as a status update - I am currently reviewing the rust implementations; only have to go through the "count_fixef_fully_nested". Hopeful I'll be able to merge this in a few days =)

s3alfisc:
Ok, finished my review + added some docstrings. Only thing to add (I think) is update the readme to "blazing fast" and "written in Rust" 😄

s3alfisc:
@all-contributors please add @schroedk for code

PR #885, opened 2025-04-28
JOSS Paper for PyFixest

s3alfisc:
First draft of a JOSS paper for pf.

### DIFF
--- a/joss_paper/pyfixest_joss.md
+++ b/joss_paper/pyfixest_joss.md
@@ -1,0 +1,65 @@
+---
+title: 'PyFixest: A Python Port of fixest for High-Dimensional Fixed-Effects Regression'
+date: 27 April 2025
+author:
+  - name: Alexander Fischer
+    affiliation: Trivago
+---
+
+
+## Summary
+
+PyFixest is an open-source Python library that implements efficient routines for regression analysis with multiple potentially high-dimensional fixed effects by applying the Frisch-Waugh-Lovell theorem. It is a faithful port of the R package fixest [@berge2018], aiming to replicate fixest’s core design principles and functionality within the Python ecosystem. Users familiar with fixest in R can seamlessly transition their analysis to Python using the same syntax and obtaining identical results. Likewise, users of PyFixest should find it easy to port their analysis to R and use the fixest package.
+
+## Statement of Need
+
+Fixed-effects models with high-dimensional categorical effects are ubiquitous in the social sciences, where researchers often need to control for multiple levels of unobserved heterogeneity. To efficiently estimate such models, researchers typically rely on specialized software that handle the computational challenges associated with high-dimensional fixed effects via the application of the Frisch-Waugh-Lowell (FWL) theorem (@correia2023reghdfe and correia2020fast in Stata, @gaure2013lfe, @berge2018 and @stammann2020package in R, @fixedeffectmodelsjl_2025 in Julia). In practice, the FWL theorem is efficiently implemented by an iterative "within-transformation" (demeaning) approach, which avoids the need to create large matrices of dummy variables, which can be computationally expensive and memory-intensive, especially with large datasets and/or many fixed effects [see @gaure2013ols for details].
+
+The Python eco-system lacks an optimized library for this purpose, and *PyFixest* aims to fill this gap. *statsmodels*, the dominant statistical regression library in Python [seabold2010statsmodels] does not provide OLS and GLM functionality that natively handles multiple group fixed effects in an optimized and efficient way via the FWL theorem. Instead, users either have to inefficiently one-hot encode all categorical features in large dummy matrices, or have to rely on a workaround that involves demeaning the design matrix and outcome variable via the *pyhdfe* library (@gortmaker_pyhdfe_2023) before passing the transformed data to *statsmodels* OLS class for fitting the model. GLM estimation is generally not supported via this two-step procedure, as the iterated least squares (ILS) approach used to fit such models requires that design matrix and outcome variable are demeand in every iteration (@correia2020fast, stammann2017fast). Another package, *linearmodels*  provides a class for estimating linear models with multiple fixed effects, *AbsorbingLS*, which calls *pyhdfe* to perform the demeaning step.
+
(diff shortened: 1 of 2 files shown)


juanitorduz:
Maybe we can use a quarto document ? (not strong option, but I think it compiles to latex right?)

joss_paper/pyfixest_joss.md
```diff
@@ -0,0 +1,64 @@
...
+date: 8 August 2025
+author:
+  - name: Alexander Fischer
+    affiliation: Trivago
+---
+
+
+## Summary
+
+PyFixest is an open-source Python library that implements efficient routines for regression analysis with multiple potentially high-dimensional fixed effects by applying the Frisch-Waugh-Lovell theorem. It is a faithful port of the R package fixest [@berge2018], aiming to replicate fixest’s core design principles and functionality within the Python ecosystem. Users familiar with fixest in R can seamlessly transition their analysis to Python using the same syntax and obtaining identical results. Likewise, users of PyFixest should find it easy to port their analysis to R and use the fixest package.
```
juanitorduz, line 12: Is there anything else PyFixEst does that fixest does not? Or could we mention the scale we support (say, to run on GPU). I think is important to emphasize it's a port but also because of a reason: the scale and how these models can integrate better in production-like environments (I know, I know you can deploy R, but all the data pipelines are in Python, so it's definitely an advantage)

joss_paper/pyfixest_joss.md
```diff
@@ -0,0 +1,64 @@
...
+
+*PyFixest* aims to faithfully implement *fixest*'s core functionality - efficient routines for OLS, IV, and Poisson regression with fixed effects - syntax, and post-estimation functionality. Identical input arguments to the main estimation functions that both packages share - *feols*, *fepois* and *feglm* - should produce identical results in R and Python. All of *fixest*'s core defaults, including the choice of variance covariance matrices, small sample corrections, handling of singleton fixed effects, and the treatment of multicollinear variables are preserved by *PyFixest*. To ensure identical behavior, both libraries are thoroughly tested against each other using rpy2 (gautier2008rpy2).
```
juanitorduz, line 22: Do we expect readers to know what VI is? We could write instrumental variables (VI)

joss_paper/pyfixest_joss.md
```diff
@@ -0,0 +1,64 @@
...
...enker2001quantile) and a range of modern event study estimators, including the linear projections approach (@dube2023local, @busch2023lpdid), the two-stage difference-in-differences imputation estimator (@gardner2022two, @butts2021did2s), and the fully-saturated event study estimator proposed by @sun2021estimating. The package provides comprehensive options for calculating non-standard inference, including cluster robust variance estimators (CRV1 and CRV3, see @mackinnon2023fast), wild cluster bootstrap (@roodman2019fast, @fischer2022fwildclusterboot), randomization inference (@hess2017randomization), and the causal cluster variance estimator (@abadie2023should). PyFixest also implements methods to control the family-wise error rate by implementing the Romano-Wolf correction (@clarke2020romano and @romano2005exact), and enables users to compute simultaneous confidence bands through a multiplier bootstrap approach (@montiel2019simultaneous). Additionally, *PyFixest* provides support for Gelbach's regression decomposition (@gelbach2016covariates), Lal's event study specification test (@lal2025can), and estimation strategies based on compression & sufficient statistics as described in
```
juanitorduz, line 24: I think we are missing key points:

- The scale (JAX on GPU for example, maybe we can even show some benchmarks)
- We also work on documentation to make the content accessible to the Python community. 
- Support multiple plotting backends (matplotlib, letsplot)
- It's a community project with ca 40 contributors. We work to have clear "first-time issues" to ensure a welcoming environment for development. 

We could maybe also add a teaser about the plans for the future?

joss_paper/pyfixest_joss.md
```diff
@@ -0,0 +1,64 @@
+---
+title: 'PyFixest: A Python Port of fixest for High-Dimensional Fixed-Effects Regression'
```
juanitorduz, line 2: I would add something more to the title, say something along the lines 

`PyFixest: A Python scalable Port of fixest for High-Dimensional Fixed-Effects Regression`

joss_paper/pyfixest_joss.md
```diff
@@ -0,0 +1,64 @@
...
+date: 8 August 2025
+author:
+  - name: Alexander Fischer
+    affiliation: Trivago
+---
+
+
+## Summary
```
juanitorduz, line 10: I know it's hard, but we should somehow prove the already important adoption of the package. Either by the number of downloads and / or some testimonies of companies that use it (I can testify we used it for price experimentation at Wolt)

joss_paper/pyfixest_joss.md
apoorvalal: Or get rid of the "this is a port" billing entirely; it isn't a port in any meaningful sense since it doesn't really use fixest's source code or port over its design principles. Suggest changing it to "pyfixest: A library for High-dimensional Fixed-Effects Regression in Python".

joss_paper/pyfixest_joss.md
s3alfisc: Yes, I think testimonies would be awesome! But very hard to get, even if we find users that confirm that pf is used in their companies, we might have to clear this with their respective legal teams? Or don't you think so?

joss_paper/pyfixest_joss.md
s3alfisc: > Suggest changing it to "pyfixest: A library for High-dimensional Fixed-Effects Regression in Python".

Yep, I think this is a good suggestion. I'd keep the reference to fixest as I still think that we borrow a lot from its API and default, but also no need to make it the core point of the paper, as by now we support quite some extra things not native to fixest (gpu, qr, compression, some DiD, multiple tesrting corrections).

joss_paper/pyfixest_joss.md
s3alfisc: Yep, makes sense, I'll add all of these!

Issue #890, opened 2025-05-01
Two-way clustered standard errors don't match Stata or statsmodels?


jestover:
I recently came across the this package which seems fantastic.  When trying a regression package I ofter try to replicate the results using Mitchell Petersen's [data for testing standard error estimation programs](https://www.kellogg.northwestern.edu/faculty/petersen/htm/papers/se/test_data.htm) to make sure I am using the correct syntax and getting the results I would expect.  Everything works as I would expect it to until trying to do two way clustering by `firmid` and `year`.  Here is a minimal reproducible example.

```{python}
import polars as pl
import statsmodels.formula.api as smf
import pyfixest as pf

# Read in the Petersen data from a fixed width text file
df = pl.read_csv("petersen-data.txt", has_header=False, new_columns=["full_str"])
df = df.with_columns(
    firmid=pl.col("full_str").str.slice(0, 18).str.strip_chars(" ").cast(pl.UInt16),
    year=pl.col("full_str").str.slice(18, 4).str.strip_chars(" ").cast(pl.UInt8),
    x=pl.col("full_str").str.slice(22, 11).str.strip_chars(" ").cast(pl.Float32),
    y=pl.col("full_str").str.slice(33, 9).str.strip_chars(" ").cast(pl.Float32),
).drop("full_str")

# Using statsmodels
smf.ols("y ~ x", data=df).fit(
    cov_type="cluster", cov_kwds={"groups": df.select(["firmid", "year"]).to_numpy()}
).summary()

# Using pyfixest
pf.feols("y ~ x", data=df, vcov={"CRV1": "firmid + year"}).summary()

```

The `statsmodels` result would give standard errors of 0.0651 and 0.0536, which match what Petersen reports whereas `pyfixest` gives standard errors of 0.0681 and 0.0553.

Is the two way clustering in `pyfixest` supposed to be different?  If so, does this clustering method have more desirable properties? Is there a way to get the same results as Petersen and `statsmodels` using `pyfixest`?  Thanks in advance for any guidance.

jestover:
After some digging and querying ChatGPT I was able to figure out how to replicate the Petersen results.  It has to do with small sample correction.  For anyone who is running into similar questions, there is fantastic documentation on `pyfixest` small sample correction at [https://py-econometrics.github.io/pyfixest/ssc.html](https://py-econometrics.github.io/pyfixest/ssc.html).  To get the Petersen results you can do

```{python}
import polars as pl
[...]

PR #897, opened 2025-05-09
#890 - Show how to replicate Stata results

jestover:
.qmd file showing how to replicate common Stata results in PyFixest.

Note: I ran each part in both Stata and PyFixest and got the same results everywhere except for `fit4` where I show how to get HC3 standard errors.  For some reason the HC3 errors in PyFixest and Stata do not match.  I haven't had a chance to try to figure out why though.

If you know why, let me know and I can update the example.  If not, I can try to figure it out at a later date.  In the meantime we can remove that section or add a note that the HC3 standard errors do not match.

### DIFF
--- a/docs/stata-2-pyfixest.qmd
+++ b/docs/stata-2-pyfixest.qmd
@@ -1,0 +1,226 @@
+---
+title: "Translating Stata to PyFixest"
+format:
+  html:
+    html-table-processing: none
+toc: true
+toc-title: "On this page"
+toc-location: left
+---
+
+# How to Get Started
+
+This guide will focus on how to replicate the regression results you would get in Stata
+with the Python package `pyfixest` and assumes you know how to do things like install 
+Python packages and load data into Pandas.  For a broader introduction to doing 
+econmetrics in Python you might check out Arthur Turrell's 
+[Coding for Economist](https://aeturrell.github.io/coding-for-economists/intro.html), 
+which includes a section on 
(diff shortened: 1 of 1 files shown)


s3alfisc:
>  For some reason the HC3 errors in PyFixest and Stata do not match. I haven't had a chance to try to figure out why though.

This is just a suspicion, but it might be that we might have to set `ssc(adj = False, cluster_adj = True)` for HC3 - this seems to be the canonical correction. Snippet below from the Zeileis at al [Various Versatile Covariances paper](file:///C:/Users/alexa/Downloads/v95i01-18.pdf): 
image

If that is it, I'd say we update the PR, if not, I would leave it in and just state that HC3 does not match. Best to be transparent I think?

s3alfisc:
General comment on the PR - looks very good, thanks! Do you think it would be a lot of work to add the Stata output to the code chunk as commented out text? So that readers could directly compare the Stata and R outputs?

s3alfisc:
pre-commit.ci autofix

s3alfisc:
@all-contributors please add @jestover for docs

jestover:
You were correct on HC3 correction, so that is fixed and I have added the Stata output which should make it so people can follow along without having to replicate everything themselves.

s3alfisc:
pre-commit.ci autofix

s3alfisc (approved):
Perfect, looks great. Thanks so much @jestover!

s3alfisc:
Looks great, link here: https://py-econometrics.github.io/pyfixest/stata-2-pyfixest.html =)

PR #900, opened 2025-05-12
Prevent Casting Dependent Variable to Integer to Avoid Information Loss in Poisson Regression

JaapCTJ:
This pull request removes the line of code in `fepois_.py` that converts the dependent variable Y to an integer. Poisson regression can be applied to any non-negative outcomes, and there is no need to limit its use to non-negative integer outcomes by converting float Y values into integers. 

This makes the results more consistent with the `Fixest `package in R, where it does not perform this conversion either.

Additionally, it might be nice to add a non-negative non-integer generated dependant variable in the `get_data()` function in `utils.py`. This would demonstrate the versatility of Poisson regression beyond integer-only outcomes and provide users with a broader perspective on its application.

### DIFF
--- a/pyfixest/estimation/fepois_.py
+++ b/pyfixest/estimation/fepois_.py
@@ -140,8 +140,6 @@ class Fepois(Feols):
         "Prepare model inputs for estimation."
         super().prepare_model_matrix()
 
-        # check if Y is a weakly positive integer
-        self._Y = _to_integer(self._Y)
         # check that self._Y is a weakly positive integer
         if np.any(self._Y < 0):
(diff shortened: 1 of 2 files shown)


JaapCTJ:
Instead of removing the `_to_integer()` function entirely, I changed it to `_check_series_or_dataframe()` to keep the essential check that the outcome variable should be a pandas series or dataframe.

s3alfisc:
Looks good to me! Thank you Jaap @JaapCTJ

s3alfisc:
All-contributors please add @JaapCTJ for code

Issue #846, opened 2025-03-16
Switch to `np.linalg.lstsq` as default solver


s3alfisc:
Should we switch to  `np.linalg.lstsq` as the default solver for pf? 

Right now, we use the normal equation solve(X'X)X'y, which should be numerically less stable. 

In my experience, it is sometimes faster than `np.linalg.lstsq`  & sometimes slower (based on some very rough benchmarks). 

Any thoughts @apoorvalal? Maybe we should properly benchmark it first? 

Maybe something you'd be interested to pick up next @AronNemeth?

AronNemeth:
Sure thing, I'll be happy to look into it!

apoorvalal:
I think failure of (X'X) full rankedness might be a more pressing reason to switch to `np.linalg.lstsq`. [This](https://lorentzen.ch/index.php/2021/05/02/least-squares-minimal-norm-solution/) post walks through differences between directly solving the normal equation vs using a smarter solver (lstsq with `gelsd` or `gelsy` solver). 
The modern minimum-norm interpolant lit relies on this pretty heavily. I'd recommend a read through [this](https://arxiv.org/abs/2309.15769)

s3alfisc:
Thanks for all the pointers @apoorvalal, I'll take a look, big gap in my education. I have to say that I haven't been to worried about failing to invert (X'X) for the coef estimates as I thought it would fail for the SEs anyways plus the multicollinearity checks pyfixest runs, but clearly it is better to get precise estimates and imprecise SEs than inprecise estimates of both.

s3alfisc:
I've actually been doing some benchmarks in the last days, in some relevant cases, `np.linalg.lstsq` is **orders** of magnitude slower than directly inverting X'X...

```python
import numpy as np
import pandas as pd
import time
import scipy.linalg
from formulaic import model_matrix

def generate_data(n_users, n_items, n_positions, seed=12):
    rng = np.random.default_rng(seed)
    position_effect = 1 / np.sqrt(np.arange(1, n_positions + 1))
    item_effect     = rng.normal(0, 1, size=n_items)
    scores     = rng.random((n_users, n_items))
    cutoff_idx = n_items - n_positions
    user_items = np.argpartition(scores, cutoff_idx, axis=1)[:, cutoff_idx:]
    badge      = rng.integers(0, 2, size=(n_users, n_positions))
    noise      = rng.normal(0, 1, size=(n_users, n_positions))
[...]

PR #904, opened 2025-05-13
Switch default solver to scipy.linalg.solve() [#846 issue]

AronNemeth:
Related issue [#846](https://github.com/py-econometrics/pyfixest/issues/846)

Changed default solver to Cholesky decomposition implemented in scipy.linalg.solve(A,b, assume_a="pos").


AronNemeth:
I updated the test matrix to be symmetric and positive definite, so that Cholesky Decomposition doesn't fail

s3alfisc (approved):
Will merge despite failing workflows. Not sure what is going on there & will debug separately. Thank you @AronNemeth!

Issue #828, opened 2025-03-08
Reorganize Tests against R


s3alfisc:
The current test suite against R requires users to install R and multiple packages from CRAN into the global R environment. 

Optimally, we would directly install R and all R dependencies available on `conda-forge` directly from `conda-forge` into the pyfixest dev environment, which should help developers to run unit tests against R locally + keep the dev's R installation "clean".

I think it would be best to reorganize the unit tests against R to have the following structure: 
- `tests-against-r-core`: tests against packages available via `conda-forge`: `fixest` and `sandwich`. These are directly added to the `tests` environment as dependencies. 
- `tests-against-r-extended`: tests against R packages not availabe on `conda-forge`: `did2s`, `ritest`, `fwildclusterboot`, etc. 

To do's: 

- [ ] add the following R dependencies to the `tests` environment: `r-fixest`, `r-sandwich`, `r-broom`, R itself
- [ ] create a new pytest marker `against-r-core` & decorate all tests that only depend of `fixest`, `sandwich`, `broom`
- [ ] create a new pytest marker `against-r-extended` & decorate all tests that only depend on "other" r packages
- [ ] adjust the github actions to install fixest, sandwich, broom from conda
- [ ] update the dev guide

shapiromh:
I've struggled enough with using R in python that I could probably help on this one. That said, a smaller scoped test-related issue might be a better first issue if you have a suggestion.

s3alfisc:
Hi @shapiromh , sounds great! I'll have to think a bit about how to best cut this one into smaller issues, will get back to you with details tomorrow =)

shapiromh:
Maybe I can give this a shot anyway? The r-core seems like it should be more straightforward than r-extended tests

s3alfisc:
Hi @shapiromh , I am super sorry, I completely lost track! 😒 Yes, please go ahead with adjusting the r-core tests! Alternatively, if you have some familiarity with HAC SEs, I have a (private) branch that implements panel HAC estimators - maybe that would be interest? If yes I'll make it public. Sorry again!

shapiromh:
Great, I'll try to do that.

As for something on the substantiative side -- I think I should open a different issue -- but I was trying and failed to introduce (p)weights in the new Sunab implementation. I think it would have just been a matter of creating an option to pass the weights argument in the event_study to the underlying estimators, but I ended up reverting to a wrapper on the R code.

s3alfisc:
[...]

PR #906, opened 2025-05-18
Reorganize Tests against R

shapiromh:
This is a PR related to #828. 

**Summary of Code Changes:**
- R and conda-forge available R packages ("core") are now installed as part of the dev environment. Other R packages ("extended") are still installed via the "r_test_requirements.R" method and pixi task. The doc environment was also updated to install R via conda so users (I think) do not need their own out-of-project R installation. 
- All tests that call R packages (even if the package is only called on a path that cannot be taken) are labeled as part of "against_r_core" or "against_r_extended. 
- Added a small utility to check for installed R packages. Tests that rely on extended R packages will be skipped with an informative message if the related package is not installed. This is similar to how the visualization tests are handled if dependencies are not available. 
- Updated the contribution docs to reflect the changes above. 
- [!] Updated the github actions, but I'm not sure if that was done correctly. 

**Some Notes:**
1. I am not sure if the doc environment also relies on "extended R packages". If so the updates may break that code. When installing the extra R packages, I am fairly sure they will only be available in the dev and not doc environment. 
2. I think `utils/set_rpy2_path.py` could be deprecated. The updates actually remove all references to this code, but then there is no built in way to extend the check R module libraries beyond the project install. 
3. I had a very tough time installing "ivDiag" as it has a dependency on "lfe" that has a configuration not updated for modern Macs. The configuration can be adjusted, but this requires a lot of system-specific configuration.
4. I found it was almost feasible to have a completely self-contained R environment for the user without dependencies outside of the project. The issue is some of the extended R packages that rely on gcc or cmake. Cmake can also be installed via conda-forge. Gcc can as well, but requires pulling different versions for different systems (this seems doable in pixi, but I can only test the Mac version). Finally, there is still the issue of installing ivDiag.

### DIFF
--- a/r_test_requirements.R
+++ b/r_test_requirements.R
@@ -1,7 +1,11 @@
 # note: R, fixest, sandwich, broom are installed via conda
 install.packages(
-    c('did2s', 'wildrwolf', 'reticulate', 'ivDiag', 'car'),
+    c('did2s', 'reticulate', 'ivDiag', 'car'),
     repos='https://cran.rstudio.com'
+);
+install.packages(
+   c('collapse', 'summclust', 'wildrwolf'),
+    repos = c('https://s3alfisc.r-universe.dev', 'https://cloud.r-project.org', 'https://fastverse.r-universe.dev')
 );
 install.packages(
     'ritest',
--- a/r_test_requirements.R
+++ b/r_test_requirements.R
@@ -1,6 +1,6 @@
 # note: R, fixest, sandwich, broom are installed via conda
 install.packages(
-    c('did2s', 'reticulate', 'ivDiag', 'car'),
+    c('did2s', 'reticulate', 'ivDiag'),
     repos='https://cran.rstudio.com'
 );
 install.packages(
--- a/r_test_requirements.R
+++ b/r_test_requirements.R
@@ -1,6 +1,6 @@
-# note: fixest and broom are installed via conda
+# note: R, fixest, sandwich, broom are installed via conda
 install.packages(
-    c('fixest', 'broom','clubSandwich', 'did2s', 'wildrwolf', 'reticulate', 'ivDiag', 'stats', 'base', 'car'),
+    c('did2s', 'wildrwolf', 'reticulate', 'ivDiag', 'car'),
     repos='https://cran.rstudio.com'
 );
 install.packages(
--- a/tests/test_did.py
+++ b/tests/test_did.py
@@ -1,4 +1,5 @@
 from importlib import resources
+from pathlib import Path
 
 import numpy as np
 import pandas as pd
@@ -22,6 +23,9 @@ fixest = importr("fixest")
 # Extended Packages
 if import_check := check_r_install("did2s", strict=False):
     did2s = importr("did2s")
+# Check mpdata avaialibility
+MPDATA_LOC = "C:/Users/alexa/Documents/pyfixest-zalando-talk/mpdta.csv"
+mpdata_check = Path(MPDATA_LOC).is_file()
 
 
 @pytest.fixture
@@ -362,9 +366,9 @@ def test_fully_interacted(unit, cluster):
 
 
 @pytest.mark.against_r_core
-@pytest.mark.skip("mpdata not available online as csv, only run test locally.")
+@pytest.mark.skipif(mpdata_check is False, reason="mpdata not available online as csv, only run test locally.")
 @pytest.mark.parametrize(
-    "mpdata_path", [r"C:/Users/alexa/Documents/pyfixest-zalando-talk/mpdta.csv"]
+    "mpdata_path", [MPDATA_LOC]
 )
 def test_fully_interacted_mpdata(mpdata_path):
     mpdata = pd.read_csv(mpdata_path)
--- a/tests/test_did.py
+++ b/tests/test_did.py
@@ -366,10 +366,11 @@ def test_fully_interacted(unit, cluster):
 
 
 @pytest.mark.against_r_core
-@pytest.mark.skipif(mpdata_check is False, reason="mpdata not available online as csv, only run test locally.")
-@pytest.mark.parametrize(
-    "mpdata_path", [MPDATA_LOC]
+@pytest.mark.skipif(
+    mpdata_check is False,
+    reason="mpdata not available online as csv, only run test locally.",
 )
+@pytest.mark.parametrize("mpdata_path", [MPDATA_LOC])
 def test_fully_interacted_mpdata(mpdata_path):
     mpdata = pd.read_csv(mpdata_path)
     mpdata["first_treat"] = mpdata["first.treat"]
--- a/tests/test_did.py
+++ b/tests/test_did.py
@@ -12,9 +12,9 @@ from rpy2.robjects.packages import importr
 import pyfixest as pf
 from pyfixest.did.estimation import did2s as did2s_pyfixest
 from pyfixest.did.estimation import event_study, lpdid
-from pyfixest.utils.set_rpy2_path import update_r_paths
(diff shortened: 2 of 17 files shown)


s3alfisc:
pre-commit.ci autofix

s3alfisc:
Hi @shapiromh, thanks so much for this! Looks good at first sight but I have to / will spend some more time on this tomorrow morning. 

Re your comments: 


> I am not sure if the doc environment also relies on "extended R packages". If so the updates may break that code. When installing the extra R packages, I am fairly sure they will only be available in the dev and not doc environment.

Yes, some of the docs do - the "[fixest vs pyfixest](https://py-econometrics.github.io/pyfixest/compare-fixest-pyfixest.html)" vignette depends on the R core dependencies. So we'd have to add all of these to the docs deps. 

> I think utils/set_rpy2_path.py could be deprecated. The updates actually remove all references to this code, but then there is no built in way to extend the check R module libraries beyond the project install.

Yeah, I think it was an util I set up because R deps needed to be installed in the global R env and this should ensure that python could talk to the global R env? Though not 100% sure anymore. Let's see if we can make things run without (I would think so) and then I'd be happy to delete the function from the code base. 

> I had a very tough time installing "ivDiag" as it has a dependency on "lfe" that has a configuration not updated for modern Macs. The configuration can be adjusted, but this requires a lot of system-specific configuration.

Hm, I think for the ivDiag tests, could we simply run them once, store results in a csv file, and then drop it as a dependency? I think this would be the general strategy for all other non-conda packages as well - we could provide a script that calls R and produces a csv with "R results", which we store and test against. This way, all results would be reproducible (though not perfectly) and users / testers would not have to install ivDiag, fwildclusterboot, wildrwolf, ritest etc? This should then also solve the issue you mention in your last point?

> I found it was almost feasible to have a completely self-contained R environment for the user without dependencies outside of the project. The issue is some of the extended R packages that rely on gcc or cmake. Cmake can also be installed via conda-forge. Gcc can as well, but requires pulling different versions for different systems (this seems doable in pixi, but I can only test the Mac version). Finally, there is still the issue of installing ivDiag.

shapiromh:
> Yes, some of the docs do - the "[fixest vs pyfixest](https://py-econometrics.github.io/pyfixest/compare-fixest-pyfixest.html)" vignette depends on the R core dependencies. So we'd have to add all of these to the docs deps.

Sorry, I was asking if non core R dependencies (those not in conda) are  used in the docs. I did up the docs environment to install R in the toml, but I probably messed up the github actions if any changes around that were needed.

> Hm, I think for the ivDiag tests, could we simply run them once, store results in a csv file, and then drop it as a dependency? I think this would be the general strategy for all other non-conda packages as well - we could provide a script that calls R and produces a csv with "R results", which we store and test against. This way, all results would be reproducible (though not perfectly) and users / testers would not have to install ivDiag, fwildclusterboot, wildrwolf, ritest etc? This should then also solve the issue you mention in your last point?

ivDiag is the only one that caused me problems, but generally this strategy makes sense to me. I saw this was the approach already taken with some of the tests (and why some of the test codes never went down paths that called the R packages). I would think the only argument against is if any of these non-conda packages are under active development with known bugs , but then you or other core maintainers should probably hold the single source of truth on what other contributors should be matching.

r_test_requirements.R
```diff
@@ -1,8 +1,12 @@
-# note: fixest and broom are installed via conda
+# note: R, fixest, sandwich, broom are installed via conda
 install.packages(
-    c('fixest', 'broom','clubSandwich', 'did2s', 'wildrwolf', 'reticulate', 'ivDiag', 'stats', 'base', 'car'),
+    c('did2s', 'reticulate', 'ivDiag', 'car'),
```
s3alfisc, line 3: We should also be able to install [r-car via conda-forge](https://anaconda.org/r/r-car).

tests/test_did.py
```diff
@@ -33,6 +34,8 @@ def data():
     return df_het
 
 
+@pytest.mark.skipif(import_check is False, reason="R package did2s not installed.")
```
s3alfisc, line 37: This is an excellent solution, big fan =)

tests/test_did.py
```diff
@@ -355,6 +361,7 @@ def test_fully_interacted(unit, cluster):
     )
 
 
+@pytest.mark.against_r_core
```
s3alfisc, line 364: This might fail in the CI as I don't think the data is available on github? It's from the did R repo: https://github.com/bcallaway11/did/tree/master/data When I wrote the test, I did not want to think about how to load dta files into pandas, and I did not simply want to copy over the file as I wasn't sure about the license. `did` uses [GPL-2](https://github.com/bcallaway11/did/blob/32a1da2427aa986c45c2b714847b8464d64ed72b/DESCRIPTION#L9C9-L9C15) which I think (?) does not allow reuse under any other license then GPL-2.

s3alfisc (commented):
Looks good to me, except for two small comments - one data set is not available at runtime in the CI & the car package can be installed from conda directly. Though, maybe it is cleaner to ask users to download it manually as it is not part of the "core" R test dependencies?

s3alfisc:
The CI currently fails because we have to rebuild the pixi lock file after updating the pyproject toml. Will do this, one moment =)

s3alfisc:
Now let's see if the CI runs - I recall that I struggled in the past: I've been installing r and packages from CRAN via [r2u](https://github.com/eddelbuettel/r2u), which provides ubuntu binaries of R packages (so no need for manual installation!). But I recall that the conda env and the R installation provided by r2u had problems talking to each other; that's why I eventually settled for the solution to "install all R depcs via r2u and none via conda".

s3alfisc:
@all-contributors please add @shapiromh for code

s3alfisc:
> Sorry, I was asking if non core R dependencies (those not in conda) are used in the docs. I did up the docs environment to install R in the toml, but I probably messed up the github actions if any changes around that were needed.

Sorry, I got it wrong then! But I think no changes are needed, only fixest and broom need to be available (and we could drop this requirement as well). 

> ivDiag is the only one that caused me problems, but generally this strategy makes sense to me. I saw this was the approach already taken with some of the tests (and why some of the test codes never went down paths that called the R packages). I would think the only argument against is if any of these non-conda packages are under active development with known bugs , but then you or other core maintainers should probably hold the single source of truth on what other contributors should be matching.

Yes, I agree - imo there are two arguments for running code via rpy2 - the first is to make sure that any bugs fixed via dependencies are eventually caught; I'd also want to know in case `fixest` syntax or defaults changed. The other (weaker) reason is that adjusting code that is directly run via rpy2 might be easier, so adding / adjusting tests might be less work? 

For the tests where we currently follow this strategy, I think the main reason was time - checking `ritest` and the bootstrap based multiple testing methods simply took too much time.

shapiromh (commented):
I think there are some changes here I could make and resubmit:
1) (Maybe) Add a check on whether the data is available and skip the test if not
2) Add cars as a core package
3) Add back "extended" test marks I may have inadvertently removed. 

I'm not sure what the conclusion was on the failing CI because of R...

s3alfisc:
All makes sense! If you address all three we can merge. On the CI- I think it's just a CI failure, these sometimes happen. I think it will go away by itself.

shapiromh:
On the 3: 
- I added a check on the availability of the local mpdata. There's probably a more elegant way than hard coding the path, but this doesn't cause any errors...
- Add r-car as a core package and changed the tests to core. 
- In CCV all the tests are labeled extended. I had already added the mark to one of the tests after removing it from the fixture.

s3alfisc:
pre-commit.ci autofix

s3alfisc:
Looks very good and I can merge now. Thanks for your first PR to pyfixest @shapiromh!

PR #909, opened 2025-05-22
add an example for etable

mortizm1988:
Add an example for `etable`, changing the reference category in an interaction term.
It also shows how the `_coefnames` change, as this is needed for updating labels for plots and tables.


s3alfisc (approved):
Looks good, thank you @mortizm1988!

Issue #919, opened 2025-05-28
Allow event study methods to take aweights argument.


shapiromh:
This is a separate issue to related to a discussion in #828 about allowing analytic weights to the saturated event study method. 

Although most methods calling subclasses of DID are estimated using feols, there was no public interface to supply analytic weights (though enabled for feols). We should be able to open this interface to all DiD subclasses; LPDID is the only exception that may not be a minor change. 

I will supply a PR for this issue shortly.

PR #920, opened 2025-05-28
Add weights option to event_study

shapiromh:
This PR opens the (analytic) weight option to the did/event_study function and all the subclasses of DiD it currently calls (twfe, did2s, and saturated_twfe). The related issue is #919 

Changes in this PR: 
- The DiD2S class already permitted weights so the new event_study function simply passes weights to it. No new test was added for this change. 
- The TWFE class was opened to accept weights. They are passed through to the underlying FEOLS call. One test was added on top of existing TWFE event study tests. 
- The Saturated Event Study was similarly opened to accept weights. It required a substantive change to use these weights in its aggregation method. New tests were added comparing it to fixest sunab with weights. 
- The ABC DID now expects weights as an argument. LPDID is the only subclass of DID that still does not allow weights.

### DIFF
--- a/pyfixest/did/saturated_twfe.py
+++ b/pyfixest/did/saturated_twfe.py
@@ -347,7 +347,13 @@ def _saturated_event_study(
                 {"+".join([f"i(rel_time, {x}, ref = -1.0)" for x in cohort_dummies.columns.tolist()])}
                 | {unit_id} + {time_id}
                 """
-    m = feols(fml=ff, data=df_int, weights=weights, vcov={"CRV1": cluster}, weights_type = "aweights")  # type: ignore
+    m = feols(
+        fml=ff,
+        data=df_int,
+        weights=weights,
+        vcov={"CRV1": cluster},
+        weights_type="aweights",
+    )  # type: ignore
     res = m.tidy()
     # create a dict with cohort specific effect curves
     res_cohort_eventtime_dict: dict[str, dict[str, pd.DataFrame | np.ndarray]] = {}


shapiromh:
I also added a fix for the saturated event study not being able to call summarize because of an out-of-date check in the summarize code.

s3alfisc:
Awesome, thanks so much! Will review this and your other PR first thing tomorrow morning!

pyfixest/did/saturated_twfe.py
```diff
@@ -411,35 +431,60 @@ def compute_period_weights(
         Column name of treatment indicator (0/1).
     include_grid : bool, default True
         If True, returns a full (cohort x period) grid with zero-filled weights.
+    weights : Optional[str], default None
```
s3alfisc, line 434: Should we here not simplify to using the weights if they are provided?

pyfixest/did/saturated_twfe.py
```diff
@@ -151,23 +158,31 @@ def test_treatment_heterogeneity(self) -> pd.Series:
         )
 
     def aggregate(
-        self, agg="period", weighting: Optional[str] = "shares"
+        self,
+        agg="period",
+        use_weights: bool = True,
```
s3alfisc, line 163: Asking the same question three times, sorry for that (😅 )- would it not make more sense to assume that if weights have been provided in the model fit, they should also be applied in the aggregation?

pyfixest/did/saturated_twfe.py
```diff
@@ -411,35 +431,60 @@ def compute_period_weights(
...
     post["weight"] = post["n_grel"] / post["n_rel"]
 
     # pre-treatment cells (l < 0)
-    pre = (
-        df[(df[treatment] == 0) & (df[cohort].isin(ever_treated))]
-        .groupby([cohort, period])
-        .size()
-        .reset_index(name="n_grel")
-    )
+    if weights is not None and use_weights:
```
s3alfisc, line 474: For some reason, no tests seem to be triggered for the aggregation method with weights, could you please add them to the [saturated_twfe tests](https://github.com/py-econometrics/pyfixest/blob/bc423748c50cf3f8297254e843813f787f1d3a98/tests/test_did.py#L338)?

s3alfisc (changes requested):

pyfixest/did/estimation.py
```diff
@@ -15,17 +15,19 @@ def event_study(
     idname: str,
     tname: str,
     gname: str,
-    xfml: Optional[str] = None,
     cluster: Optional[str] = None,
-    estimator: Optional[str] = "twfe",
+    weights: Optional[str] = None,
+    xfml: Optional[str] = None,
     att: Optional[bool] = True,
+    estimator: Optional[Literal["did2s", "twfe", "saturated"]] = "twfe",
```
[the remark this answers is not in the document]
shapiromh: Sure. Let me try and have a think about this and dealing with the ATT item in a separate PR.

shapiromh (commented):
Thanks for the comments, let me make sure the weighted test covers the saturated twfe function and update the PR. 

The "use_weights" issue is down to your taste, I think. I added to match fixest, but I also don't see the purpose in having this as a separate argument if the user already supplied weights. 

I think the questions on freq versus analytic comes back to the discussion we were having about guessing which it was fixest used. At least feeding the argument as analytic to feols matched the std. errors from fixest.

pyfixest/did/saturated_twfe.py
```diff
@@ -411,35 +431,60 @@ def compute_period_weights(
...
-    post = (
-        df[df[treatment] == 1]
-        .groupby([cohort, period])
-        .size()
-        .reset_index(name="n_grel")
-    )
+    if weights is not None and use_weights:
+        post = (
+            df[df[treatment] == 1]
+            .groupby([cohort, period])[weights]
```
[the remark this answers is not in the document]
s3alfisc: Based on the fixest docs - if `use_weights = True`, then the aggregation using weights and the weights are treated as frequency weights: 

> #' @param use_weights Logical, default is `TRUE`. If the estimation was weighted, 
> #' whether the aggregation should take into account the weights. Basically if the 
> #' weights reflected frequency it should be `TRUE`.

So my understanding would be - `use_weights = FALSE` -> analytical weights, `use_weights = TRUE` -> frequency weights?

pyfixest/did/saturated_twfe.py
s3alfisc: Just did a spot check and agree that it is likely an error in the CI / test setup. 

When running 

```python
import pyfixest as pf
import numpy as np
import pandas as pd

rng = np.random.default_rng(1243)
df_het = pd.read_csv("pyfixest/did/data/df_het.csv")
df_het["weights"] = rng.uniform(0, 10, size=len(df_het))
df_het["X"] = rng.normal(size=len(df_het))
print(df_het.head())


saturated_py = pf.event_study(
    data=df_het,
    yname="dep_var",
    idname="unit",
    tname="year",
    gname="g",
    estimator="saturated",
    cluster="unit",
    weights="weights",
)

saturated_py.aggregate()
```
which is more or less what happens in the added tests, the lines flagged as non-covered are actually reached.

s3alfisc: Yes, sorry? My mistake - this test should run under the test-against-fixest ci run.

s3alfisc: Fairly certain now that it is an upload error to codecov. I've tried to change the CI, let's see if that solves it =)

s3alfisc: Looks like it's fixed! 🎉

s3alfisc:
pre-commit.ci autofix

s3alfisc:
Now I think every thing looks good - the only thing I'd like to clarify before merging: 

- Does `use_weights = True` trigger fweights or aweights? If fweights, we need to update the docs to make this behavior explicit
- If `fweights`, should we maybe a) deviate from fixest syntax and call the argument "weights_type" instead of "use_weights", and b) default to analytical weights? As aweights are default in feols(), this is what I think users would expect as behavior?

shapiromh:
Thanks, Alex. I checked to verify that the tests would fail if we were using "fweights" instead of "aweights". I had expected this since you had mentioned that fixest isn't explicit about the types of weights it uses, but it should be aweights.

As far as I can see, the sunab code only additionally weights in the way I already implemented in the code. I agree both the text and the usage make it fairly clear that the weights here are used as "fweights", though I don't think this is ever seen by the underlying estimation routine that is fitting the model. 

My best current guess is that the weights argument is used in both ways in fixest. It is used as aweights when running the core estimation routine. Then for aggregating from the cohort-period estimates to whatever the user specifies, it is practically treated as a frequency weight. 

Finally, I realized that the "aggregate" function is not really user facing in fixest. It should just be called behind the scenes when the user specifies the "agg" argument when summarizing the result from the estimation (https://lrberge.github.io/fixest/reference/sunab.html). In the code base I didn't find a case where use_weights = False so I am guessing they always use these weights if supplied when setting up in the initial model. 

If my understanding of all this is correct, then I don't think there is a way to match fixest's output and be consistent in pyfixest's use of the weights. If you want to match the output, then I guess the right thing is to remove the "use_weights" option and force the aggregation to use the weights as sampling weights. If you don't want to match the fixest output, necessarily, the whole PR might need a re-think.

s3alfisc:
Sorry for the delayed response, my parents were visiting over the weekend =) 

> My best current guess is that the weights argument is used in both ways in fixest. It is used as aweights when running the core estimation routine. Then for aggregating from the cohort-period estimates to whatever the user specifies, it is practically treated as a frequency weight.

This is also my understanding, and I wonder if this does not lead to errors because frequency weights and analytical weights SEs are different? I.e. when computing the vcov matrix, fixest uses "analytical" errors? There's also a slight chance that analytical and frequency errors have the exact same form and I still misunderstand something. 

Mabye the best next step here would be to open a PR in the `fixest` repo and ask Laurent directly?

s3alfisc:
Also linking to the Stata documentation that might be helpful / I need to take a closer look at later: [link](https://www.stata.com/manuals13/u20.pdf)

shapiromh:
Thanks, I'll also try to read through and get more clarity before pushing an update. I'll should have time again this weekend.

shapiromh:
Sorry for disappearing on this one, Alex. Let me take a look this weekend.

s3alfisc:
> Sorry for disappearing on this one, Alex. Let me take a look this weekend.

Hi, no worries at all! Thanks for the update!

Issue #927, opened 2025-06-05
Potential Performance Optimizations for demean.rs


s3alfisc:
I was wondering if there is still some room to improve performance of the demean.rs function and created a flamegraph via the flamegraph package. 

![Image](https://github.com/user-attachments/assets/aaccbf4b-bde3-4ddc-a1e2-859d0056dbf1)

Not yet 100% sure how to read it but thought I'd ping you all nevertheless @schroedk @k-behr @mre

s3alfisc:
Code for the flamegraph is here: https://github.com/py-econometrics/pyfixest/blob/c9ed1906928ee9dba15d9f80c621c2aa3debbcaf/src/bin/profile.rs#L139

s3alfisc:
Feedback from @mre - we should take a look at the `zip()` calls and check if we can optimize ranges or even do some caching. I.e. in the code snippet below, where `sample_weights` and `group_ids` are fixed. 

```rust
        x.iter()
            .zip(sample_weights)
            .zip(group_ids)
            .for_each(|((&xi, &wi), &gid)| {
                group_weighted_sums[gid] += wi * xi;
            });
```

PR #928, opened 2025-06-05
Optimise demean_rs

kinto-b:
Partially addresses #927.

I had a play around with this this evening. The code was already pretty optimised, but I thought I could potentially squeeze a little more juice out of it, so I rearranged a little bit. Not sure the juice was worth the squeeze in this instance? I got about 5% out if we just look at the point estimates

image

Note: I didn't run the full test suite. I just ran the benchmarks.


s3alfisc:
Sweet! Taking a look now (though I don't think I can do a proper review, getting too late for that =D )

PR #930, opened 2025-06-07
Global Options

### LINKED ISSUE #766: Global Options to Set Estimation Parameters
`fixest` allows to set multiple function parameters via global options. See here for details: [link](https://lrberge.github.io/fixest/reference/setFixest_estimation.html)

It would be great to support something similar in pyfixest to allow users to set defaults globally. 

One example where this is useful is this issue by @SwannChelly https://github.com/py-econometrics/pyfixest/issues/765.

s3alfisc:
Allows to set global options for estimation. Would close #766. 

Example: 

```python
%load_ext autoreload
%autoreload 2

import pyfixest as pf 
data = pf.get_data()

pf.set_option(
    data = pf.get_data().dropna(), 
    vcov = {"CRV1": "f1"}, 
    weights = "weights", 
    demeaner_backend = "rust"
)

# if arg not explicity provided, looks for and uses global default
fit1 = pf.feols(fml = "Y ~ X1 + X2 | f2", vcov = "hetero")
fit2 = pf.feols(fml = "Y ~ X1 | f1 + f2")
fit3 = pf.feols(fml = "Y ~ X1 + X2 ")
pf.etable([fit1, fit2, fit3],)
```
Is this a feature we'd like to have? Are there any risks? @apoorvalal @juanitorduz

### DIFF
--- a/docs/quickstart.qmd
+++ b/docs/quickstart.qmd
@@ -577,3 +577,41 @@ pf.etable([fit_twfe, fit_did2s])
 ```
 
 For more details see the vignette on [Difference-in-Differences Estimation](https://py-econometrics.github.io/pyfixest/difference-in-differences.html).
+
+## Setting Global Options
+
+We can set all function arguments of the estimation functions `feols`, `fepois`, and `feglm` as global options.
+
+For example, we can set the estimation data set, variance covariance matrix, weights, small sample corrections and demeaner backend defaults via `pf.set_options()`.
+
+```{python}
+pf.set_options(
+    data = pf.get_data().dropna(),
+    vcov = {"CRV1": "f1"},
+    weights = "weights",
+    demeaner_backend = "rust"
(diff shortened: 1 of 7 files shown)


apoorvalal:
I can see the appeal of something like this for interactive use, but also it introduces a lot of potentially consequential hidden state that makes reproducibility a pain and should generally be discouraged. 
What if we did the intermediate thing where we let one set a bunch of preferences in a dict and pass the entire dict to each model fit call and unpack it inside feols with **kwargs ? That's roughly how scikit learners with many hyperparameters work right?

juanitorduz:
I agree with @apoorvalal ! Having these hidden variables can be very painful to debug. the intermediate suggestion seems reasonable :)

s3alfisc:
Closing as assessed to be a poor idea =)

PR #941, opened 2025-06-15
Implements "preprocessing frisch newton" algorithm

s3alfisc:
Implements the "preprocessing frisch newton" algorithm as in Portnoy & Koenker (see [Chernozukov et al Algorithm 1 ](https://arxiv.org/pdf/1909.05782)for reference). 

Questions: 

- should "pfn" also be used when computing standard errors? currently always defaults to "fn"
- should we block CRV inference with the pfn algo? If yes, should we use the same set of starting observations? Likely yes? 
- As the method assumes independent observations (due to the independent sampling), should we block CRV inference? Could be generalized to using a block bootstrap in the preprocessing step? fyi @apoorvalal 

Usage: 

```{python}
import pyfixest as pf 
import pandas as pd 
import numpy as np 

N = 1_000_000
data = pd.DataFrame({
    "Y": np.random.randn(N),
    "X1": np.random.randn(N),
    "X2": np.random.randn(N),
    "X3": np.random.randn(N),
})

fml = "Y ~ X1 + X2 + X3"

# ~ 18s
pf.quantreg(fml, data, method = "fn")

# ~ 12s
pf.quantreg(fml, data, method = "pfn")
```


s3alfisc:
Btw, applying the preprocessing also in the CRV estimation, we can cut down run time to 1.5 seconds: 

```python
# 18 s
fit_fn = pf.quantreg(fml, data, method = "fn")
# 1.5 s
fit_pfn = pf.quantreg(fml, data, method = "pfn")
```

apoorvalal:
AFK for most of today but re: quantregmulti - might be worth looking into [VQR](https://arxiv.org/abs/2205.14977) and its repo https://github.com/vistalab-technion/vqr

s3alfisc:
Some updates: as a default, I will implement the Powell Estimator (then uses a non-parametric KDE to estimate sparsity) for computing iid, hetero, CRV errors. This allows use one class of estimators systematically throughout (the CRV estimator is a Powell style estimator). "nid" as in the R package will also be supported (based on linear interpolation, significantly slower than Powell). I will not port the R-quantreg defaults, iid errors in pf and R-quantreg will not match. Main motivation is consistency + I don't know how to implement the R-quantreg SEs without breaching the GPL license as the actual computations are not explained anyhwhere (except for nid, which is explained in the Koenker QR book). 

I've also added some benchmarks on the QR process (the Chernozhukov methods 2 is very nice and fast); the performance of the pfn algo in the quantile regression process still seems to be lacking.

s3alfisc:
I hope it's done now (minus some smaller cleanups) ...

s3alfisc:
pre-commit.ci autofix

Issue #933, opened 2025-06-11
Allow users to set the maximum numbers of iterations in the demeaner algo from the function APIs


s3alfisc:
Currently, it is not possible for users to set the maximum number of iterations via the functional APIs `feols` and `fepois`. 

We would like to introduce a function argument, `fixef_maxiter`, that is being passed from `feols` / `fepois` to the demeaner function. 

We'd like to update tests in [test_demean](https://github.com/py-econometrics/pyfixest/blob/master/tests/test_demean.py).

WiktorTheScriptor:
is the fixef_maxiter argument supposed to be passed into feols/fepois or is it being passed into demean?

WiktorTheScriptor:
is it to change the maxiter=100_000 values to something custom? i see demean already has a maxiter argument, or is this to do with demean_model?

s3alfisc:
Hi @WiktorTheScriptor , yes, you are correct, thanks for taking a look! We had already assigned this one inofficially to @damandhaliwal but I unfortunately forgot to assign him. Sorry in case you already did a bit of work on this one!

WiktorTheScriptor:
no problem, it was a bit too complicated for me 😄 I will have a look how damanhaliwal does it

s3alfisc:
Yes it turned out quite a lot of different files had to be touched ... much more than I had anticipated actually!

s3alfisc:
Closing it as completed.

PR #942, opened 2025-06-16
Issue933 - added the ability to customize the maximum iterations for the demeaning algorithm

damandhaliwal:


damandhaliwal:
Fixes #933

s3alfisc:
Looks already quite good. As written on discord, more in depth review will follow later tonight =) 

Could you delete all the non python files? And maybe add the git ignore so that these won't be committed going forward?

s3alfisc:
Hi @damandhaliwal , this looks good! 

The CI fails because [quantreg](https://github.com/py-econometrics/pyfixest/blob/837f06f63088760b4a2fbe651d160321eef7320c/pyfixest/estimation/estimation.py#L1124) does not get the `fixef_maxiter` argument for `_estimation_input_checks`, you can just hardcode it to `fixef_maxiter = 100_000`, similar to `iwls_maxiter`. Generally I think this is a design failure on my end - `_estimation_input_checks` should only check "common" inputs that all functional APIs share; while separate functions should check "specific" inputs, as [_quantreg_input_checks](https://github.com/py-econometrics/pyfixest/blob/837f06f63088760b4a2fbe651d160321eef7320c/pyfixest/estimation/estimation.py#L1132) does for quantreg (maybe a follow up PR? =D ). 

Could you please delete the `.idea` folder and its content? Also delete it from the git history if possible. Something like `git filter-repo --path secrets/ --invert-paths --refs HEAD` might work (LLM told me that this works, better to check yourself 😄 ). 

Also some minor comments and suggestions in the code. 

Thank you!

damandhaliwal:
Thanks for the review Alex! I'll make those suggested changes and get rid of .idea. I might have checked out my branch before you merged ```quantreg``` so I'll update for those issues. 

That's a good idea to check for specific inputs for each function. After we close, this I'll file a new issue and then we can resolve that too over time.

damandhaliwal:
Ignore that ```git filter-repo --path secrets/ --invert-paths --refs HEAD``` made too many changes. I'll close this PR and file a new one with a clean branch. Apologies for the mess.

PR #969, opened 2025-07-15
Doc improvements

apoorvalal:
Docs currently only build on Alex's windows machine (the quarto file points to `python.exe`) which is not a great state of affairs. I'm going to work on a setup so we have documentation deployment and preview in each branch. 

[likely ugly, WIP]

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -9,17 +9,6 @@ on:
       - master
 
 jobs:
-  lint-docs:
-    name: "Lint Docs"
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout source
-        uses: actions/checkout@v4
-      - name: Install markdownlint-cli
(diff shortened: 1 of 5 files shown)


apoorvalal:
Drops the comparing with `fixest` notebook since it is a massive pain in the neck to make `rpy2` work on a prod system with github actions. IMO we can link to a notebook that demonstrates equivalent results run on someone's local machine. 

Also possibly time to update those comparisons to [ryp](https://github.com/Wainberg/ryp)

s3alfisc:
I actually think all the docs are building on the CI, it's just that I think the folder is deleted straight away? Else I couldn't explain why the docs CI step usually takes 20 minutes or so? It's usually painfully slow ...

apoorvalal:
yeah it takes well over an hour on my local machine; i think the github action is doing some fiddly caching. I'll dig into the logs on a flight tomorrow.

PR #971, opened 2025-07-16
Add missing underscore to `show_se_type` argument in docs

jsr-p:
It is correctly written in the code [example](https://py-econometrics.github.io/pyfixest/table-layout.html#hide-fixed-effects-or-se-type-rows), though. I only read the text (with the typo) and was confused why it didn't hide the the se type row :smile:


s3alfisc:
@all-contributors please add @jsr-p for docs

s3alfisc:
Ah great catch, thanks for fixing this Jonas!

PR #988, opened 2025-08-13
add waterfall

s3alfisc:
Add a waterfall chart to the Gelbach Decomposition method.


s3alfisc:
pre-commit.ci autofix

Issue #983, opened 2025-08-04
Showing R2


SwannChelly:
Hi, 

In some reporting packages (as Stargazer) there is the possibility not to show R2. I could be great to have the same functionality in pf.etable I think. 

Thank you for all your work !

Best, 
Swann

s3alfisc:
Yes, this should also be quite doable!

SwannChelly:
> Yes, this should also be quite doable!

Thank you !

dsliwka:
Currently the display of specific model statistics like R2 is "hard wired" in etable. Last year when I contributed to the etable code and thought about making the display of model statistics more flexible I had the following thought on this. But @s3alfisc I am not sure whether the following makes sense: 
Suppose all estimation classes instead of (or on top of) having the model statistics as attributes itself (like _r2) have a _modelstats attribute which is a dictionary with the name of the model statistics as keys and the respective statistics as values. 
This would allow a flexible implementation of the display of model statistics in etable and other output methods. In etable the user could, for instance, just specify a list of the names of the model statistics to be displayed in the table (with a default). etable would check whether the respective statistics is in _modelstats for the respective column and display it. It would then also be able to display model statistics of new estimators implemented in pyfixest in the future or newly added statistics without the need to adapt the code.

dsliwka:
Sorry, ignore the previous response. It should be much easier without changing the estimation classes to implement the same functionality using getattr() . If I find the time in the next few weeks I can try to implement this.

leostimpfle:
This is now supported in `maketables`:

https://py-econometrics.github.io/maketables/docs/ETable.html#specify-displayed-model-statistics

PR #992, opened 2025-08-18
Allow user to specify the model statistics to be displayed as well as their order

dsliwka:
Replaces the previously hard‑wired bottom panel model statistics logic in etable with more flexible system.

Key changes:
- New arguments: model_stats (list of attribute names without leading _) and model_stats_labels (optional label overrides).
- Any model attribute can be displayed by listing its name without "_" (e.g. _r2_within -> "r2_within", _N -> "N"). Should therefore also work with future estimation methods and attributes.
- Default (when model_stats=None) reproduces legacy behavior: ["N", "se_type?" if show_se_type, "r2", ("r2_within" if available else "adj_r2")].
- Passing an empty list model_stats=[] suppresses all built‑in stats.

Addresses #983.


dsliwka:
pre-commit.ci autofix

s3alfisc:
Thank you Dirk, I'll try to take a look tomorrow (latest by Wednesday). Sorry for not replying earlier to your comments!

leostimpfle:
I think this can be closed because it is implemented in `maketables`:

https://py-econometrics.github.io/maketables/docs/ETable.html#specify-displayed-model-statistics

PR #995, opened 2025-08-21
Feature/demean accelerated

schroedk:
This PR implements a simple version of the Iron-Tucks acceleration.

### DIFF
--- a/pyfixest/core/_core_impl.pyi
+++ b/pyfixest/core/_core_impl.pyi
@@ -20,3 +20,10 @@ def _count_fixef_fully_nested_all_rs(
     cluster_data: NDArray[np.uint64],
     fe_data: NDArray[np.uint64],
 ) -> tuple[np.ndarray, int]: ...
+def _demean_accelerated_rs(
+    x: NDArray[np.float64],
+    flist: NDArray[np.uint64],
+    weights: NDArray[np.float64],
+    tol: float = 1e-08,
+    maxiter: int = 100_000,
+) -> tuple[np.ndarray, bool]: ...
--- a/pyfixest/core/demean_accelerated.py
+++ b/pyfixest/core/demean_accelerated.py
@@ -1,0 +1,73 @@
+import numpy as np
+from numpy.typing import NDArray
+
+from ._core_impl import _demean_accelerated_rs
+
+
+def demean_accelerated(
+    x: NDArray[np.float64],
+    flist: NDArray[np.uint64],
+    weights: NDArray[np.float64],
+    tol: float = 1e-08,
+    maxiter: int = 100_000,
+) -> tuple[NDArray, bool]:
+    """
+    Demean an array.
+
+    Workhorse for demeaning an input array `x` based on the specified fixed
+    effects and weights via the alternating projections algorithm.
+
+    Parameters
+    ----------
+    x : numpy.ndarray
+        Input array of shape (n_samples, n_features). Needs to be of type float.
+    flist : numpy.ndarray
+        Array of shape (n_samples, n_factors) specifying the fixed effects.
+        Needs to already be converted to integers.
+    weights : numpy.ndarray
+        Array of shape (n_samples,) specifying the weights.
+    tol : float, optional
+        Tolerance criterion for convergence. Defaults to 1e-08.
+    maxiter : int, optional
+        Maximum number of iterations. Defaults to 100_000.
+
+    Returns
+    -------
+    tuple[numpy.ndarray, bool]
+        A tuple containing the demeaned array of shape (n_samples, n_features)
+        and a boolean indicating whether the algorithm converged successfully.
+
+    Examples
+    --------
+    ```{python}
+    import numpy as np
+    import pyfixest as pf
+    from pyfixest.utils.dgps import get_blw
+    from pyfixest.estimation.demean_ import demean
+    from formulaic import model_matrix
+
+    fml = "y ~ treat | state + year"
+
+    data = get_blw()
+    data.head()
+
+    Y, rhs = model_matrix(fml, data)
+    X = rhs[0].drop(columns="Intercept")
+    fe = rhs[1].drop(columns="Intercept")
+    YX = np.concatenate([Y, X], axis=1)
+
+    # to numpy
+    Y = Y.to_numpy()
+    X = X.to_numpy()
+    YX = np.concatenate([Y, X], axis=1)
+    fe = fe.to_numpy().astype(int)  # demean requires fixed effects as ints!
+
+    YX_demeaned, success = demean(YX, fe, weights = np.ones(YX.shape[0]))
+    Y_demeaned = YX_demeaned[:, 0]
+    X_demeaned = YX_demeaned[:, 1:]
+
+    print(np.linalg.lstsq(X_demeaned, Y_demeaned, rcond=None)[0])
+    print(pf.feols(fml, data).coef())
+    ```
+    """
+    return _demean_accelerated_rs(x, flist.astype(np.uint64), weights, tol, maxiter)
--- a/src/demean_accelerated.rs
+++ b/src/demean_accelerated.rs
@@ -1,0 +1,199 @@
+use ndarray::{Array2, ArrayView1, ArrayView2, Zip};
+use numpy::{PyArray2, PyReadonlyArray1, PyReadonlyArray2};
+use pyo3::prelude::*;
+use rayon::prelude::*;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+
+mod internal {
+    pub(super) fn sad_converged(a: &[f64], b: &[f64], tol: f64) -> bool {
+        a.iter().zip(b).all(|(&x, &y)| (x - y).abs() < tol)
+    }
+
+    pub(super) fn calc_group_weights(
+        sample_weights: &[f64],
+        group_ids: &[usize],
+        n_samples: usize,
+        n_factors: usize,
+        n_groups: usize,
+    ) -> Vec<f64> {
+        let mut group_weights = vec![0.0; n_factors * n_groups];
+        for i in 0..n_samples {
+            let weight = sample_weights[i];
+            for j in 0..n_factors {
+                let id = group_ids[i * n_factors + j];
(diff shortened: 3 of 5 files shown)


s3alfisc:
@schroedk just rebased / merged changes from master in here (new features plus I moved to using a pixi toml, plus the dev env no longer installs on windows due to compatibility challenges). 

As before, you can build the Rust bindings by typing
```
pixi r -e dev maturin-develop
```
and run the benchmarks via 
```
pixi run -e dev pytest  tests/test_demean.py::test_demean_complex_fixed_effects
```

s3alfisc:
Benchmarks of the accelerated vs regular rust vs fixest and FixedEffectsModels.jl. Looks like good progress to me! 

image

schroedk:
> Benchmarks of the accelerated vs regular rust vs fixest and FixedEffectsModels.jl. Looks like good progress to me!
> 
> image

@s3alfisc do you have the code to run this benchmark?

s3alfisc:
Yes, it's here: https://github.com/s3alfisc/fixest_benchmarks

Hope I documented the setup well, but I think clone + just + task runner should get you started. 

It's the OLS benchmarks for the hard problem that are relevant. 

Note: @grantmcdermott mentioned the other day there might be a minor issue with the benchmarks (though I don't know what exactly) so best to take it with a grain of salt (though I couldn't spot it, everything looked ok to me).

s3alfisc:
Wait it looks like I didn't push my local changes including the just setup. One sec

s3alfisc:
It's in the justfile branch on the remote 😅

https://github.com/s3alfisc/fixest_benchmarks

Requirements: global R and Julia installations. Just. 

Then type 

just setup 

To install all package deps as well as python (in local env). 

Then just bench-ols for benchmarks.

s3alfisc:
Is this confirmed? 👀 😄 🚀 

image

s3alfisc:
Re failing tests atm - seems to be the prediction method, which is notoriously fickle and should be fixable by changing the tolerance or just checking on a different subset of the prediction array: 

```python
 x1 = array([23.03009222,  7.43960451, 14.04928582, 17.83110832, 17.46495242])
x2 = array([23.03009052,  7.43960323, 14.04928428, 17.83110934, 17.4649537 ])
```

schroedk:
@s3alfisc Looks like it is on par now on the benchmark
plot_ols

s3alfisc:
> @s3alfisc Looks like it is on par now on the benchmark

So cool! Super awesome!!!

s3alfisc:
Let me know once you think it is time for a review =) (I will start anyways as it will talk me some time to study the code 😄 )

schroedk:
@s3alfisc the PR is ready for review. My suggestions/questions would be:

1. Remove the added benchmark code (no diff to .gitignore, benchmarks)
2. I added a rust version of detect_singletons and wired it into feols for speed -> what to do with the numba implementation?
3. Rust backend now resolves to the accelerated version -> what to do with the original demean.rs?
4. Too many commits, after review I would squash merge

s3alfisc:
Awesome, I will start my review then! I will try to be fast, but also have a [big PR](https://github.com/py-econometrics/pyfixest/pull/1118) from @leostimpfle to review thoroughly + all the Rust code will be a bit of a challenge for me 😄 

On your questions, I'd say: 

1. Yes, let's delete the benchmark code? I will push some changes to the [fixest_benchmark](https://github.com/s3alfisc/fixest_benchmarks) repo that I have been working on going forward, we can use it to benchmark the routines?
2. This is super awesome! For the numba implementation - I guess I would keep it for now? I haven't fully decided what to do with the numba code once we make the switch and make the rust backend the default - option 1 is to allow users to choose a numba backend, option 2 would be to delete all numba code. Do you happen have an opinion on this?
3. I would probably delete it? Or is there any use in keeping it for testing / as a reference? 
4. Yes, squash commit makes sense to me 👍

pyfixest/core/_core_impl.pyi
```diff
@@ -20,3 +20,11 @@ def _count_fixef_fully_nested_all_rs(
     cluster_data: NDArray[np.uint64],
     fe_data: NDArray[np.uint64],
 ) -> tuple[np.ndarray, int]: ...
+def _demean_accelerated_rs(
+    x: NDArray[np.float64],
+    flist: NDArray[np.uint64],
+    weights: NDArray[np.float64],
+    tol: float = 1e-08,
```
s3alfisc:
```suggestion
    tol: float = 1e-06,
```

pyfixest/core/demean_accelerated.py
```diff
@@ -0,0 +1,79 @@
...
+from numpy.typing import NDArray
+
+from ._core_impl import _demean_accelerated_rs
+
+
+def demean_accelerated(
+    x: NDArray[np.float64],
+    flist: NDArray[np.uint64],
+    weights: NDArray[np.float64],
+    tol: float = 1e-08,
```
s3alfisc:
```suggestion
    tol: float = 1e-06,
```

schroedk:
> Awesome, I will start my review then! I will try to be fast, but also have a [big PR](https://github.com/py-econometrics/pyfixest/pull/1118) from @leostimpfle to review thoroughly + all the Rust code will be a bit of a challenge for me 😄
> 
> On your questions, I'd say:
> 
> 1. Yes, let's delete the benchmark code? I will push some changes to the [fixest_benchmark](https://github.com/s3alfisc/fixest_benchmarks) repo that I have been working on going forward, we can use it to benchmark the routines?
> 2. This is super awesome! For the numba implementation - I guess I would keep it for now? I haven't fully decided what to do with the numba code once we make the switch and make the rust backend the default - option 1 is to allow users to choose a numba backend, option 2 would be to delete all numba code. Do you happen have an opinion on this?
> 3. I would probably delete it? Or is there any use in keeping it for testing / as a reference?
> 4. Yes, squash commit makes sense to me 👍

1. deleted
2. I will keep it for now, feols already uses the rust version. There is no other usage (beside tests) of the numba code, so one can rid of it any time and re-export the rust backed version for backwards compatibility. I cannot think of a reason to keep the numba backend in the middle/long run, so I would go with Option 2.
3. I deleted it and rewired the python bindings to the accelerated version (commit 
6c091d7) -> no breaking change in the python API
4. I created a smaller more meaningful set of commits, so you could also step through the commits for the review. If you like, we can squash to one commit for merge.

s3alfisc:
If I understand it correctly, the algorithm is set up to operate in "coef space" - this means that we have estimated fixed effects at the end of the demeaning routine. Would it be possible to return these fixed effects? In these fixed effects models, we are usually not interested in them, but we might need them to conduct predictions, which in turn is needed to compute marginal effects quickly. This is more or less how the R `margins` package does it, and Stata and R and Pythons `marginal effects` too. From the [margins paper](https://cran.r-project.org/web/packages/margins/vignettes/TechnicalDetails.pdf): 
image
image

In pyfixest, we currently get the estimated fixed effects post hoc by solving a sparse system via LSMR (see [here](https://github.com/py-econometrics/pyfixest/blob/8d07f0b36b969194ab52ef737833ff1d039f71c3/pyfixest/estimation/feols_.py#L2005)), which I think we could skip now?

schroedk:
@s3alfisc latest benchmark
With numba:
plot_ols

Without numba:
plot_ols

s3alfisc:
This looks awesome! Should we maybe update the benchmarks in the readme?

s3alfisc:
Just a thought - is it currently possible for users to turn of the "reorder fixed effects" solution via the API? 
I think it could be useful to allow this for cases where users know which fixed effects are most complicated, which might not necessarily be identical to the "largest" fixed effects. 

Example: 

We have three fixed effects: nba_player, district, city. NBA player is just an identifier of a player, while district and city tell us about where a game is played. Because of schedules, players play games in different cities and districts and hence there is little correlation. But of course there are much more players than cities / districts. 

As cities are nested within districts, the fixed effects will be super correlated. In this case, it might make sense to work on district & city in the two way part of the three part strategy for 3+ FE even if the cardinality of the FE is lower?

schroedk:
> Just a thought - is it currently possible for users to turn of the "reorder fixed effects" solution via the API? I think it could be useful to allow this for cases where users know which fixed effects are most complicated, which might not necessarily be identical to the "largest" fixed effects.
> 
> Example:
> 
> We have three fixed effects: nba_player, district, city. NBA player is just an identifier of a player, while district and city tell us about where a game is played. Because of schedules, players play games in different cities and districts and hence there is little correlation. But of course there are much more players than cities / districts.
> 
> As cities are nested within districts, the fixed effects will be super correlated. In this case, it might make sense to work on district & city in the two way part of the three part strategy for 3+ FE even if the cardinality of the FE is lower?

It is not, but easy to change on the rust side. Actually it is just reverting [35ba573](https://github.com/py-econometrics/pyfixest/pull/995/commits/35ba573fd2c6e4f262ddd89ec4cd3465e2409b3b), as I had it before but decided to keep it more simple. But I did not consider your very valid point. The bigger problem is, how to add it to the python part?

s3alfisc:
> The bigger problem is, how to add it to the python part?

Fixest has a Boolean argument "reorder_coef" - I think we could just add it to the python interface and then pass it to the Rust code? For the non-accelerated algos, it would just not do anything. Maybe I am missing something?

schroedk:
> > The bigger problem is, how to add it to the python part?
> 
> Fixest has a Boolean argument "reorder_coef" - I think we could just add it to the python interface and then pass it to the Rust code? For the non-accelerated algos, it would just not do anything. Maybe I am missing something?

@s3alfisc that would be one possibility, but then you have to also expose it to calling functions like feols. In this case the parameter is confusing, because it is only relevant for the rust backend.

What about this, I make it configurable on the rust side and we handle the python side in a separate issue? In this case, here we only have to talk about the default:

- true: sorts every time, the average user would profit, but "power user" sorting the FE would not be able to
- false: on average, the convergence might be slower due to the missing sorting (no idea how much), we would have to document, that manual sorting the hard FE to the front can increase convergence speed. But again, this only holds true for the rust backend.

Edit: I would vote for false (not sorting by default), it is more explicit (has a higher value for me than pure maximal performance)

Issue #675, opened 2024-10-26
HAC Standard Errors


s3alfisc:
See equations (9) and (10) in [Various Versatile Variances](file:///C:/Users/alexa/Downloads/v95i01-15.pdf). 

Implement utility functions `_vcov_hc()` and `_vcov_hac()` following the R `sandwich` package. 

Allow arguments "hac" for `pf.feols()` and `pf.fepois()`.

Likely, everything needs to be implemented in `numba` for performance reasons; the estimators are hard to vectorize.

s3alfisc:
Hi @damandhaliwal, if you are looking for a challenge, I think this would be an interesting issue. It covers some econometrics, API design, and optimization (via numba). Would be a good issue to get started on via a pair programming session!

damandhaliwal:
Hi Alex, sign me up! I can get started on this later in the week and get a good session in over the weekend. I'll read the notes you attached and the R ```sandwich``` package documentation.

s3alfisc:
Awesome! I also have some code lying around in a branch, will share it with you

s3alfisc:
So, for the API design I made the following comments - I thought it would be a good idea to add a new argument to the estimation functions, `vcov_kwargs`, that would allow users to pass information required for the panel HAC estimation: 

```python
    vcov : Union[VcovTypeOptions, dict[str, str]]
        Type of variance-covariance matrix for inference. Options include "iid",
        "hetero", "HC1", "HC2", "HC3", "NW" for Newey-West HAC standard errors,
        "DK" for Driscoll-Kraay HAC standard errors, or a dictionary for CRV1/CRV3 inference.
        Note that NW and DK require to pass additional keyword arguments via the `vcov_kwargs` argument.

    vcov_kwargs: dict[str, Any]
        Additional keyword arguments to pass to the vcov function. These keywoards include
        "lags" for the number of lags to use in the Newey-West (NW) and Driscoll-Kraay (DK) HAC standard errors.
        "time_id" for the time ID used for NW and DK standard errors, and "panel_id" for the panel
        identifier used for NW and DK standard errors.
```

Then, we'd have to add `def _nw_meat` and `_dk_meat` to `vcov_utils.py`. These would implement the meat matrix for the Newey West and Driskool-Kray vcov's. Here we can lazily use for loops and after we're done, JIT compile them via `numba`.

Lag lengths can either be specified by the user, or we default to the standard heuristic.

[...]

PR #997, opened 2025-08-24
Issue675 - adding HAC Standard Errors

damandhaliwal:
Draft PR for my implementation of the HAC Standard errors. This is not yet completely functional and I'm working on it, but would love another set of eyes on it @s3alfisc 

Addresses #675


pyfixest/estimation/.DS_Store
s3alfisc, line 1: Can you add the file to the git ignore?

damandhaliwal: I didn't push any changes to it but I'll get rid of it in the next commit

PR #1000, opened 2025-08-27
HAC Standard Errors

damandhaliwal:
New clean PR with updated for changes to the master. Addresses #675. @s3alfisc let me know what you think. Hoping to wrap this up this weekend. 

If you could review the `_nw_meat` function in `vcov_utils.py` that would be great. Just want to confirm my understanding and the implementation.

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -961,7 +961,7 @@ def test_errors_vcov_kwargs():
 
     # Error 11: panel_id value is not a string (list)
     with pytest.raises(
-        ValueError, match=r'The function argument `vcov_kwargs` must be a.'
+        ValueError, match=r"The function argument `vcov_kwargs` must be a."
     ):
         pf.feols(
             "Y ~ X1", data=data, vcov="NW", vcov_kwargs={"panel_id": ["col1", "col2"]}
@@ -995,9 +995,7 @@ def test_errors_hac():
     )
 
     # Error 3: time_id is not provided if vcov is NW or DK
-    with pytest.raises(
-        ValueError, match=r"Missing required 'time_id' for NW/DK vcov"
-    ):
+    with pytest.raises(ValueError, match=r"Missing required 'time_id' for NW/DK vcov"):
         pf.feols(
             "Y ~ X1",
             data=data,


pyfixest/estimation/vcov_utils.py
```diff
@@ -96,6 +96,88 @@ def _get_vcov_type(
 
     return vcov_type  # type: ignore
 
+@nb.njit(parallel=False)
```
s3alfisc, line 88: For the function API, it would be nice to follow this: 

```python
def _nw_meat_panel(
    X: np.ndarray, u_hat: np.ndarray, time_id: np.ndarray, L: int
):
```
Mostly, I would like to avoid to have to pass the input data frame. The time id should suffice? 

Also, please add type hints =) 

Plus, no need for extra IV logic, I think? These args are not used at all: is_iv=False, 
             tXZ=None, tZZinv=None, tZX=None):

pyfixest/estimation/vcov_utils.py
```diff
@@ -96,6 +96,88 @@ def _get_vcov_type(
...
+    if lags is None:
+        lags = int(np.floor(n_time ** (1 / 4)))
+
+    # bartlett kernel weights
+    weights = np.array([1 - j/(lags + 1) for j in range(lags + 1)])
+    weights[0] = 0.5  # Halve first weight
+
+    n, k = ordered_scores.shape
+    meat = np.zeros((k, k))
+
```
s3alfisc, line 140: See our discussion in the discord =)

pyfixest/estimation/vcov_utils.py
```diff
@@ -96,6 +96,88 @@ def _get_vcov_type(
...
+        if j == 0:
+            gamma_j = ordered_scores.T @ ordered_scores
+
+
+            meat += weights[j] * gamma_j
+        else:
+            gamma_j = ordered_scores[j:].T @ ordered_scores[:-j]
+            meat += weights[j] * (gamma_j + gamma_j.T)
+
+    meat = meat / n
```
s3alfisc: I don't think you have to divide by /n here, these should cancel out when forming the full vcov matrix.

pyfixest/estimation/vcov_utils.py
```diff
@@ -96,6 +96,88 @@ def _get_vcov_type(
...
+
+    if is_iv:
```
s3alfisc: can be dropped I believe?

s3alfisc: This should be handled in the vcov method. Here we only want to compute the meat, which is identical in OLS & IV.

s3alfisc:
Regarding tests, you can take a look at these: https://github.com/py-econometrics/pyfixest/blob/72b57a7ef7c8055094a74ac8e5ce810b818d3c23/tests/test_vs_fixest.py#L207

Basically, we call r-fixest via rpy2 and then compare results. We can write a similar test for HAC and test fewer formulas?

pyfixest/estimation/vcov_utils.py
s3alfisc, line 88: Note that we could parallelize if we do not modify in place (which might be faster than parallelization).

damandhaliwal: would it be faster to apply the IV logic in `feols_.py` than the meat function?

s3alfisc: No but I think it is cleaner this way - because the "meat" func then really only computes the meat, which can then in turn be used for OLS, IV, everything else?

s3alfisc:
To do's: 

- [ ] maybe check if the implementation can be improved in terms of performance (avoid memory allocations, in place operations if possible)
- [ ] add tests for IV
- [x] add that post-estimation update via `.vcov()` produces correct results
- [x] add tests for errors (i.e. when lags not provided, or a string) in test_errors.py
- [ ] update docs
- [ ] test that if time_id is not provided and sorted, we get the same result as when setting time id
- [ ] maybe other things, will have to think this through

Also, should we require a time_id value to be set in the vcov_kwargs for the HAC estimator? Advantage: we make very explicit what happens. Disadvantage: if users run multiple models, just sorting the data once before the feols() should increase model performance? Maybe a compromise would be to add a warning to inform users about the default behavior when they do not add a time_id arg?

damandhaliwal:
> Also, should we require a time_id value to be set in the vcov_kwargs for the HAC estimator? Advantage: we make very explicit what happens. Disadvantage: if users run multiple models, just sorting the data once before the feols() should increase model performance? Maybe a compromise would be to add a warning to inform users about the default behavior when they do not add a time_id arg?

Hmm, I can definitely see the advantage of requiring the time_id value. But would the performance be that improved? I will add to the documentation.

damandhaliwal:
@s3alfisc I'm working on the implementation for the Newey West Panel meat. Let me know what you think of this function for checking balance, it's different from what `fixest` implements:

```
def _check_balanced(panel_arr: np.ndarray, time_arr: np.ndarray) -> bool:
    unique_panels = np.unique(panel_arr)
    unique_times = np.unique(time_arr)
    expected_time_count = len(unique_times)

    for panel_id in unique_panels:
        mask = panel_arr == panel_id
        panel_times = np.unique(time_arr[mask])

        if len(panel_times) != expected_time_count:
            return False
    
    return True
```

`fixest` implementation in C++ below. I changed the function because this requires the `panel_arr` to be integers and in order while we don't

```
  // checking the balance
  bool balanced = true;
  if(unit[0] != 1){
    balanced = false;
  } else {
    int n_unit = 1;
    int t_current = time[0];

    for(int i=1 ; i<N ; ++i){
      if(t_current != time[i]){
        // we change time period
        // we check all is fine
        if(n_unit != G){
          balanced = false;
          break;
        }

        // OK
        n_unit = 0;
        t_current = time[i];
      } else if(unit[i] - unit[i - 1] != 1){
        balanced = false;
        break;
      }

      ++n_unit;
    }
  }
```

s3alfisc:
Oh wait a second, I actually have some code to share! Will push it now. One sec

s3alfisc:
Changed the following: 

- Add meat matrix for Panel HAC.
- Requires users to always provide time_id.
- Requires users to always provide lag when time series HAC.
- Corrects time series HACs default -> fixes uses bwNeweyWest and does not use the same default lag as for panel HAC, see attached screenshot
- adds unit tests

s3alfisc:
What we still have to do: 

- optimize the HAC code for both time series and panel 
- implement DK-HAC? but could do this in a separate PR
- I currently turn of the small sample adjustments in the tests to make fixest & pyfixest match for panel HAC. I think the panel HAC in fixest might take nesting into account as well, which we handle for CRV errors [here](https://github.com/py-econometrics/pyfixest/blob/b147cbe845b216ae766aa0276016f806ad712db9/pyfixest/estimation/feols_.py#L719) but is currently ignored for panel-HAC

pyfixest/estimation/estimation.py
```diff
@@ -1345,6 +1409,39 @@ def _estimation_input_checks(
                 "The function argument `separation_check` must be a list of strings containing 'fe' and/or 'ir'."
             )
 
+    if vcov_kwargs is not None:
```
s3alfisc, line 1423: I am wondering if we should keep all these checks here, or move them to the `vcov` method? Because users can call vcov post-estimation and provide new arguments, and we would like to check these as well?

pyfixest/utils/utils.py
```diff
@@ -393,3 +393,32 @@ def capture_context(context: Union[int, Mapping[str, Any]]) -> Mapping[str, Any]
         procedure like: `.get_model_matrix(..., context=<this object>)`.
     """
     return _capture_context(context + 2) if isinstance(context, int) else context
+
+def _check_balanced(panel_arr: np.ndarray, time_arr: np.ndarray) -> bool:
```
s3alfisc, line 426: Can you add a unit test for it? 

Btw, I think we could also have easily done this in pandas with a groupby(["panel_id", "time_id"]).size() 
and throw an error if we ever see a result that is not one? 

Plus, please use the @njit decorator + maybe we can parallelize? 

Other comment: as we are sorting in the _meat_hac_panel function already, maybe we should call this function there and then make use of the fact that we have already sorted once? This should make things more efficient right?

damandhaliwal:
> What we still have to do:
> 
> * implement DK-HAC? but could do this in a separate PR
> * I currently turn of the small sample adjustments in the tests to make fixest & pyfixest match for panel HAC. I think the panel HAC in fixest might take nesting into account as well, which we handle for CRV errors [here](https://github.com/py-econometrics/pyfixest/blob/b147cbe845b216ae766aa0276016f806ad712db9/pyfixest/estimation/feols_.py#L719) but is currently ignored for panel-HAC

I think we should just implement DK-HAC as well in this PR while we still have this fresh in our minds. This way we can resolve more issues and potentially even push cleaner and more optimized code. I'll take a look at the nesting thing in fixest and see what we're missing

pyfixest/utils/utils.py
damandhaliwal: I guess I don't need this anymore after your latest commit

s3alfisc:
Looks like fixest by default treats all time lags as consecutive: [link](https://lrberge.github.io/fixest/reference/feols.html#arg-panel-time-step).

**Edit**: looks like these have only been added with fixest 0.13, which is currently in dev. I suggest that we wait until fixest 0.13.0 is released and then add all of this logic for handling lags and assume that, until then, time series data is always consecutive?

s3alfisc:
> I think we should just implement DK-HAC as well in this PR 

Sounds good to me. 

> Do we want to implement the same weight logic as nw_meat and fixest to this with w[0]=0.5 and the loop going from 0, lag+1

Yes, though I think it is not 100% needed as we basically match fixest already?

pyfixest/estimation/vcov_utils.py
```diff
@@ -97,6 +97,195 @@ def _get_vcov_type(
...
+        The time variable for clustering. Assume that there are no duplicate time periods.
+    lag: int, optional
+        The number of lag for the HAC estimator. Defaults to floor (# of time periods)^(1/4).
+    """
+    order = np.argsort(time_arr)
+    ordered_scores = scores[order]
+
+    time_periods, k = ordered_scores.shape
+    time_scores = np.zeros((time_periods, k))
+
```
s3alfisc: I think this aggregation is wrong, no? We need to aggregate over a fixed t for each unit?

s3alfisc: Fixed with [707909a](https://github.com/py-econometrics/pyfixest/pull/1000/commits/707909a5924fdb81c475409285d6283ca0e52683)

s3alfisc:
Some optimized HAC implementations + options to handle non equal spacing. https://github.com/py-econometrics/pyfixest/tree/hac-optimise

s3alfisc:
I learned that fixest does not support "gaps" in HAC SEs & assumes always equidistant time variables: 

image

For this PR to get through, I would suggest we simplify our lives and for now focus on implementing only equidistant and time delta = 1 HAC errors. 

And once we got this to work, we can focus on more complex time patterns?

PR #1040, opened 2025-10-02
Test refactor

s3alfisc:
Big refactor of the testing suite that drops rpy2 support.


s3alfisc:
We will not do this, slow and too complex.

PR #1058, opened 2025-11-03
HAC: fix bug with indexes plus vectorization

s3alfisc:


s3alfisc:
pre-commit.ci autofix

Issue #1049, opened 2025-10-26
Decrease fixef default tolerance to 1e-06


s3alfisc:
As this is what fixest is doing: 

<img width="304" height="97" alt="Image" src="https://github.com/user-attachments/assets/83f782de-6c3c-4e57-9703-7e21437bcfb7" />

pyfixest currently defaults to `fixef_tol=1e-08`.

Might need some adjustments of test tolerances.

Dpananos:
I can handle this one

s3alfisc:
Tolerance adjusted with c0bfb3ecb77a809235be81665ba826bedd474e1d. 
Remaining: setting the max iterations.

s3alfisc:
Closing this as I updated the number of maxiters.

PR #1061, opened 2025-11-07
Decreasing default tolerance to 1e-06

pdoupe:
- adjusted fixest tolerance level from 1e-08 to 1e-06 
- adjusted docstrings

PR #1067, opened 2025-11-08
edits fixest tol -> 1e-06

Dpananos:
Addresses #1049

* Updated default parameter values for feols(), fepois(), feglm(), and feiv() functions and docstrings
* Updated test ficures to use new default tolerances in `tests/test_demean.py`

Tests passing using `pixi run -e dev pytest tests/test_demean.py -v`

~~In the process of running `pixi run -e dev tests-regular` to find if other tests break~~

`pixi run -e dev tests-regular` passing


Dpananos:
Closing since is addressed by #1061

PR #1071, opened 2025-11-10
removes aliasing from feglm

### LINKED ISSUE #857: Replace aliasing in `Feglm`
In many instances, attributes within `Feglm` are accessed via aliases. 

We would like to deprecate the aliasing notation wherever possible. 

For examples in the `Feols` class, see [this issue](https://github.com/py-econometrics/pyfixest/issues/853).

Dpananos:
Closes #857


s3alfisc (approved):
Perfect, thank you!

PR #1072, opened 2025-11-10
removes alias in fepois

### LINKED ISSUE #854: Replace aliasing in `Fepois` method
In many instances, attributes within `Fepois` are accessed via aliases. 

We would like to deprecate the aliasing notation wherever possible. 

For examples in the `Feols` class, see [this issue](https://github.com/py-econometrics/pyfixest/issues/853).

Dpananos:
Closes #854

### DIFF
--- a/pyfixest/estimation/fepois_.py
+++ b/pyfixest/estimation/fepois_.py
@@ -227,6 +227,7 @@ class Fepois(Feols):
             Demeaned dependent variable (from the last iteration of the IRLS
             algorithm).
         """
+
         def compute_deviance(Y: np.ndarray, mu: np.ndarray):
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore")


pyfixest/estimation/fepois_.py
```diff
@@ -227,75 +227,64 @@ def get_fit(self) -> None:
...
-        _N = self._N
-        _convergence = self.convergence  # False
-        _maxiter = self.maxiter
-        _tol = self.tol
-        _fixef_tol = self._fixef_tol
-        _fixef_maxiter = self._fixef_maxiter
-        _solver = self._solver
-
-        def compute_deviance(_Y: np.ndarray, mu: np.ndarray):
+        def compute_deviance(Y: np.ndarray, mu: np.ndarray):
```
Dpananos, line 231: Should this be a class method and not defined inside `get_fit`?

s3alfisc: Yes, I think this would make sense 👍 Do you want to add the change to this PR?

s3alfisc: Actually, let's just do it in a separate PR, I want to merge this one. 😄 Thanks!

PR #1074, opened 2025-11-10
makes _compute_deviance a private class method for fepois

Dpananos:
This PR makes `compute_deviance` a private method for `Fepois` as opposed to having it defined in `get_fit`.

Note:  Had some problems with mypy, so no type hint on the return type.  LMK if that should change.


s3alfisc (approved):

s3alfisc:
Looks good to me, no changes for mypy needed (it can be a pain). Love the reference!

PR #1075, opened 2025-11-11
removes aliasing

### LINKED ISSUE #853: Replace aliasing in Feols methods
In almost all methods of `Feols`, attributes are accessed via aliases. This leads to more lines of code than required, at no gain in readability. 

Example: for the `_vcov_iid` method, we currently have

```python
def _vcov_iid(self):
    N = self._N
    u_hat = self._u_hat
    bread = self._bread
    sigma2 = np.sum(u_hat.flatten() ** 2) / (N - 1)
    vcov = bread * sigma2

    return vcov
```

but we want 

```python
def _vcov_iid(self):
    sigma2 = np.sum(self._u_hat.flatten() ** 2) / (self._N - 1)
    return self._bread * sigma2
```

Dpananos:
Removes aliasing from

* `Feiv`
* `FixestMulti`
*`FeolsCompressed`

Closes  #853 since all methods listed have aliasing removed.

(Note, used Claude to clean up the remainig files)

### DIFF
--- a/pyfixest/estimation/FixestMulti_.py
+++ b/pyfixest/estimation/FixestMulti_.py
@@ -288,10 +288,11 @@ class FixestMulti:
         FixestFormulaDict = self.FixestFormulaDict
         _fixef_keys = list(FixestFormulaDict.keys())
 
-        split_values: list[Union[str, int]] = []
-        if self._run_split and self._splitvar is not None:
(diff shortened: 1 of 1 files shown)


s3alfisc (approved):
Looks good to me, thanks!

PR #1096, opened 2025-12-16
Issue991

### LINKED ISSUE #991: `i()` errors if the reference category contains some white space
Hello, being a fan of `fixest` in R, I'm very happy to see it ported to Python!

It seems that specifying the reference category with `i()` fails if the category contains some whitespace:

```python
import pyfixest as pf
import polars as pl

data = pl.DataFrame({"Y": [1, 2, 3], "cat": ["cat 1", "cat 2", "cat 2"]}).cast(
    {"cat": pl.Categorical}
)

### Works
pf.feols("Y ~ cat", data=data)

### Fails
pf.feols("Y ~ i(cat, ref='cat 2')", data=data)
```

```
ValueError: Value `'cat2'` for `TreatmentContrasts.base` is not among the provided levels.
```

It seems that whitespaces are stripped during the internal processing. In the example above, if I change the categories to `cat1` and `cat2`, then the call with `i()` works fine.

Thanks!

leostimpfle:
Closes #991

### DIFF
--- a/pyfixest/estimation/FormulaParser.py
+++ b/pyfixest/estimation/FormulaParser.py
@@ -516,7 +516,7 @@ def _deparse_fml(
     variable with `w2+w3` as its instruments.
     """
     # Split the formula string into its components
-    fml_split = re.split(r"\s*\|\s*", fml)
+    fml_split = re.split(r"\s*\|\s*", fml.strip())
     depvars, covars = re.split(r"\s*~\s*", fml_split[0])
 
     if len(fml_split) == 1:


leostimpfle:
I should clarify that this PR allows using spaces in variable names. For example, the following works

```
import pyfixest as pf
import polars as pl

data = pl.DataFrame({"Y": [1, 2, 3], "cat": ["cat 1", "cat 2", "cat 2"]}).cast(
    {"cat": pl.Categorical}
)

pf.feols("Y ~ i(cat, ref='cat 2')", data=data)
```

This is achieved by updating `_deparse_fml` to preserve spaces and a few tweaks around `FormulaParser` which essentially make formula parsing whitespace aware (by using `re.split` instead of `str.split`).

PR #1099, opened 2025-12-19
feat(sensitivity): add OVB analysis and diagnostics

damandhaliwal:
## Summary

Adds omitted-variable-bias sensitivity analysis for unweighted, non-IV OLS following Cinelli and Hazlett (2020), together with Matplotlib contour and extreme-scenario diagnostics. The implementation uses PyFixest's prepared model matrix so absorbed fixed effects, transformed terms, the estimation sample, and `store_data=False` remain aligned.

## Compatibility

- Matches R `sensemakr` for statistics and benchmark bounds across plain OLS, robust-fitted/IID-analysis, and fixed-effects cases.
- Sensitivity inference intentionally uses IID standard errors and residual degrees of freedom; robust, clustered, and HAC fits warn.
- Weighted, IV, GLM/Poisson, quantile, and `lean=True` fits fail explicitly.
- Plotting forwards the same benchmark bounds and supports manual scenarios and existing axes.

## Verification

- Focused Python tests: 37 passed.
- R `sensemakr` reference tests: 6 passed.
- Repository Python suite: 830 passed, 13 skipped, 1 xfailed.
- Extended R suite: 6 passed, 45 skipped for unrelated optional packages.
- Ruff, Ruff format, mypy, `git diff --check`, Codecov, and the full 102-page docs render passed.


damandhaliwal:
@s3alfisc the only thing I did not add in (didn't know if it was too useful/would add bloat) are the default sensitivity plots that PySensemakr makes. But I can add them if you would like

pyfixest/estimation/sensitivity.py
```diff
@@ -0,0 +1,472 @@
...
+            Benchmarks.
+        kd, ky : float or list
+            Strength multipliers.
+
+        Returns
+        -------
+        pd.DataFrame
+            DataFrame with 'r2dz_x' and 'r2yz_dx' columns.
+        """
+        from pyfixest.estimation.estimation import feols
```
s3alfisc: This path has changed in the main branch, you know have to load as `from pyfixest.estimation.api import feols`

s3alfisc:
Could you add some defensive checks (plus tests in `test_errors_warnigns`) that throw an error when users try to use the sensitivity method for models other than the OLS class? Add a `self._supports_sensitivity_analysis = True` for `Feols`, but set it to `False` in the `__init__` of the other model classes? Else all other model classes will inherit the sensitivity method from `Feols` and incorrectly run the analysis?

s3alfisc:
> @s3alfisc the only thing I did not add in (didn't know if it was too useful/would add bloat) are the default sensitivity plots that PySensemakr makes. But I can add them if you would like

Probably makes sense to add them, but we can do it in a separate PR. Not sure if these are really widely used in practice.

s3alfisc:
We should add an example on how to use the method to the quickstart notebook. 

Last - it would be great to write a "how to" notebook - what cam you do with the method, why is it needed, how does it work, maybe why the authors argue it is better than other methods (as e.g. Oster)? Maybe [something like this](https://cran.r-project.org/web/packages/sensemakr/vignettes/sensemakr.html), but even more to the point?

Note that this can also be done in a later PR.

damandhaliwal:
> Last - it would be great to write a "how to" notebook - what cam you do with the method, why is it needed, how does it work, maybe why the authors argue it is better than other methods (as e.g. Oster)? Maybe [something like this](https://cran.r-project.org/web/packages/sensemakr/vignettes/sensemakr.html), but even more to the point?

I have added everything but this. I think I will add this in a separate PR. I can draft an issue for this once we finalize and merge this PR

Issue #1097, opened 2025-12-17
Let `Fepois` inherit from `Feglm`?


leostimpfle:
`Feglm` implements the iterative weighted least squares algorithm by [Stammann (2018)](https://arxiv.org/abs/1707.01815) in [`Feglm.get_fit`](https://github.com/py-econometrics/pyfixest/blob/297307ef35c4f842914489ce9b78a4b2a2503965/pyfixest/estimation/feglm_.py#L140). While this method also works for Poisson regressions, there is currently a separate [`Fepois.get_fit`](https://github.com/py-econometrics/pyfixest/blob/297307ef35c4f842914489ce9b78a4b2a2503965/pyfixest/estimation/fepois_.py#L238).

I guess this is historically grown because we first had Poisson regressions and only recently implemented GLMs, but I had two questions:

- Are `Feglm.get_fit` and `Fepois.get_fit` in fact equivalent?
- Would it make sense to make `Fepois` inherit from `Feglm` (similarly to `Felogit` etc)?

I believe that in `fixest` Poisson is just another GLM "family" but follows the same structure and uses the same methods as other GLM families.

leostimpfle:
The discussion in #668 is relevant

s3alfisc:
In general, I believe this makes sense. But would like to discuss this more throughly before we start: 

- should we implement fixed effects support for the other GLMs first? 
- should we do an even larger refactor and create a BaseRegression class that all classes, including Feols, inherit from?

leostimpfle:
Yes to both, this is probably more a medium term problem to tackle.

Specifically, I think it makes sense to first add fixed effects to GLMs #872 (and also fix prediction for Poisson with fixed effects and `newdata` #468 which I'm working on currently).

I also think that having a `BaseRegression` class is a good idea. For example, `Fepois.predict` currently calls `Feols.predict` which itself calls some methods that do not allow passing a prediction `type`, i.e., `"response"` or `"link"`, likely because it's the same for OLS (but not for Poisson). Having `Fepois` not inherit from `Feols` would help us separate concerns more clearly.

leostimpfle:
Closing this in favour of #1148

PR #1107, opened 2025-12-21
`Fepois`: Add support for `predict` with `newdata` argument

### LINKED ISSUE #468: Fepois: Add support for `predict` with `newdata` argument

leostimpfle:
Closes #468

This PR updates `Feols.predict` to enable `Fepois.predict` with `newdata`. Note that this inelegantly introduces checks of the form `self._method == 'fepois'` in the `Feols` parent class. However, this is in some sense consistent with the current tight coupling of `Feols` and `Fepois` and helps keeping the code changes at a minimal level. In the future, we should revisit the coupling between `Feols` and `Fepois` as mentioned in #1097.

I have also enabled the relevant tests in `tests/test_predict_resid_fixef.py`. Note, however, that `Fepois._sumFE` and `Fepois.predict` with `newdata` match "only" within `atol=1e-07`.

See here:
https://github.com/py-econometrics/pyfixest/blob/66e877221041688c52e52a49fe36142c05556db3/tests/test_predict_resid_fixef.py#L126-L128

And here:
https://github.com/py-econometrics/pyfixest/blob/66e877221041688c52e52a49fe36142c05556db3/tests/test_predict_resid_fixef.py#L157-L164

### DIFF
--- a/tests/test_predict_resid_fixef.py
+++ b/tests/test_predict_resid_fixef.py
@@ -158,8 +158,6 @@ def test_vs_fixest(data, fml):
     if not np.allclose(
         fepois_mod.predict(newdata=data2, type="link")[11:16],
         stats.predict(r_fixest_pois, newdata=data2, type="link")[11:16],
-        fepois_mod.predict(newdata=data2, type="link"),
-        stats.predict(r_fixest_pois, newdata=data2, type="link"),
         atol=1e-07,
         equal_nan=True,
     ):


leostimpfle:
pre-commit.ci autofix

tests/test_predict_resid_fixef.py
```diff
@@ -154,8 +155,13 @@ def test_vs_fixest(data, fml):
         raise ValueError("Predictions for OLS are not of the same length.")
 
     # test predict for Poisson
-    # if not np.allclose(fepois_mod.predict(data = data2), stats.predict(r_fixest_pois, newdata = data2)):
-    #    raise ValueError("Predictions for Poisson are not equal")
+    if not np.allclose(
```
s3alfisc, line 158: I have generally started to check for identity only for a select range of predictions. What usually happens is that 1 or 2 out of 1000 values fail a strict tolerance and then it seemed like the better approach to me to keep a stricter tolerance but only test a few values? 
Though I generally think that 1e-07 is more than good enough =) 
```suggestion
    if not np.allclose(
       fepois_mod.predict(newdata=data2, type="link")[11:16],
       stats.predict(r_fixest_pois, newdata=data2, type="link")[11:16],
```

Issue #782, opened 2025-01-06
Mimic `fixest::i()` by relying on formulaic stateful transforms


s3alfisc:
Prompted by https://github.com/matthewwardrop/formulaic/issues/238, implement a proper [i()](https://lrberge.github.io/fixest/reference/i.html) operator and replace the [ugly string parsing](https://github.com/py-econometrics/pyfixest/blob/020f976a15f67e8a3e9a603b7f19d30fb93e23d5/pyfixest/estimation/model_matrix_fixest_.py#L101).

Easily done via `formulaic` stateful transforms: 

```python
import numpy as np
import pandas as pd
from formulaic.transforms import stateful_transform
from formulaic.transforms.contrasts import C, TreatmentContrasts
from formulaic import model_matrix
import pyfixest as pf

data = pf.get_data()

@stateful_transform
def i(factor_var, ref=None, _state=None, _metadata=None, _spec=None):

    if "i" not in _state:
        _state["i"] = C(data = factor_var, contrasts = TreatmentContrasts(ref))

    return _state["i"]

model_matrix("i(f1, ref = 1.0)", data = data).head()
```
Challenge: How to add a second variable that can be interacted with `factor_var`? 

I.e. API as 

```python
@stateful_transform
def i(factor_var, var = None, ref=None, ref2 = None, _state=None, _metadata=None, _spec=None):

    if var is None: 
        if "i" not in _state:
            _state["i"] = Formula(C(data = factor_var, contrasts = TreatmentContrasts(ref)))
    else: 
        if "i" not in _state:
            # this does not work, need to find where : interaction implemented in formulaic
            _state["i"] = C(data = factor_var, contrasts = TreatmentContrasts(ref)) : C(data = var, contrasts = TreatmentContrasts(ref2))

    return _state["i"]
```
Maybe too ambitions (but certainly useful for pooling DiD time periods, i.e. months to years): `bin` and `bin2` arguments to combine multiple fixed effects levels.

leostimpfle:
I've started to look into this and can see a few ways to solve this using `formulaic` without excessive string parsing, see https://github.com/matthewwardrop/formulaic/issues/244#issuecomment-3689527910

leostimpfle:
I think this can be closed for now because of the formula parsing refactor #1362 (although we may want to revisit if https://github.com/matthewwardrop/formulaic/issues/244 moves ahead).

PR #1110, opened 2025-12-22
Match reference level even if quoted

### LINKED ISSUE #1109: Reference level not omitted in `i()` syntax
`pyfixest` does not appear to omit the reference category when using the `i()` syntax. This is not consistent with `fixest`.

In the example below, using `i(cyl, hp, ref='4')` still includes `C(cyl, contr.treatment(base='4'))[4]:hp` which should be omitted.

# `pyfixest`
```
import pandas as pd

import pyfixest as pf
import statsmodels.api as sm

import rpy2.robjects as ro
from rpy2.robjects import pandas2ri
from rpy2.robjects.packages import importr

pandas2ri.activate()
fixest = importr("fixest")
broom = importr("broom")


#%% data
data = sm.datasets.get_rdataset("mtcars", "datasets").data
data["cyl"]  = data["cyl"].astype("string").astype("category")
data["gear"] = data["gear"].astype('string').astype("category")

#%% i-syntax factor and float
formula: str = "mpg ~ i(cyl, hp, ref='4')"
pf.feols(fml=formula, data=data).summary()
```

```
###
Estimation:  OLS
Dep. var.: mpg, Fixed effects: 0
Inference:  iid
Observations:  32
| Coefficient                             |   Estimate |   Std. Error |   t value |   Pr(>|t|) |   2.5% |   97.5% |
|:----------------------------------------|-----------:|-------------:|----------:|-----------:|-------:|--------:|
| Intercept                               |     25.389 |        2.828 |     8.979 |      0.000 | 19.597 |  31.182 |
| C(cyl, contr.treatment(base='4'))[4]:hp |      0.008 |        0.035 |     0.241 |      0.811 | -0.063 |   0.079 |
| C(cyl, contr.treatment(base='4'))[6]:hp |     -0.045 |        0.025 |    -1.810 |      0.081 | -0.096 |   0.006 |
| C(cyl, contr.treatment(base='4'))[8]:hp |     -0.047 |        0.014 |    -3.496 |      0.002 | -0.075 |  -0.020 |
---
RMSE: 3.31 R2: 0.689 
```

# `fixest`

```
r = fixest.feols(ro.Formula(formula), data=data)
r_summary = pd.DataFrame(broom.tidy_fixest(r, conf_int=ro.BoolVector([True]))).T
print(r_summary.set_index(0)[1])
```

```
0
(Intercept)    26.024247
cyl::6:hp      -0.049939
cyl::8:hp      -0.050232
Name: 1, dtype: object
```

leostimpfle:
Close #1109


leostimpfle:
Good point! Yes, there is: https://github.com/py-econometrics/pyfixest/blob/master/tests/test_i.py

I've also just seen that there is an issue to implement it the `i`-syntax differently #782. I'm just starting to familiarise myself with formulaic but would be keen to implement it "properly", so maybe I'll look into this a bit more before proceeding with this PR.

s3alfisc:
I'd love that! I think that we could do much better than the hacky string parsing, both for i() syntax and for multiple estimation syntax by defining a proper syntax tree. But this was way above my skill level back when I initially implemented all of this 😀

PR #1118, opened 2025-12-28
Refactor formula parsing

leostimpfle:
This is a proof of concept for a refactor of PyFixest's formula parsing. The PR introduces a new module [`parse`](https://github.com/py-econometrics/pyfixest/blob/d0a8821656c99a50660fe5166c85e5afd5354805/pyfixest/estimation/formula/parse.py) that refactors formula parsing from the ground up.

The core logic is implemented in [`pyfixest.estimation.formula.parse.parse`](https://github.com/py-econometrics/pyfixest/blob/d0a8821656c99a50660fe5166c85e5afd5354805/pyfixest/estimation/formula/parse.py#L287) which takes in a formula string and returns a collection of parsed formulas represented by [`pyfixest.estimation.formula.parse.Formula`](https://github.com/py-econometrics/pyfixest/blob/d0a8821656c99a50660fe5166c85e5afd5354805/pyfixest/estimation/formula/parse.py#L59).

All references to the old [`FormulaParser`](https://github.com/py-econometrics/pyfixest/blob/d0a8821656c99a50660fe5166c85e5afd5354805/pyfixest/estimation/FormulaParser.py) are bypassed (mostly by renaming the old `FixestFormula` using imports of the form `from pyfixest.estimation.formula.parse import Formula as FixestFormula`)


pyfixest/estimation/formula/parse.py
```diff
@@ -0,0 +1,320 @@
...
+        Separated by "+". "0" if no fixed effect in the model.
+    endogenous : Optional[str]
+        Endogenous variables in the model, separated by '+'.
+    instruments : Optional[str]
+        Instrumental variables for the endogenous variables, separated by '+'.
+    """
+
+    dependent: str
+    independent: str
+    fixed_effects: str = "0"
```
s3alfisc: I am not 100% sure why back in the day I decided it was needed to set the fixed effect placeholder as string "0" - maybe something to adjust in this / a future PR?

pyfixest/estimation/formula/parse.py
```diff
@@ -0,0 +1,320 @@
...
+    parts: re.Pattern = re.compile(r"\s*\|\s*")
+    dependence: re.Pattern = re.compile(r"\s*~\s*")
+    variables: re.Pattern = re.compile(r"\s*\+\s*")
+    args: re.Pattern = re.compile(r"\s*,\s*")
+    multiple_estimation: re.Pattern = re.compile(
+        rf"(?P<key>{'|'.join(e.name for e in _MultipleEstimationType)})\((?P<variables>.*?)\)"
+    )
+
+
+def _parse_parts(formula: str) -> tuple[str, list[str]]:
```
s3alfisc: When I first implemented this, `formulaic` did not support IV syntax. This got eventually merged and now you can specify IVs as 

```python
 Formula("y ~ x1 + x2 + [ x3 + x4 ~ z1 + z2]")
```

This of course does not match r-fixest and I am not super keen to implement any breaking changes. 

On the other hand, there is clear benefit to have this handled by `formulaic` and being able to drop all of the IV parsing code in this section. Maybe one option would be to allow the formulaic-native syntax and handle it in the model_matrix method in a backwards compatible way?

s3alfisc: (Would of course only be relevant for a separate PR if we were to go down this route?)

s3alfisc:
After first look at the code base: much better and much cleaner than before. No fundamental suggestions for improvement from my side. Thank you!

leostimpfle:
Note to self: Look into allowing `formulaic`'s [multi stage formula notation](https://matthewwardrop.github.io/formulaic/latest/migration/#migrating-from-r) in the parser. For more background see https://github.com/matthewwardrop/formulaic/pull/108 and https://github.com/matthewwardrop/formulaic/issues/24

pyfixest/estimation/formula/model_matrix.py
```diff
@@ -18,8 +18,12 @@
 
 
 def _factorize(series: pd.Series, encode_null: bool = False) -> np.ndarray:
-    factorized, _ = pd.factorize(series, use_na_sentinel=True)
-    if not encode_null:
+    factorize: bool = not pd.api.types.is_numeric_dtype(series)
```
s3alfisc: We actually need to ensure the fixed effects are int - this is what the demeaning algorithms expect down the line. That's why this function has been applied unconditionally. One option could be to check if the fixed effects are int and only apply the function if not? For this I'd have to check if there is an expectations that each fixed effect starts at 0, 1, 2, ... but I do not think so.

s3alfisc:
@leostimpfle fixed a bug in the bin() function + adjusted the tests. You can run them via ` pixi r -e dev pytest tests/test_i.py`. Currently give two errors: 


```python
      40 passed
       2 failed
         - tests/test_i.py:286 test_factor_x_factor[Y ~ i(f_str, i.g)-Y ~ i(f_str, g)]
         - tests/test_i.py:304 test_factor_x_factor_with_fe[Y ~ i(f_str, i.g) | fe1-Y ~ i(f_str, g) | fe1]
E           AssertionError: Name mismatch:
E               py=['f_str::apple:g::X', 'f_str::apple:g::Y', 'f_str::apple:g::Z', 'f_str::banana:g::X', 'f_str::banana:g::Y', 'f_str::banana:g::Z', 'f_str::cherry:g::X', 'f_str::cherry:g::Y']
E               r=['f_str::apple:g::Y', 'f_str::apple:g::Z', 'f_str::banana:g::X', 'f_str::banana:g::Y', 'f_str::banana:g::Z', 'f_str::cherry:g::X', 'f_str::cherry:g::Y', 'f_str::cherry:g::Z']
E           assert ['f_str::appl...na:g::Z', ...] == ['f_str::appl...ry:g::X', ...]
E             
E             At index 0 diff: 'f_str::apple:g::X' != 'f_str::apple:g::Y'
```

pyfixest/estimation/feols_.py
```diff
@@ -410,40 +409,50 @@ def _not_implemented_did(*args, **kwargs):
...
-        mm_dict = model_matrix_fixest(
-            FixestFormula=self.FixestFormula,
+        model_matrix = model_matrix_fixest.get(
+            formula=self.FixestFormula,
             data=self._data,
             drop_singletons=self._drop_singletons,
-            drop_intercept=self._drop_intercept,
             weights=self._weights_name,
             context=self._context,
         )
```
s3alfisc, line 420: Going forward, we might also want to simplify / align the naming conventions between e.g. `self._Y` and `model_matrix.dependent` etc?

pyfixest/estimation/feols_.py
```diff
@@ -410,40 +409,50 @@ def _not_implemented_did(*args, **kwargs):
...
+        self._Y_untransformed = model_matrix.dependent.copy()
+        self._X = model_matrix.independent
+        self._fe = model_matrix.fixed_effects
+        self._endogvar = model_matrix.endogenous
+        self._Z = model_matrix.instruments
+        self._weights_df = model_matrix.weights
+        # self._na_index = model_matrix.get("na_index")
+        self._na_index_str = model_matrix.na_index_str
+        # TODO: set dynamically based on naming set in pyfixest.estimation.formula.factor_interaction._encode_i
+        is_icovar = (
```
s3alfisc, line 430: Do we still need the `python is_icovar` variables and the like with this PR?

s3alfisc: Yes, still needed in pilot plus in the did2s function which still relies on the old `model_matrix_fixest`. We can either deprecate in this PR or in a follow up.

pyfixest/estimation/formula/model_matrix.py
```diff
@@ -0,0 +1,233 @@
...
+        factorized, _ = pd.factorize(series, use_na_sentinel=True)
+    else:
+        factorized = series.to_numpy()
+    if not encode_null and factorize:
+        # Keep nulls (otherwise they are encoded as -1 when use_na_sentinel=True)
+        factorized = np.where(factorized == -1, np.nan, factorized)
+    return factorized
+
+
+def _interact_fixed_effects(fixed_effects: str, data: pd.DataFrame) -> pd.DataFrame:
```
s3alfisc: In the spirit of this PR, we should also move this to a stateful transform in formulaic? But can be done in a separate PR?

pyfixest/estimation/formula/model_matrix.py
```diff
@@ -0,0 +1,233 @@
...
+def _get_weights(data: pd.DataFrame, weights: str) -> pd.Series:
+    if weights not in data.columns:
+        raise ValueError(f"The weights column '{weights}' is not a column in the data.")
+    w = data[weights]
+    try:
+        w = pd.to_numeric(w, errors="raise")
+    except ValueError:
+        raise ValueError(f"The weights column '{weights}' must be numeric.")
+    if not (w.dropna() > 0.0).all():
+        raise ValueError(
```
s3alfisc: We already check that the weights column exists in the [estimation_input_check](https://github.com/py-econometrics/pyfixest/blob/d83e15ba24790a69b4da2605d80a60b5aae242ee/pyfixest/estimation/estimation.py#L1351) - should we do it in both spots (likely not), and at this point, or earlier in the code flow?

pyfixest/estimation/formula/model_matrix.py
```diff
@@ -0,0 +1,233 @@
...
+    dependent: pd.DataFrame
+    independent: pd.DataFrame
+    model_spec: formulaic.ModelSpec
+    na_index_str: str
+    fixed_effects: Optional[pd.DataFrame] = None
+    endogenous: Optional[pd.DataFrame] = None
+    instruments: Optional[pd.DataFrame] = None
+    weights: Optional[pd.DataFrame] = None
+
+    def __post_init__(self) -> None:
```
s3alfisc: Should we test here that: 
- all variables have the same number of observations (status quo)
- are 2dimensional and 1dimensional when expected (also useful for code documentation purposes)
- all variables are numeric and fixed_effects are int32 (also useful for documentation purposes)?


I find the initial section very hard to read - though this might be due to my lack of exposure to core python code =)

pyfixest/estimation/formula/model_matrix.py
```diff
@@ -0,0 +1,233 @@
...
+    -------
+    ModelMatrix
+
+    """
+    # Process input data
+    data.reset_index(drop=True, inplace=True)  # Sanitise index
+    n_observations: Final[int] = data.shape[0]
+    # Set infinite to null
+    numeric_columns = data.select_dtypes(include="number").columns
+    data[numeric_columns] = data[numeric_columns].where(
```
s3alfisc: Maybe we can bring the warning message back? 

```suggestion
inf_mask = ~np.isfinite(data[numeric_columns])
if inf_mask.any().any():
    warnings.warn("Infinite values detected and will be treated as missing.")
    data[numeric_columns] = data[numeric_columns].where(np.isfinite(data[numeric_columns]), pd.NA)
```

s3alfisc: Plus, add a comment that we need to do so as `formulaic` drops NaN values by default, but not Inf values (I think R would handle this differently).

pyfixest/estimation/model_matrix_fixest_.py
```diff
@@ -8,7 +8,7 @@
 from formulaic import Formula
```
s3alfisc, line 8: Should we strive to fully delete / deprecate this function within this PR? Would require to adjust the DID2S code but not much else I believe?

pyfixest/estimation/formula/parse.py
```diff
@@ -0,0 +1,435 @@
...
+            independent = f"{independent}-1"
+        formula = f"{self.dependent}~{independent}"
+        if self.endogenous is not None and self.instruments is not None:
+            formula = f"{formula}|{self.endogenous}~{self.instruments}"
+        if self.fixed_effects is not None:
+            formula = f"{formula}|{self.fixed_effects}"
+        return formula
+
+    @property
+    def fml_first_stage(self) -> str | None:
```
s3alfisc: Looking at this, it took me a while to figure out why we'd need not add the fixed effects to the fml first and second stages. I am updating the docstrings to make this clearer, but I also think that maybe we should rethink how things are set up? Because ex ante, I (and more casual users likely as well) would assume that the first and second stage formulas contain the fixed effects?

pyfixest/estimation/formula/parse.py
```diff
@@ -0,0 +1,455 @@
...
+            formula = Formula(intercept=self.intercept, **kwargs)
+            estimations[formula.fixed_effects].append(formula)
+        return estimations
+
+
+@dataclass(frozen=True)
+class _Pattern:
+    parts: re.Pattern = re.compile(r"\s*\|\s*")
+    dependence: re.Pattern = re.compile(r"\s*~\s*")
+    variables: re.Pattern = re.compile(r"\s*\+\s*")
```
s3alfisc: Technically, I think users could also 'subtract' variables as `X1 - X1 + X2` . This is something the status quo implementation also ignores.

pyfixest/estimation/formula/parse.py
```diff
@@ -0,0 +1,455 @@
...
+        ]
+        if instruments_are_covariates:
+            raise InstrumentsAsCovarsError(
+                f"Instruments specified as covariates: {instruments_are_covariates}"
+            )
+        if len(endogenous) > len(instruments):
+            raise UnderDeterminedIVError(
+                "The IV system is underdetermined. "
+                "Please provide as many or more instruments as endogenous variables."
+            )
```
s3alfisc: We should be able to delete these checks, as multiple estimation for IV is already blocked in `ParsedFormula`? 

```python
   def __post_init__(self):
        if self.is_multiple and self.is_iv:
            raise NotImplementedError(
                "Multiple Estimations is currently not supported with IV. "
                "This is mostly due to insufficient testing and will be possible with a future release of PyFixest."
            )
```

s3alfisc:
Review status: parser looks great 👍 (plus a few smaller comments). Will continue with i-operator in a bit.

s3alfisc (changes requested):

pyfixest/estimation/formula/parse.py
leostimpfle: We now use `None` when there are no fixed effects

pyfixest/estimation/formula/parse.py
leostimpfle: Needs further investigation, see discussion here: https://github.com/matthewwardrop/formulaic/issues/264

pyfixest/estimation/formula/model_matrix.py
leostimpfle: The fixed effects are cast to `int32` before apply singleton detection (see [here](https://github.com/py-econometrics/pyfixest/blob/0e0402d977bbe27ec6f1992d578f385b7cc46970/pyfixest/estimation/formula/model_matrix.py#L190)).

I've added this because we have test cases of the form [`"Y ~ X1 + X2:f1 | f3 + f1"`](https://github.com/py-econometrics/pyfixest/blob/e23e7b2b81432c5547afab879be91bdb0594bad2/tests/test_vs_fixest.py#L54) which are currently interpreted as the interaction of two numerical variables `X2:f1`. I'm now not sure if this is intentional or if this should be the interaction of a continuous variable `X2` with a factor `f1`?

In any case, we need to fix this because the casting to `int32` clearly is wrong when the input are floats.

pyfixest/estimation/feols_.py
leostimpfle: Yes! I was actually wondering if we `Feols` should have a single attribute for a `ModelMatrix` object, so that we can use `self.model_matrix.dependent` etc?

For now, I'v tried to keep code changes to a minimal level only replacing the parsing/model matrix creation hence, tried to not rename any other attributes.

pyfixest/estimation/formula/model_matrix.py
leostimpfle: Yes, we can do it now. Will look into it

pyfixest/estimation/formula/model_matrix.py
leostimpfle: Probably earlier? Happy to remove the checks on weights in the model matrix creation

pyfixest/estimation/formula/model_matrix.py
leostimpfle: Yes! I'm still not happy with this part. I was wondering if it would be best to convert the `formulaic.ModelMatrix` object back to a data frame-like structure with a single index? This would enable simple checks on consistency of index as well as consistent modification (e.g., drop singletons). I was also to see if `formulaic` allows modification of a `ModelMatrix` across all parts but couldn't find any convenience functions

pyfixest/estimation/formula/model_matrix.py
leostimpfle: Yes, we should bring back the warning. Are we happy with dropping non-finite values at his point or should be revert to doing it post encoding?

This is important in certain cases which I only realised just now. The current code allows infinite values until the encoding step which is explicitly used in [`_saturated_event_study`](https://github.com/py-econometrics/pyfixest/blob/35b51b1a79e2660525244a2305d2eff84b7e7dfb/pyfixest/did/saturated_twfe.py#L315-L321): `rel_time` is `np.inf` for never treated units but categorically encoded (so `np.inf` is a single level). The previous code allowed `np.inf` as a float because it was encoded in `formulaic`, but this current version would drop it before encoding.

pyfixest/estimation/model_matrix_fixest_.py
leostimpfle: Yes, my plan is to remove both `model_matrix_fixest_` and `FixestFormulaParser` before merging this PR

pyfixest/estimation/feols_.py
s3alfisc: > Yes! I was actually wondering if we Feols should have a single attribute for a ModelMatrix object, so that we can use self.model_matrix.dependent etc?

Ok, then let's discuss this in the future (after merging this PR)! 

There are some other related cleanups needed - e.g. I dislike that design matrices etc are pre-multiplied with sqrt(weights), which makes the code hard to read / forces the reader to check if `self._X` has weights applied or not.

pyfixest/estimation/formula/model_matrix.py
s3alfisc: > The fixed effects are cast to int32 before apply singleton detection (see [here](https://github.com/py-econometrics/pyfixest/blob/0e0402d977bbe27ec6f1992d578f385b7cc46970/pyfixest/estimation/formula/model_matrix.py#L190)).

```python 
fixed_effects = (
        model_matrix[_ModelMatrixKey.fixed_effects].astype("int32")
        if formula.fixed_effects is not None
        else None
    )
```

Is this as safe and general as calling `pd.factorize`? As far as I recall it, factorize would also convert strings and other data types, this would not be guaranteed here? Or am I missing even more?

> interaction of two numerical variables X2:f1

Actually desired in this case! `f1` is treated as continuous but `f1` as fixed effect. This of course is somewhat contradictory ...

pyfixest/estimation/formula/model_matrix.py
s3alfisc: Sounds good - I would keep the check for non-negative weights plus weights being numeric? The former is a must have imo as all math operations will continue to work, while the ladder while "fail loudly" if the weights vector is e.g. a string?

pyfixest/estimation/formula/model_matrix.py
s3alfisc: Oh, true! Given this, maybe I'd revert to post encoding?

leostimpfle: Moved all checks into `ModelMatrix` class, so after `formulaic`'s encoding in https://github.com/py-econometrics/pyfixest/pull/1118/commits/8a1ec7469822347c151440fcec35bad108ddc6d2

s3alfisc:
pre-commit.ci autofix

pyfixest/estimation/formula/model_matrix.py
leostimpfle: Have reverted to encoding fixed effects columns in `get_model_matrix`, so everything should now be consistent with the old methodology.

pyfixest/estimation/formula/parse.py
leostimpfle: Good point! This will be non-trivial to implement because parsing `-` would require keeping track of a variable's sign. I think the only issue would be that a formula of the form `X2 - X1 + X1` would be parsed as if `X1` was present, so checks against specification of instrumental variable specifications might fail. (The formula would be passed correctly to `formulaic` though.)

pyfixest/estimation/model_matrix_fixest_.py
leostimpfle: See #1123

pyfixest/estimation/formula/parse.py
leostimpfle: Are we happy with the implementation as documented now @s3alfisc? For example, I haven't spent much thought on whether we want to keep the separate attributes `fml` and `fml_second_stage` but agree that it is not really intuitive

pyfixest/estimation/formula/parse.py
leostimpfle: This check will become relevant thought once multiple estimation is implemented, or am I missing something?

leostimpfle:
@s3alfisc I've gone over the PR once more today. The biggest change is that I have rewritten `pyfixest.estimation.formula.factor_interaction.factor_interaction` to make the implementation (hopefully) easier to follow. More importantly, I have removed the monkey-patch of `formulaic` so we do not actually need the `drop` argument discussed in https://github.com/matthewwardrop/formulaic/pull/263

I've also let Claude improve the docstrings and renamed a few functions/attributes for purely aesthetic reasons. The PR is now ready from my side (although there are many other potential improvements around formula parsing and model matrix construction: #1125, #1126, #1127, #1130).

s3alfisc:
Review in 4 steps: 

- [x] `parse` function 
- [ ] `ModelMatrix` class 
- [ ] new ```python i()``` interaction 
- [ ] downstream impact on `did2s` and `saturated_twfe`

pyfixest/estimation/formula/parse.py
```diff
@@ -0,0 +1,474 @@
...
+        )
+    dependent, independent = (
+        re.split(_Pattern.variables, variables)
+        for variables in re.split(_Pattern.dependence, string=part)
+    )
+    return dependent, independent
+
+
+def _parse_fixed_effects(parts: list[str]) -> list[str] | None:
+    part_fe: str | None = next((part for part in parts if "~" not in part), None)
```
s3alfisc: I think this line might be dangerous? 

Fixed effects are detected as the formula part with "no" tilde. 

But we might pass formulas like `Y ~ X1 | f1 | f2` . We probably need to add a check that ensures 

- if only one model part, it has to have a tilde
- if more than one model part, we have to have two tildes 

If two parts and tilde in position 1, 2, then it is IV. 
If three parts, always IV. 

Tilde positions: 
- for one part: only position 1
- two parts: either one or two 
- three parts: one or three

We should add tests to ensure that `Y ~ X1 | f1 | f2` does not pass through.

s3alfisc: I think this should have been addressed with [a520f06](https://github.com/py-econometrics/pyfixest/pull/1118/commits/a520f06f08a3df37c535a01a5769c6caf5e0b74a).

s3alfisc:
## parse

- I am still not a big fan of `Formula.first_stage`and `Formula.second_stage`not containing fixed effects - potentially misleading to users despite documentation? Maybe we should add `Formula.first_stage_no_fixed_effects` etc as extra attributes to make more explicit what type of formula users are dealing with? 
- Can you specify the reason for the `FORMULAIC_FEATURE_FLAG is DefaultFormulaParser.FeatureFlags.ALL` in several spots in the code base? Why is it needed? Are there potential downsides? 
- Is the `sort`argument in `parse`still needed? 

I committed a few changes, I hope all of these make sense to you @leostimpfle and are more or less self-explanatory by the commit message?

leostimpfle:
> * I am still not a big fan of `Formula.first_stage`and `Formula.second_stage`not containing fixed effects - potentially misleading to users despite documentation? Maybe we should add `Formula.first_stage_no_fixed_effects` etc as extra attributes to make more explicit what type of formula users are dealing with?

Agreed that this is somewhat unintuitive. An alternative to changing the attribute names could be to include the encoded fixed effects directly in the formula. For example, instead of `formula_kwargs = {'second_stage': 'Y ~ X1', 'fixed_effects' : 'f1 + f2'}`, we could use `formula_kwargs = {'second_stage': 'Y ~ X1 + __fixed_effect__(f1) + __fixed_effect__(f2)'}` (where the sentinel `__fixed_effect__` indicates the integer encoding of fixed effects). The main point is that the latter formula is what we already pass implicitly to `formulaic`, so in this approach we should call the attribute `second_stage_formulaic`.

> * Can you specify the reason for the `FORMULAIC_FEATURE_FLAG is DefaultFormulaParser.FeatureFlags.ALL` in several spots in the code base? Why is it needed? Are there potential downsides?

This is a hangover from my early attempts to use `formulaic`'s multistage syntax (see #1125). `DefaultFormulaParser.FeatureFlags.ALL` indicates that the multistage syntax is enabled but the `FORMULAIC_FEATURE_FLAG` is set to `DefaultFormulaParser.FeatureFlags.DEFAULT` (i.e., multistage syntax is disabled). For clarity, I have removed references to `FORMULAIC_FEATURE_FLAG` in the parser for now.

> * Is the `sort`argument in `parse`still needed?

Not needed, and I have removed it

> I committed a few changes, I hope all of these make sense to you @leostimpfle and are more or less self-explanatory by the commit message?

Yes, all good. Thanks @s3alfisc!

s3alfisc:
Ok, now refactored into 4 PRs - #1186 , #1187, #1188 , #1189 .

leostimpfle:
Closed in favour of #1186

PR #1132, opened 2026-01-06
Close #1122

### LINKED ISSUE #1122: Clean up tests to avoid `FutureWarning: ChainedAssignmentError`
There are many tests that trigger the avoidable [`FutureWarning: ChainedAssignmentError`](https://pandas.pydata.org/docs/reference/api/pandas.errors.ChainedAssignmentError.html) which is caused by assigning a value to a copy of a `pandas.DataFrame` (see, e.g. [here](https://github.com/py-econometrics/pyfixest/actions/runs/20623579632/job/59230142706?pr=1118#step:7:330) which is caused by [this line](https://github.com/py-econometrics/pyfixest/blob/d83e15ba24790a69b4da2605d80a60b5aae242ee/tests/test_errors.py#L52)). It would be nice to update the affected tests to reduce the amount of clutter in the test output.

leostimpfle:
Few adjustments to avoid `FutureWarning: ChainedAssignmentError` and `SettingWithCopyWarning`. Closes #1122

### DIFF
--- a/pyfixest/estimation/fepois_.py
+++ b/pyfixest/estimation/fepois_.py
@@ -175,14 +175,12 @@ class Fepois(Feols):
             )
 
         if na_separation:
-            self._Y = self._Y.drop(na_separation, axis=0, inplace=False)
-            self._X = self._X.drop(na_separation, axis=0, inplace=False)
-            self._fe = self._fe.drop(na_separation, axis=0, inplace=False)
-            self._data = self._data.drop(na_separation, axis=0, inplace=False)
+            self._Y.drop(na_separation, axis=0, inplace=True)
+            self._X.drop(na_separation, axis=0, inplace=True)
+            self._fe.drop(na_separation, axis=0, inplace=True)
+            self._data.drop(na_separation, axis=0, inplace=True)
             if self._weights_df is not None:
-                self._weights_df = self._weights_df.drop(
-                    na_separation, axis=0, inplace=False
-                )
+                self._weights_df.drop(na_separation, axis=0, inplace=True)
             self._N = self._Y.shape[0]
             self._N_rows = self._N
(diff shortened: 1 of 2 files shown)


pyfixest/estimation/fepois_.py
```diff
@@ -175,12 +175,14 @@ def prepare_model_matrix(self):
             )
 
         if na_separation:
-            self._Y.drop(na_separation, axis=0, inplace=True)
-            self._X.drop(na_separation, axis=0, inplace=True)
-            self._fe.drop(na_separation, axis=0, inplace=True)
-            self._data.drop(na_separation, axis=0, inplace=True)
+            self._Y = self._Y.drop(na_separation, axis=0, inplace=False)
```
s3alfisc: Why can't we do this in place? Would it not be better to just add an ignore flag? Not doing the operation in place is likely less efficient?

leostimpfle:
@s3alfisc Can we merge this small fix? 🙂

pyfixest/estimation/fepois_.py
s3alfisc: Approved, only minor comment was this ⬆️

leostimpfle: The `SettingWithCopyWarning` indicates that the attributes are not properly set upstream. You're right that we should keep it inplace and fix the root cause (probably where they assigned for the first time). Will investigate

s3alfisc: Thanks!

leostimpfle: The warning was triggered by tests in `test_vs_fixest.py` where views of the test data were repeatedly modified. I have reverted to dropping separated observations inplace and cleaned up the tests.

Issue #1144, opened 2026-01-20
`Fepois` and `GLM` models fail to detect multicollinearity of independent variables nested within fixed effects


leostimpfle:
`Fepois` fails to detect multicollinearity of independent variables nested within fixed effects. To fix this it should be sufficient to check for multicollinearity after demeaning in the first iteration of the IWLS algorithm: https://github.com/py-econometrics/pyfixest/blob/18bea76b73106f09c301ba51d84ec835ecf4b4a9/pyfixest/estimation/fepois_.py#L303-L310

# To reproduce
``` python
import numpy as np
import pandas as pd

import pyfixest as pf

np.random.seed(42)

n = 200
units = 20

unit_id = np.repeat(range(units), n // units)
unit_fixed_effect = np.array([i % 3 for i in unit_id])
year = np.random.choice(range(2015, 2020), size=n)
independent = np.random.choice([0, 1], size=n)
linear_pred = 0.5 * independent - 0.3 * unit_fixed_effect + np.random.normal(0, 0.5, n)

data = pd.DataFrame(
    {
        "verified": np.random.poisson(np.exp(linear_pred)),
        "independent": independent,
        "unit_fixed_effect": unit_fixed_effect,
        "unit_id": unit_id,
        "year": year,
    }
)

# This works: unit_fixed_effect nested within unit_id dropped due to multicollinearity
pf.feols(
    fml="verified ~ C(independent)*C(unit_fixed_effect) | unit_id + year",
    data=data,
    vcov={"CRV1": "unit_id"},
)

# This doesn't work: multicollinearity not properly detected
pf.fepois(
    fml="verified ~ C(independent)*C(unit_fixed_effect) | unit_id + year",
    data=data,
    vcov={"CRV1": "unit_id"},
)
```

s3alfisc:
Confirmed bug, thanks @leostimpfle !

s3alfisc:
Slightly related to https://github.com/py-econometrics/pyfixest/issues/1139

s3alfisc:
As the glm fixed effect PR is merged, this is also relevant for the GLM class!

PR #1154, opened 2026-01-29
move multicollinear checks into all estimation classes

s3alfisc:
Close to (but does not close) #1144. all multicollinearity checks are moved to the fit method.


s3alfisc:
cc @leostimpfle

PR #1219, opened 2026-03-05
Modular Benchmarks

s3alfisc:
@schroedk 

Run via 

```
pixi run -e dev python benchmarks/modular/benchmark_main.py
```


s3alfisc:
@schroedk from my side it would be ok to merge this into master - so please do it if it facilitates iteration for you.

PR #1221, opened 2026-03-06
Feature/within demeaner

schroedk:
## Summary

  This PR adds a new Rust-based demeaning path using the within solver stack and wires it into pyfixest as an opt-in backend.

  ## Motivation

  Alternating-projections demeaning can be slow on some FE graphs (for example weakly connected/blocky structures).
  This adds a preconditioned-CG based alternative via within for better convergence behavior in those cases.

  ## Changes

  - Added Rust binding for within-based demeaning (_demean_within_rs) and exported it from the core extension.
  - Added Python wrapper demean_within(...) in pyfixest.core.
  - Added new backend option: rust-cg.
  - Registered rust-cg in backend dispatch and backend literal typing.
  - Extended demeaning benchmark coverage to include demean_within.
  - Switched within dependency from local path to GitHub-hosted dependency, pinned to:
      - ac5a859e82dda01ff863c46cc78f675d55c22a3c
  - Updated Cargo.lock accordingly.

  ## User-facing impact

  - New opt-in demeaning backend is available (rust-cg).
  - Existing behavior/backends remain unchanged unless explicitly selected.

  ## Validation

  - cargo check passes with the new dependency wiring.
  - Python files touched by this work compile cleanly (py_compile).

  ## Notes

  - Dependency is currently git+ssh; CI/build environments need SSH GitHub access.
  - If needed, this can be switched to HTTPS + token-based auth for CI compatibility.

### DIFF
--- a/pyfixest/estimation/internals/demean_.py
+++ b/pyfixest/estimation/internals/demean_.py
@@ -350,6 +350,10 @@ def _set_demeaner_backend(
         from pyfixest.core.demean import demean as demean_rs
 
         return demean_rs
+    elif demeaner_backend == "rust-cg":
+        from pyfixest.core.demean import demean_within
+
+        return demean_within
     elif demeaner_backend == "numba":
         return demean
     elif demeaner_backend == "jax":


s3alfisc:
Looks good to me - also please feel free to merge this (or tell me to merge) if it helps with iterations @schroedk

s3alfisc:
Given the fact that the cgs algo is non-deterministic, should we add a demeaner_options dict that allows user to pass a seed and other function arguments relevant for the demeaner algorithms? Relevant args are `fixef_tol`, `fixef_maxiter`. For the LSMR, we set `atol = btol = fixef_tol` and could allow for more explicit argument naming?

s3alfisc:
Plus, should I add benchmarks for the full pyfixest - feols pipeline?

PR #1222, opened 2026-03-06
Enhance formula parsing

### LINKED ISSUE #684: Bug: `predict()` method fails with new data and pyfixest special syntax - `i()`, `C()`, `var1^var2` operators
Examples: 

```python
%load_ext autoreload
%autoreload 2 

import pyfixest as pf 
data = pf.get_data()

#fit = pf.feols("Y ~ i(X1) | f1", data = data)
#fit.predict(newdata = data.iloc[0:100])
#FactorEvaluationError: Unable to evaluate factor `i(X1)`. [NameError: name 'i' is not defined]

fit = pf.feols("Y ~ X1 | f1^f2", data = data)
fit.predict(newdata = data.iloc[0:100])
# KeyError: "None of [Index(['f1^f2'], dtype='object')] are in the [columns]"
```

### LINKED ISSUE #1125: Implement `formulaic`'s multistage syntax
When I first implemented this, `formulaic` did not support IV syntax. This got eventually merged and now you can specify IVs as 

```python
 Formula("y ~ x1 + x2 + [ x3 + x4 ~ z1 + z2]")
```

This of course does not match r-fixest and I am not super keen to implement any breaking changes. 

On the other hand, there is clear benefit to have this handled by `formulaic` and being able to drop all of the IV parsing code in this section. Maybe one option would be to allow the formulaic-native syntax and handle it in the model_matrix method in a backwards compatible way?

_Originally posted by @s3alfisc in https://github.com/py-econometrics/pyfixest/pull/1118#discussion_r2649663970_

### LINKED ISSUE #1213: Use `formulaic`'s `MULTIPART` syntax
`formulaic` already supports `MULTIPART` formulas using the `|` operator (see [here](https://matthewwardrop.github.io/formulaic/dev/guides/grammar/)). We should just use this feature directly.

### LINKED ISSUE #1218: Implement fixed effects encoding as `formulaic` transformation
Move the encoding of fixed effects to inside the model matrix construction: https://github.com/py-econometrics/pyfixest/blob/dec2f1b509c7590e5bc771df8359d16d5dd639c9/pyfixest/estimation/formula/model_matrix.py#L356-L358

leostimpfle:
- Enable `formulaic`'s `MULTISTAGE` syntax #1125 
- Enable `formulaic`'s `MULTIPART` syntax #1213 
- Use `stateful_transform` for fixed effects encoding #1218 

# To do

- [ ] The `^` operator is currently reserved for interactions of fixed effects (see [here](https://github.com/py-econometrics/pyfixest/blob/39409167419fbe1e4b7a77d615e4733c4357ca39/pyfixest/estimation/formula/parse.py#L34-L51)). This is necessary only because the operator expects a number on the right side (e.g., `X^2`). We could avoid this if we decided to use a different symbol for fixed effect interactions (e.g., `f1:f2`).
- [ ] Validate that the `stateful_transform` works correctly on the captured context (see [here](https://github.com/py-econometrics/pyfixest/blob/39409167419fbe1e4b7a77d615e4733c4357ca39/pyfixest/estimation/formula/transforms/fixed_effects_encoding.py#L5-L14) and related to #769).
- [x] Set transforms consistently across project to address  #684
- [ ] Clean up tests to remove deprecated syntax (`Y+Y2 ~X1` should be replaced with `sw(Y, Y2) ~ X1`; `Y ~ X1 | X2 ~ Z2` with `Y ~ X1 + [X2 ~ Z2]`)
- [ ] `formulaic.formula.Formula.required_variables` calls `ast.parse(factor.expr, mode="eval")` which might fail if variable names are not valid Python expressions (e.g., `0_X2_1`). This, however, is necessary to account for transformed variables (e.g., `Y ~ C(X2) + [X2 ~ Z2]`).
- [x] Guard against specification of multiple multistage parts. For example, we do not support `Y ~ [X1 ~ Z1] + [X2 ~ Z2]`.
- [x] Guard against specification of more than two formula parts (we only allow a single second part for fixed effects).


leostimpfle:
@s3alfisc The core changes to formula parsing are now ready. Feel free to have a look and let me know any comments. I'll continue to work through the to-dos listed above.

s3alfisc:
Just took a quick look and was thinking - maybe it would be best if we'd do a pair review for this through a call? I would make sure to spent some time with it before to come prepared 😃

leostimpfle:
Very happy to jump on a call!

leostimpfle:
Superseded by #1362

PR #1217, opened 2026-03-05
Migrate pixi config to pyproject.toml with multi-Python environments

timmens:
## Migrate pixi config to pyproject.toml and clean up CI

Moves all pixi configuration from `pixi.toml` into `pyproject.toml` and overhauls the CI setup.

### CI
- **Renamed jobs** for clarity: `test` → `test-python`, `test_slow` → `test-r`
- **R tests run in parallel** — each test task (`test-r-core`, `test-r-extended`, `test-r-fixest`, `test-r-hac`) is now a separate matrix job instead of running sequentially
- **Added macOS** to the Python test matrix
- Fixed coverage uploads — previously the sequential R tests overwrote each other's `coverage.xml`

### Pixi tasks
- Task names now indicate their environment: `test-py-*` for Python, `test-r-*` for R
- Internal tasks (`_setup`, `_update-test-data`) are hidden with `_` prefix
- All user-facing tasks have descriptions (visible via `pixi task list`)
- Replaced default pytest parallelism

### Other
- Deleted `pytest.ini` (config moved to `pyproject.toml`)
- Updated contributing docs

PR #1231, opened 2026-03-07
Fix/pycharm conda prefix

leostimpfle:
I've run into issues executing files in PyCharm after the recent pixi updates in #1217 which appear to be related to the maturin import hook.

This PR solves this by adjusting `scripts/setup_maturin_hook.py` to add `CONDA_PREFIX` if not set (which is the case when calling the environment from PyCharm). (I additionally have re-added [`pixi-pycharm`](https://pixi.prefix.dev/latest/integration/editor/jetbrains/) to the conda dependencies which I have incorrectly removed in #1209)

This is mostly Claude's work but I have tested that this works reliably with PyCharm and doesn't break any other `pixi` commands. Please let me know if there's a better way to resolve this @timmens. This might also be relevant for @schroedk.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -5,6 +5,8 @@ environments:
     - url: https://conda.anaconda.org/conda-forge/
     indexes:
     - https://pypi.org/simple
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
     packages:
       linux-64:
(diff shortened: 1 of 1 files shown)


s3alfisc:
I have no clue about this at all, but I guess this will hardly break anything, so feel free to merge if it helps you =)

PR #1239, opened 2026-03-08
Pytest sugar

leostimpfle:
Disable verbose option `-v` in `tool.pytest.ini_options]` for compatibility with `pytest-sugar`.

I much prefer the outputs of `pytest-sugar` but please let me know if `-v` is preferred @s3alfisc ?


s3alfisc (approved):
Yes, that's much better 👍 this used to be the default but changed somehow? 🤔

PR #1242, opened 2026-03-09
fix: add env bin to PATH in sitecustomize

leostimpfle:
The pixi-pycharm conda shim does not source the environment's activate.d scripts, so sys.prefix/bin is absent from PATH. The maturin-import-hook calls find_maturin() at import time and raises MaturinError when the binary is not on PATH. Prepend sys.prefix/bin to PATH in the generated sitecustomize.py to fix this. The same fix also ensures R (needed by rpy2) is discoverable without a separate R_HOME workaround.

### DIFF
--- a/scripts/setup_maturin_hook.py
+++ b/scripts/setup_maturin_hook.py
@@ -41,6 +41,12 @@ try:
     if not os.environ.get("CONDA_PREFIX") and not os.environ.get("VIRTUAL_ENV"):
         os.environ["CONDA_PREFIX"] = sys.prefix
 
+    # Ensure the environment's bin directory is on PATH so that maturin can be found.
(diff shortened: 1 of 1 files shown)


timmens:
@leostimpfle @s3alfisc 

If the maturin import hook is causing so many unexpected problems, I would consider removing it. I guess we can simply tell the rust developers to run the compilation manually..

I will write an issue soon about splitting pyfixest into a pure-python no-arch package and a pyfixest-core package which collects the rust kernels. In this case also development would get much simpler and there would be no need for the maturin import hook.

leostimpfle:
As far as I can tell everything pixi-related works nicely with the maturin import hook. 

The issues I've encountered are simply related to PyCharm not integrating directly with pixi (but only through the pixi-pycharm shim)

PR #1264, opened 2026-03-28
Proposal for Caching the Within-Preconditioner

s3alfisc:
## Summary

This PR introduces a typed fixed-effects demeaning API, passes it through the code base, and adds reusable `within` preconditioners to speed up repeated multi-way FE estimation.

The main user-facing additions are:

- We introduce typed `demeaner=` objects:
  - `MapDemeaner(...)`
  - `WithinDemeaner(...)`
  - `LsmrDemeaner(...)`
- `demeaner` now takes precedence over `demeaner_backend`, `fixef_tol`, and `fixef_maxiter`
- `feols` now exposes `fit.preconditioner_` for `WithinDemeaner`
- users can reuse that object via `WithinDemeaner(preconditioner=...)`
- `WithinPreconditioner` can be pickled across sessions, following the upstream `within-py` pattern

### 1. Typed demeaner API
We now support a typed `demeaner=` interface instead of relying only on string backend selection via `demeaner_backend`. This makes the demeaner backend configuration explicit, extensible, and easier to validate.

The typed demeaner objects control actual runtime execution.

- `WithinDemeaner` options flowing into the `within` backend
- `LsmrDemeaner` options flowing into the SciPy/CuPy solvers
- `MapDemeaner` options flowing into the MAP solvers 
- 
### 2. Reusable `within` preconditioners

For multi-way FE `WithinDemeaner` fits, we now build and cache a reusable `WithinPreconditioner`.

That preconditioner is:
- exposed on `feols` as `fit.preconditioner_`
- reusable across later estimations via `WithinDemeaner(preconditioner=...)`
- reused internally for multiple estimation syntax 
- reused internally across IWLS iterations in `feglm` and `fepois`

### 3. IWLS reuse

For `feglm` and `fepois`, we now reuse a fit-local `within` preconditioner across IWLS iterations.

Current policy is intentionally simple:
- build once
- reuse by default
- refresh once when inner FE tolerance tightens

@schroedk - For generalized linear models, we need to demean repeatedly with "different weights". Here we currently say it is better to run on "stale" preconditions for a while than to update the preconditioner in every iteration. This can of course be refined later. 

### 4. Persistence

`WithinPreconditioner` is pickleable across sessions.

This mirrors `within-py` closely:
- the Rust `FePreconditioner` is serialized with `postcard`
- Python pickle support is implemented through `__reduce__`

### Why expose preconditioners, but not solvers?

We intentionally expose **preconditioners** as reusable objects, but **not solvers**.

Because reusing a `within` preconditioner is where most of the practical speedup comes from. It is "ok" if the preconditioner becomes "a bit stale" - this is a feature, not a bug imo. Plus, we do not need to check that the fixed effects and the sample and weights are identical, which is what we would have needed to do if we allowed users to pass solvers. 

For iterated weighted least squares as used for GLMs, the preconditioner is the better choice as in each iteration, a stale preconditioner will be "good enough". 

### Stay close to `within-py`

Tried to stay close to the implementation in `within-py`.

In particular:
- preconditioners are built through the same conceptual flow as upstream
- reused preconditioners are passed back into the solver through the same upstream abstractions
- persistence uses the same postcard + pickle pattern as `within-py`

## Workflow

Typical usage now looks like:

```python
fit1 = pf.feols(
    "Y ~ X1 | f1 + f2",
    data=data,
    demeaner=pf.WithinDemeaner(),
)

# use the cached preconditioner for the next fit
fit2 = pf.feols(
    "Y ~ X2 | f1 + f2",
    data=data,
    demeaner=pf.WithinDemeaner(preconditioner=fit1.preconditioner_),
)
```
For persistent reuse, you can do: 

```python
import pickle

payload = pickle.dumps(fit1.preconditioner_)
restored = pickle.loads(payload)
```
And then fit the regression model.


pyfixest/core/demean.py
```diff
@@ -1,7 +1,220 @@
...
+        return _serialize_within_preconditioner_rs(self._preconditioner_handle)
+
+    @classmethod
+    def _from_pickle(
+        cls,
+        data: bytes,
+        n_obs: int,
+        n_factors: int,
+        factor_cardinalities: tuple[int, ...],
+        preconditioner_type: str,
```
leostimpfle, line 74: Can we use a `typing.Literal` for clarity here and anywhere `preconditioner_type` is used? For example, ` PreconditionerType = Literal["additive", "multiplicative"]`

pyfixest/demeaners.py
```diff
@@ -0,0 +1,170 @@
...
+        if backend not in {"numba", "rust", "jax"}:
+            raise ValueError("`backend` must be one of 'numba', 'rust', or 'jax'.")
+        object.__setattr__(self, "backend", backend)
+
+
+@dataclass(frozen=True, slots=True)
+class WithinDemeaner(BaseDemeaner):
+    """Krylov-based demeaner configuration for the Rust `within` backend."""
+
+    krylov_method: str = "cg"
```
leostimpfle, line 70: Also use a `typing.Literal`, e.g., ` Literal["cg", "gmres"]`?

pyfixest/demeaners.py
```diff
@@ -0,0 +1,170 @@
...
+            and self.krylov_method != "gmres"
+        ):
+            raise ValueError("Multiplicative Schwarz requires `krylov_method='gmres'`.")
+
+
+@dataclass(frozen=True, slots=True)
+class LsmrDemeaner(BaseDemeaner):
+    """Sparse LSMR demeaner for CPU and GPU backends."""
+
+    precision: str = "float64"
```
leostimpfle, line 119: Use ` Literal["float32",  "float64"]`?

pyfixest/core/demean.py
```diff
@@ -1,7 +1,220 @@
...
+def _compute_factor_cardinalities(flist: NDArray[np.uint32]) -> tuple[int, ...]:
+    # FE columns are dense 0..K-1 codes from `pd.factorize`, so cardinality is max + 1.
+    return tuple((flist.max(axis=0) + 1).tolist())
+
+
+def _sanitize_krylov_and_preconditioner(
+    krylov_method: str,
+    preconditioner_type: str,
+) -> tuple[str, str]:
+    krylov_method = krylov_method.lower()
```
leostimpfle, line 39: Do we need this kind of string parsing or should we tighten the arguments to a `Literal` (see other comments)?

pyfixest/core/_core_impl.pyi
```diff
@@ -20,7 +22,22 @@ def _demean_within_rs(
     weights: NDArray[np.float64],
     tol: float = 1e-06,
     maxiter: int = 1_000,
+    krylov_method: str = "cg",
+    gmres_restart: int = 30,
+    preconditioner_type: str = "additive",
```
leostimpfle, line 27: Tighten to `Literal["additive", "multiplicative"]`?

pyfixest/demeaners.py
```diff
@@ -0,0 +1,170 @@
...
+        object.__setattr__(
+            self,
+            "fixef_tol",
+            _validate_positive_float(self.fixef_tol, "fixef_tol"),
+        )
+        object.__setattr__(
+            self,
+            "fixef_maxiter",
+            _validate_positive_int(self.fixef_maxiter, "fixef_maxiter"),
+        )
```
leostimpfle, line 46: Why not just call `_validate_positive_float` and `_validate_positive_int` instead of reassigning the attributes?

pyfixest/demeaners.py
```diff
@@ -0,0 +1,170 @@
...
+    kind: ClassVar[str] = "map"
+
+    def __post_init__(self) -> None:
+        BaseDemeaner.__post_init__(self)
+        if not isinstance(self.backend, str):
+            raise TypeError("`backend` must be a string.")
+        backend = self.backend.lower()
+        if backend not in {"numba", "rust", "jax"}:
+            raise ValueError("`backend` must be one of 'numba', 'rust', or 'jax'.")
+        object.__setattr__(self, "backend", backend)
```
leostimpfle, line 63: We can avoid the reassign when tightening `backend` to be case sensitive (consistent with the type hint)?

pyfixest/estimation/internals/demeaner_options.py
```diff
@@ -0,0 +1,117 @@
...
+from pyfixest.demeaners import (
+    AnyDemeaner,
+    BaseDemeaner,
+    LsmrDemeaner,
+    MapDemeaner,
+    WithinDemeaner,
+)
+from pyfixest.estimation.internals.literals import DemeanerBackendOptions
+
+ResolvedDemeaner: TypeAlias = AnyDemeaner
```
leostimpfle, line 14: Why is this introduced? Couldn't we just type hint with `AnyDemeaner`?

pyfixest/estimation/internals/demean_.py
```diff
@@ -163,6 +183,135 @@ def demean_model(
...
+    *,
+    tol: float | None = None,
+) -> ResolvedDemeaner:
+    """Override FE tolerance on a typed demeaner when needed. Used for IWLS acceleration."""
+    if tol is None or tol == demeaner.fixef_tol:
+        return demeaner
+    return replace(demeaner, fixef_tol=tol)
+
+
+def dispatch_demean(
```
leostimpfle, line 230: How about giving `demeaner` a method `demean` that avoids this dispatch function?

leostimpfle (commented):
This looks very nice! After an initial look (mostly at the design), I have three (fairly minor) comments:

- I would favour more consistent use of `typing.Literal` to avoid passing `str` arguments (e.g., `preconditioner_type`)
- Many of the `object.__setattr__` inside the frozen dataclasses can probably be avoided so that `__post_init__` just calls validations (but does not modify provided arguments)
- I'm wondering if adding a `demean` method to `BaseDemeaner` would be nicer than the explicit `dispatch_demean`

s3alfisc:
Continuing this PR in three sub PRs - first introduce the typed Demeaner interface in #1282 (already merged), then adding the gmres / multiplicative schwarz options in #1285, then adding caching, then using caching in multiple estimation. Just too many things happened in this PR for a review. @leostimpfle thanks for taking a look, I incorporated some (but not all) of your feedback in #1282!

PR #1275, opened 2026-04-13
Add weighting bootstrap inference (weightingboottest)

### LINKED ISSUE #1274: Add weighting bootstrap inference (weightingboottest)
Feature request
Add a weightingboottest() method to Feols that implements observation-reweighting bootstrap inference, as a complement to the existing wildboottest() and analytic vcov() options.

Motivation
The classical pairs bootstrap (sampling with replacement) is equivalent to weighting observations by counts drawn from a Multinomial(N, 1/N, ..., 1/N) distribution. The Bayesian bootstrap (Rubin 1981) is its continuous analogue using Dirichlet(α) weights, which assigns positive mass to every observation and avoids the discreteness of the multinomial.

Proposed API

fit = pf.feols("Y ~ X1 | f1", data)
fit.weightingboottest(reps=500, method="bayesian")   # Dirichlet α=1
fit.weightingboottest(reps=500, method="multinomial") # = pairs bootstrap

Notes

- Works with FE models (re-demeans per iteration with bootstrap weights)
- ~12× faster than wildboottest() on large panels (uses native numba demeaning)
- multinomial is mathematically equivalent to the classical pairs/XY bootstrap

mdizadi:
## Summary

Closes #1274

The classical pairs bootstrap (sampling with replacement) is equivalent to weighting observations by counts drawn from a Multinomial(N, 1/N, ..., 1/N) distribution. The Bayesian bootstrap (Rubin 1981) is its continuous analogue: weights are drawn from a Dirichlet(α) distribution, which assigns positive mass to every observation and avoids the discreteness of the multinomial. Both are implemented as `weightingboottest()`, a new observation-reweighting inference method alongside the existing `wildboottest()` and analytic `vcov()` options.

This also adds the **first pairs/XY bootstrap** implementation in pyfixest — `method="multinomial"` is provably equivalent to resampling rows with replacement.

### API

```python
fit = pf.feols("Y ~ X1 + X2 | f1", data)

# Bayesian bootstrap (Rubin 1981) — Dirichlet(α=1), continuous weights
fit.weightingboottest(reps=500, method="bayesian")

# Multinomial bootstrap — equivalent to classical pairs/XY bootstrap
fit.weightingboottest(reps=500, method="multinomial")

# With clustering (panel-aware: unit weights broadcast to observations)
fit.weightingboottest(reps=500, method="bayesian", cluster="unit_id")

# Return full draw matrix for custom inference
res, draws = fit.weightingboottest(reps=500, return_draws=True)
```

### Implementation notes

- Re-demeans per bootstrap iteration with the bootstrap weights using pyfixest's native numba demeaning kernel — no one-hot FE encoding, no external package dependency
- ~12× faster than `wildboottest()` on large FE panels (500 FEs, N=5,000)
- Handles unbalanced panels correctly via `np.unique` inverse mapping
- Filters zero-weight rows before demeaning to prevent division-by-zero in the numba kernel (arises when multinomial draws exclude an entire FE group)
- Stores `_X_untransformed_df` (pre-demeaning DataFrame) to enable correct per-iteration FE re-absorption, mirroring the existing `_Y_untransformed`
- `_X_untransformed_df` is freed in lean mode via `_clear_attributes()`

## Test plan

- [x] Returns correct DataFrame shape and columns (no FE, single FE, two-way FE)
- [x] Bootstrap mean close to point estimate
- [x] `multinomial` SE matches manual pairs bootstrap SE within MC noise
- [x] Bootstrap SE (α=1) close to analytic iid SE
- [x] Invalid `method` and `alpha` raise `ValueError`
- [x] Seed reproducibility
- [x] `cluster=` argument works

Co-developed by @mdizadi and Claude Code (Anthropic).


pyfixest/estimation/models/feols_.py
```diff
@@ -1462,6 +1466,210 @@ def wildboottest(
         else:
             return res_df
 
+    def weightingboottest(
+        self,
+        reps: int,
+        method: Literal["bayesian", "multinomial"] = "bayesian",
+        alpha: float = 1.0,
```
s3alfisc: Maybe we could name the alpha arg differently? Right now I believe we might risk that users confuse alpha from the dirchlet list with the significance alpha level?

pyfixest/estimation/models/feols_.py
```diff
@@ -1462,6 +1466,210 @@ def wildboottest(
         else:
             return res_df
 
+    def weightingboottest(
+        self,
+        reps: int,
+        method: Literal["bayesian", "multinomial"] = "bayesian",
```
s3alfisc: Should we name the method argument `weights_type` for consistency with the wildboottest method?

pyfixest/estimation/models/feols_.py
```diff
@@ -1462,6 +1466,210 @@ def wildboottest(
...
+        if method not in ("bayesian", "multinomial"):
+            raise ValueError(
+                f"method must be 'bayesian' or 'multinomial', got '{method}'."
+            )
+        if method == "bayesian" and alpha <= 0:
+            raise ValueError(f"alpha must be positive, got {alpha}.")
+
+        rng = np.random.default_rng(seed)
+
+        # --- resolve cluster variable (same logic as wildboottest) ---
```
s3alfisc: If same logic as in wildboottest, can we put this section in a utility function?

pyfixest/estimation/models/feols_.py
```diff
@@ -1462,6 +1466,210 @@ def wildboottest(
...
+                _X_b = Xd.to_numpy()
+            else:
+                _Y_b = Y_f.to_numpy().flatten()
+                _X_b = X_f.to_numpy()
+
+            # apply combined weights via WLS transform: X* = sqrt(w)*X
+            sqrt_w = np.sqrt(w_f)
+            Xw = _X_b * sqrt_w[:, None]
+            Yw = _Y_b * sqrt_w
+            beta_boots[b] = solve_ols(Xw.T @ Xw, Xw.T @ Yw, self._solver)
```
s3alfisc: Currently, the method only works for OLS - if we keep it as is for now, we should add restrictions to the IV / GLM / other estimators. 
An alternative could be to simply loop over `pf.feols()`, `pf.feglm()` etc, applying different weights. This way, the method(s) would generalize to all GLM / OLS / IV classes supported in pyfixest.

pyfixest/estimation/models/feols_.py
mdizadi: Thanks for the comment. It is now changed to concentration.
https://en.wikipedia.org/wiki/Concentration_parameter

mdizadi:
pre-commit.ci please re-run

PR #1286, opened 2026-04-25
Close #1283

### LINKED ISSUE #1283: Add `split` to `tidy`
> This takes care of etable. It's still very clunky to just get a straight-forward `tidy()`-style pandas DataFrame of results out of a `FixestMulti` object. Ideally I'd just want the standard output of tidy(), with the level of the `split` as an additional column. 

 _Originally posted by @RoyalTS in [#993](https://github.com/py-econometrics/pyfixest/issues/993#issuecomment-4294462878)_

https://github.com/py-econometrics/pyfixest/blob/c533be263b6092e58139d66376785d7d95043150/pyfixest/estimation/FixestMulti_.py#L531-L549

Sample splits are implicit in the `fml` level of the `tidy` results object. I agree that it would be nice to be able to filter directly on the split variable. @s3alfisc What do you think? It should be fine to add a column to the results `pandas.DataFrame`, no?

leostimpfle:
This adds a new column `Sample` to the tidy results frame but only if `fsplit/split` was used. Note that I have decided to not add it as a `MultiIndex` level because I expect that we will sooner or later move away from `pandas` and alternatives (`polars`/`duckdb`) don't really use indices anyway. Closes #1283.

### DIFF
--- a/pyfixest/estimation/models/_result_accessor_mixin.py
+++ b/pyfixest/estimation/models/_result_accessor_mixin.py
@@ -170,20 +170,22 @@ class ResultAccessorMixin:
                 UserWarning,
             )
 
-        tidy_df = pd.DataFrame(
-            {
-                "Coefficient": self._coefnames,
-                "Estimate": self._beta_hat,
-                "Std. Error": self._se,
-                "t value": self._tstat,
-                "Pr(>|t|)": self._pvalue,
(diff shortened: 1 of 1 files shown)


s3alfisc (approved):

s3alfisc:
Thanks @leostimpfle!

Issue #500, opened 2024-06-10
Inference: Implement `Conley` Standard Errors


s3alfisc:
## Context 

For spatial regression, `Conley` standard errors are quite popular. 

## To do

- Implement Conley Standard Errors in `numba`.
- You can use @kylebutts Rcpp code for reference: [link](https://gist.github.com/kylebutts/c289044c3233f9f0c3edfd14bee164e4)
- See [this paper ](https://www.princeton.edu/~umueller/SpatialRegression.pdf)for background reading. 
- Implement Unit Tests against `fixest`'s implementation of Conley SEs.

PR #1318, opened 2026-05-27
feat: add Conley standard errors

damandhaliwal:
Hi @s3alfisc,

I'm back after finishing my Masters. Decided to pick off an easy PR, addresses #500. This was way easier than the NW/DK HAC SEs given how good Codex/CC have gotten, but I have limited experience with Rust so let me know if there are any big issues and I can address them.

Thanks,
Daman

### DIFF
--- a/docs/quickstart.qmd
+++ b/docs/quickstart.qmd
@@ -222,7 +222,7 @@ pf.feols(
 
 # Standard Errors and Inference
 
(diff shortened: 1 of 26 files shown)


s3alfisc:
Wow, looks fantastic! I'm currently knee deep in writing up the new demeaner algo from within, and then I'll go back to reviewing PRs. Will try to give feedback as fast as possible (might even try last in first out as organizing principle 🙈). Congrats on finishing the masters!

Issue #1204, opened 2026-02-24
Weights for Gelbach Decomposition


s3alfisc:
The Gelbach Decomposition currently does not support WLS. 

It would be great if we could support 

- frequency weights
- analytical weights
- and potentially even compression strategies via sufficient statistics

s3alfisc:
Weights are implemented (as analytical weights): https://github.com/py-econometrics/pyfixest/blob/0c24f79ef08d9c1213f6aa7a286868f07df4d0e6/pyfixest/estimation/post_estimation/decomposition.py#L285

Issue #1203, opened 2026-02-24
Gelbach Decomposition: Analytical Standard Errors


s3alfisc:
Currently, we rely on the bootstrap, which can be very slow for larger problems. 

See the derivation of analytical standard errors in the paper: [link](https://ideas.repec.org/a/ucp/jlabec/doi10.1086-683668.html)

PR #1363, opened 2026-06-23
Gelbach analytical standard errors

damandhaliwal:
Addresses #1203. I might just address #1204 in the same branch so marking this as draft for now


s3alfisc:
Nice one, much appreciated!

damandhaliwal:
> and potentially even compression strategies via sufficient statistics

@s3alfisc any resource you recommend for this (from #1204)? Or any implementation that you'd like me to follow? If not, then that's okay too and I can come up with something

s3alfisc:
Hi @damandhaliwal , can you please share the Stata results as a csv / in any other file format that is Python readbable? In addition, would it be difficult to also add support for iid and crv inference?

damandhaliwal:
Hi @s3alfisc,

Just pushed the stata log to the branch as well. There is a minor difference in standard errors that I need to investigate (codex summary below):
> The Gelbach point estimates match Stata’s b1x2 results to machine precision for both analytical and frequency weights. The robust standard errors differ slightly—PyFixest is approximately 0.07%–0.36% larger for the direct and decomposed effects—because PyFixest applies one full-model HC1 correction, whereas b1x2 mixes different corrections across its restricted, full, and auxiliary covariance blocks. The full-effect standard error matches, but confidence intervals also differ because PyFixest uses t critical values while b1x2 reports z-based intervals.


It makes sense to me, but I'll update some tests and read the Stata documentation again to understand their design choice. I can also add iid and crv inference in this PR. Let me get started on it

s3alfisc:
Thanks @damandhaliwal!

damandhaliwal:
@s3alfisc the estimates match pre-SSC with Stata. I pushed that too

s3alfisc:
I'll look into this in the next days (still on vacation so I have some time, but currently working through a larger internal refactor). Thanks @damandhaliwal!

s3alfisc:
Btw, CI might fail, it's got something to do with MPS tests I activated on the main branch.

damandhaliwal:
Please review this after your vacation! Thanks a lot Alex.

PR #1364, opened 2026-06-25
Guard against zero observations or fixed effects

### LINKED ISSUE #1351: Panic at `src/demean.rs`
The rust demeaner backend can panic when the input data is degenerate which causes the Python process to be killed entirely. It would be great if this got handled more gracefully so that the Python session survives. (I often run code in an interactive console which gets completely terminated by the rust panic, so any data processing would have to be rerun.)

For example:
```python
import numpy as np
import pandas as pd
import pyfixest

df = pd.DataFrame({
    "y":  [np.nan, np.nan, np.nan],   # all rows get dropped (non-finite outcome)
    "x":  [1.0, 2.0, 3.0],
    "fe": [1, 1, 2],
})

pyfixest.feols("y ~ x | fe", data=df)
```

> thread '<unnamed>' (6539362) panicked at src/demean.rs:71:48:
> called `Option::unwrap()` on a `None` value
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> Process finished with exit code 134 (interrupted by signal 6:SIGABRT)

leostimpfle:
Adds guard against zero observations or fixed effects to `demean_impl`. Closes #1351.

### DIFF
--- a/src/demean.rs
+++ b/src/demean.rs
@@ -68,7 +68,16 @@ fn demean_impl(
 ) -> (Array2<f64>, bool) {
     let (n_samples, n_features) = x.dim();
     let n_factors = flist.ncols();
-    let n_groups = flist.iter().cloned().max().unwrap() + 1;
+    // No observations or no fixed-effects: nothing to demean. Return the input unchanged
+    if n_samples == 0 || n_factors == 0 {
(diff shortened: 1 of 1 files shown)


s3alfisc:
Thanks @leostimpfle !

PR #1366, opened 2026-06-28
Add wildboottest(): Wild Cluster Bootstrap inference (Cameron, Gelbach & Miller 2008)

hass-nation:
## Summary

Adds `wildboottest()` — Wild Cluster Bootstrap (WCB) inference for clustered regression, the gold standard for reliable p-values when the number of clusters is small (< 50).

Standard cluster-robust standard errors can be severely undersized with few clusters. The WCB imposes the null hypothesis, re-samples at the cluster level using Rademacher or Mammen weights, and reports a bootstrap p-value that remains valid with as few as 5–10 clusters.

## Algorithm (impose-null approach, Cameron, Gelbach & Miller 2008)

1. Fit the unrestricted model; record t_obs for coefficient j.
2. Fit the restricted model (drop variable j); compute restricted residuals ẽ.
3. For b = 1, …, B: draw G cluster-level weights wg; construct y*_b = X_restricted·β̂_r + Σg wg·ẽg; refit full model → t*_b.
4. p-value = mean(|t*_b| ≥ |t_obs|). Fully vectorised — no Python loop over B.

## Interface

```python
from pyfixest.inference.wildboottest import wildboottest, wildboottest_numpy

# Low-level numpy interface (works without pyfixest internals):
result = wildboottest_numpy(Y, X, cluster_ids, param_idx=1, B=999, seed=42)
# result: {'param_idx': 1, 't_stat': ..., 'p_value': ..., 'B': 999, 'weights_type': 'rademacher'}

# High-level interface (wraps a fitted pyfixest model):
result = wildboottest(model, param='x1', B=999, weights_type='mammen')
```

## New files

| File | Description |
|---|---|
| `pyfixest/inference/wildboottest.py` | `wildboottest` + `wildboottest_numpy` |
| `pyfixest/inference/__init__.py` | package init |
| `tests/test_wildboottest.py` | 22 tests |

## Tests

22/22 pass. Coverage: output keys, p-value in [0,1], null-true / null-false scenarios, Rademacher and Mammen weights, reproducibility via seed, exact B iterations, cluster-level resampling correctness.

## Reference

Cameron, A. C., Gelbach, J. B., & Miller, D. L. (2008). Bootstrap-based improvements for inference with clustered errors. *Review of Economics and Statistics*, 90(3), 414–427.

PR #1368, opened 2026-07-02
Raise an informative error in did2s when first stage fixed effects cannot be estimated

lazizbekravshanov:
## Why

`pf.did2s()` (and `event_study(estimator="did2s")`) crashes with an opaque `ValueError: operands could not be broadcast together with shapes (190,) (200,) ...` deep inside `_did2s_vcov` whenever the first stage cannot estimate some fixed effect levels. As @marcandre259 traced in #1244, the trigger is easy to hit on ordinary DiD data: the first stage is fit on **not-yet-treated observations only**, so e.g. an *always-treated unit* never gets its fixed effect estimated.

## Root cause

- `_did2s_estimate` fits the first stage on the not-yet-treated subset, then predicts on the full data: rows whose fixed-effect level was never seen predict to `NaN`.
- The `NaN`s flow into the second-stage dependent variable, where `feols` silently drops those rows — so `_second_u` comes out shorter than `_first_u` and every full-length array built from `data` in `_did2s_vcov`.
- The mismatch only surfaces later, at `second_u *= weights_array`, as a broadcasting error with no hint of the actual cause.

## What this PR does

- After the first-stage `predict`, check for `NaN` predictions; if any, resolve which fixed-effect levels are present in the data but absent from `fit1.fixef()` and raise a `ValueError` naming them:

  > The following fixed effect levels could not be estimated in the first stage regression, which is fit on not-yet-treated observations only: unit: ['1']. This happens when a level has no not-yet-treated observations (e.g. an always-treated unit) or is collinear in the first stage. First stage predictions for the affected observations are NaN, hence did2s cannot proceed. Please drop the affected observations before calling did2s().

- Adds a regression test (`tests/test_errors.py::test_did2s_unestimated_first_stage_fixef`) with a synthetic panel containing an always-treated unit — it reproduces the broadcast crash on `master` and asserts the informative error with this change.
- Adds a changelog entry under 0.70.0.

## Design note

The reporter suggested an early, informative error, so that is what this implements. If you'd rather handle it gracefully (drop the affected observations with a warning and keep the downstream arrays consistent), I'm happy to rework it that way — it just touches more of `_did2s_vcov`.

## Evidence

- New regression test fails on `master` (broadcast `ValueError`) and passes with the fix.
- `pixi run test-py`: 686 passed, 11 skipped (the 2 collection errors in `test_batched_lsmr.py`/`test_lsmr_compiled.py` are a missing optional `torch` in the default env, unrelated to this change; same on `master`).
- `pixi run lint`: all hooks pass except 2 pre-existing `mypy` errors in `separation.py`/`ritest.py`, untouched by this diff.
- Note on the red pre-commit.ci check: it fails on those same 2 pre-existing `mypy` errors. The hook's `additional_dependencies: [numpy>=1.20, ...]` floats to the newest numpy, whose stricter type stubs introduced them repo-wide (#1366 hit the same failure; #1367 passed earlier on what looks like a cached hook env). `mypy` only checks `pyfixest/`, and this branch differs from `master` there only in `did2s.py`. Happy to send a separate PR fixing the two errors or pinning numpy in the hook.

Closes #1244

### DIFF
--- a/docs/changelog.qmd
+++ b/docs/changelog.qmd
@@ -39,6 +39,14 @@ As previously announced, we removed:
 - Deprecated the SciPy and CuPy demeaner backends.
 - Deprecated `FeolsCompressed`.
 
+### Bug Fixes
+
+- `did2s()` now raises an informative error when some fixed effect levels cannot be
+  estimated in the first stage regression (e.g. for always-treated units, which have
+  no not-yet-treated observations), instead of failing later with an opaque
+  broadcasting `ValueError` in the vcov computation
+  ([#1244](https://github.com/py-econometrics/pyfixest/issues/1244)).
+
 
 ## PyFixest 0.60.0
 
--- a/pyfixest/did/did2s.py
+++ b/pyfixest/did/did2s.py
@@ -225,33 +225,17 @@ def _did2s_estimate(
     )  # iid as it might be faster than CRV
 
     # obtain estimated fixed effects
-    fixef_dict = fit1.fixef()
+    fit1.fixef()
 
     # demean data
     Y_hat = fit1.predict(newdata=data)
     if np.isnan(Y_hat).any():
         # levels present in the full data but absent from the first stage fit
-        # (fit on not-yet-treated observations only) predict to NaN and would
-        # silently be dropped in the second stage, see #1244
-        missing_by_var = []
-        fixef_vars = [x.strip() for x in _first_stage.split("|")[1].split("+")]
-        for fixef_var in fixef_vars:
-            if fixef_var not in data.columns:
-                continue
-            estimated_levels = set(fixef_dict.get(f"C({fixef_var})", {}))
-            data_levels = set(data[fixef_var].astype(str))
-            missing = sorted(data_levels - estimated_levels)
-            if missing:
-                missing_by_var.append(f"{fixef_var}: {missing}")
-        details = "; ".join(missing_by_var) if missing_by_var else "unknown levels"
+        # (fit on not-yet-treated observations only) predict to NaN
         raise ValueError(
-            "The following fixed effect levels could not be estimated in the "
-            f"first stage regression, which is fit on not-yet-treated observations "
-            f"only: {details}. This happens when a level has no not-yet-treated "
-            "observations (e.g. an always-treated unit) or is collinear in the "
-            "first stage. First stage predictions for the affected observations "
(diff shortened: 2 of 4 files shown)


lazizbekravshanov:
The red pre-commit.ci check here is exactly the two pre-existing mypy errors described in the Evidence section — I've split the fix for those into #1369 (pre-commit.ci passes there). Once that lands, a rebase/re-run should turn this PR green.

leostimpfle:
Hi @lazizbekravshanov ! Thanks for the PR. This looks overall good to me, but I've moved the warning logic around a bit:

I've implemented warnings for unseen fixed effect levels directly in the prediction methods: https://github.com/py-econometrics/pyfixest/blob/e468160ffe5908d03e3e3abe8e0de8dfb553e3f1/pyfixest/estimation/post_estimation/prediction.py#L155-L163

This avoids having to deal with any fixed effects parsing in `did2s` where we now just check for NaNs in the prediction (the stack-trace will show exactly which levels are missing from the warnings above): https://github.com/py-econometrics/pyfixest/blob/e468160ffe5908d03e3e3abe8e0de8dfb553e3f1/pyfixest/did/did2s.py#L231-L239

@s3alfisc The new warnings in predict means that the changes in this PR go beyond `did2s` but I thought it is a sensible addition with the benefit of stripping the boilerplate in `did2s` to find the missing levels. Please let me know your thoughts, happy to revert to @lazizbekravshanov local testing in `did2s`.

leostimpfle (commented):
Simplified handling in `did2s` by adding warnings in `_apply_fixef_numpy`

lazizbekravshanov:
Thanks @leostimpfle — this is cleaner than my original. Moving the unseen-level detection into `_apply_fixef_numpy` as a `UserWarning` drops the fixed-effect parsing boilerplate I had in `did2s`, and as you note it benefits every `predict` caller instead of just this path. The `stacklevel=4` pointing the warning at the `predict` call site is a nice touch.

The combined behavior still resolves #1244 the way the issue asked: the warning names exactly which levels are missing, and the `np.isnan(Y_hat).any()` check in `did2s` halts with an actionable message instead of the opaque broadcasting `ValueError` downstream. `test_did2s_unestimated_first_stage_fixef` still passes against the new flow and CI is green across the matrix.

The one genuinely new surface is that `predict` now warns on unseen levels by default (`warn=True`) for all callers, not just `did2s` — which I think is the right default, but that's the call I'd leave to @s3alfisc. Happy either way.

PR #1370, opened 2026-07-04
feat: add DFM omnibus test for treatment effect heterogeneity (#989)

### LINKED ISSUE #989: Omnibus Test For Effect Heterogeneity
Implement the Omnibus Test of systematic treatment effect heterogeneity following Ding, Feller, Miratrix (2018). 

See here for an R implementation: [link](https://github.com/Netflix-Skunkworks/causaltransportr/blob/main/R/dfmTest.R) (published under MIT). 

Can be implemented as a post-estimation method to the Feols class.

FYI @apoorvalal

genrichez:
Implement the Ding, Feller, Miratrix (2019) test as a post-estimation method on Feols. The test checks whether treatment effects vary systematically with observed covariates by fitting separate OLS in each arm and jointly testing the slope differences with a sandwich variance.

- New standalone function in post_estimation/dfm_test.py
- New method Feols.dfm_test(treatment=...) returning pd.Series
- Tests covering size, power, edge cases, and error handling

Closes #989


s3alfisc:
Hi @genrichez , thanks for this! Two initial questions: 

- The test should only work for OLS, correct? In this case, can we please throw an error for IV, GLM, Quantreg? 
- Is the `dfm_test` the best name? It fits the `wald_test`, but would it maybe be better to give it a more revealing name? I.e. `dfm_treatment_heterogeneity_test`? Of something like it, but shorter?

genrichez:
> [Codecov](https://app.codecov.io/gh/py-econometrics/pyfixest/pull/1370?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=py-econometrics) Report

Hey @s3alfisc , thanks for the feedback! 
1. Yes ,100% , have added it 
2. Agree, it's not the best one. What do you think of dfm_het_test or dfm_heterogeneity_test?

s3alfisc:
Thank you! I think I'd prefer dfm_heterogeneity_test, even though it's slightly verbose?

s3alfisc:
Plus, could we maybe add some direct tests against the r code linked in the issue?

genrichez:
> Plus, could we maybe add some direct tests against the r code linked in the issue?

Sounds good, renamed it + added r code tests

genrichez:
Hey @s3alfisc ,saw #1411 flagged this PR for a missing quartodoc entry, added it. Also double-checked the test coverage from the refactor commit, looks solid. Let me know what you think !

PR #1385, opened 2026-07-08
Add a no-JIT test run so codecov captures numba coverage

### LINKED ISSUE #829: Add tests with `JIT=False` to add `numba`, `JAX` to coverage reports
The `codecov` reports currently ignore `numba` and `JAX` code as `codecov` does not recognize it. To add these lines (which are all covered by tests) to the coverage reports, we need to run the respective tests in non-jitted mode, which can be achieved by setting `NUMBA_DISABLE_JIT=1`. 

In #827 , we add an "extended" test suite with "heavy" unit tests that we do not want to run at every single CI run, which will run once a week. I think it would make sense to add the no-jit runs there too. 

Tests do run in no-jit mode: 

- [ ] [test_ses](https://github.com/py-econometrics/pyfixest/blob/master/tests/test_ses.py) for CRV inference
- [ ] [test_demean](https://github.com/py-econometrics/pyfixest/blob/master/tests/test_demean.py) for the demeaning algos
- [ ] [test_detect_singletons](https://github.com/py-econometrics/pyfixest/blob/master/tests/test_detect_singletons.py) for the algos that detect and rop singleton fixed effects
- [ ] [test_ritest](https://github.com/py-econometrics/pyfixest/blob/master/tests/test_ritest.py) optional, these already take forever in jitted mode

We might also consider to add a direct test for multicollinearity check, which is currently only implicitly tested in test_vs_fixest.py.

lazizbekravshanov:
## Why

codecov under-reports coverage of pyfixest's numba code. numba `@njit` functions are JIT-compiled, so their bodies never execute as traced Python bytecode — `coverage.py` records the `def`/decorator line but marks the function body as uncovered. This is the gap described in #829.

Setting `NUMBA_DISABLE_JIT=1` makes numba run the pure-Python function bodies, so coverage records them.

## What this does

- Adds a `test-py-nojit` pixi task that runs the numba-exercising tests (`test_ses`, `test_demean`, `test_collinearity`, `test_count_fixef_fully_nested`) with `NUMBA_DISABLE_JIT=1`, writing `coverage-nojit.xml`.
- Adds two steps to the weekly `extended_tests.yaml`: run `test-py-nojit`, then upload its coverage under a new `tests-nojit` codecov flag. Kept out of per-commit CI because no-jit runs are much slower.
- `.gitignore`: ignore `coverage-nojit.xml`.

## Evidence

Measured on current `master`, coverage of each live numba module — JIT on (how CI runs today) vs. off (this task):

| numba module | JIT on (today's CI) | `NUMBA_DISABLE_JIT=1` (this task) | surfaced by |
|---|---|---|---|
| `demean_nb.py` | 17% | 100% | `test_demean` |
| `find_collinear_variables_nb.py` | 12% | 88% | `test_collinearity` |
| `nested_fixef_nb.py` | low | 100% | `test_count_fixef_fully_nested` |

The full `test-py-nojit` run is green: **278 passed, 6 skipped in ~9m15s** (numba object mode gives identical results). It's slow — hence weekly-only, not per-commit. Lint (ruff, mypy, GitHub-workflow validation) is green.

## Scope notes (verified against current `master`)

- **JAX dropped.** The issue also names JAX, but JAX is no longer a dependency and there is no `jax` usage in the source (the demean-backend rewrite deprecated it) — so this is numba-only.
- **`detect_singletons` excluded.** It now delegates to the Rust core (`pyfixest.core._core_impl._detect_singletons_rs`); `NUMBA_DISABLE_JIT` doesn't affect Rust, which coverage can't see either way.
- Live numba modules this surfaces: `demean_nb` (`test_demean`), `find_collinear_variables_nb` (`test_collinearity`), `nested_fixef_nb` (`test_count_fixef_fully_nested`), and the CRV/vcov numba path (`test_ses`) — I widened the target set from the initial `test_ses`+`test_demean` so `find_collinear`/`nested_fixef` are actually exercised, not just named.

Happy to trim the target set or adjust the flag/upload wiring to your preference.

## Acceptance criteria

- [x] New behavior covered: numba-heavy tests run under `NUMBA_DISABLE_JIT=1` in the weekly extended suite, uploaded under a `tests-nojit` codecov flag
- [x] Coverage effect verified per module: `demean_nb` 17%→100%, `find_collinear_variables_nb` 12%→88%, `nested_fixef_nb`→100% with JIT disabled
- [x] Full no-jit run green (`278 passed, 6 skipped`)
- [x] Lint clean (ruff, mypy, GitHub-workflow validation)
- [x] No behavior change to library code; CI/tooling only, kept off per-commit CI (weekly)

Closes #829.

### DIFF
--- a/.github/workflows/extended_tests.yaml
+++ b/.github/workflows/extended_tests.yaml
@@ -31,3 +31,15 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: coverage.xml
           flags: tests-extended
+
+      - name: Run no-JIT tests (numba coverage)
+        run: pixi run -e py312 test-py-nojit
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: 4
+
+      - name: Upload no-JIT coverage to Codecov
+        uses: codecov/codecov-action@v7
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage-nojit.xml
+          flags: tests-nojit
--- a/.github/workflows/extended_tests.yaml
+++ b/.github/workflows/extended_tests.yaml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+  schedule:
+    - cron: "0 0 * * 0"
+  workflow_dispatch:
 
 jobs:
   test-extended:
@@ -33,11 +36,13 @@ jobs:
           flags: tests-extended
 
       - name: Run no-JIT tests (numba coverage)
+        if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
         run: pixi run -e py312 test-py-nojit
         env:
           PYTEST_XDIST_AUTO_NUM_WORKERS: 4
 
       - name: Upload no-JIT coverage to Codecov
+        if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
         uses: codecov/codecov-action@v7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ dist/
 # Unit test / coverage reports
 .coverage*
 coverage.xml
+coverage-nojit.xml
 htmlcov/
 .pytest_cache/
 
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,6 +3,12 @@ ignore:
   - "pyfixest/utils/dgps.py"
   - "pyfixest/utils/_exceptions.py"
 
+flags:
+  tests-nojit:
+    paths:
+      - "pyfixest/estimation/numba/*"
+    carryforward: true
+
 coverage:
   status:
     project:
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,8 +2,6 @@ ignore:
   - "benchmarks/**"
   - "pyfixest/utils/dgps.py"
   - "pyfixest/utils/_exceptions.py"
-  # cannot compute codecov for numba files (but all are tested)
-  - "pyfixest/estimation/numba/*"
 
 coverage:
   status:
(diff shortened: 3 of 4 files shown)


s3alfisc:
Amazing, thank you! I will review and merge after work this evening =)

lazizbekravshanov:
Hi @s3alfisc — gentle nudge on this one whenever you have a spare moment 🙂 No rush at all. All checks are green and it's still mergeable, so it should be ready to go whenever it's convenient for you. Thanks so much for your time and for the reviews!

lazizbekravshanov:
Small update, @s3alfisc: I widened the `test-py-nojit` target set (178edd5) to also run `test_collinearity` and `test_count_fixef_fully_nested`, so the weekly run now actually exercises `find_collinear_variables_nb` (12% → 88%) and `nested_fixef_nb` (→ 100%) under `NUMBA_DISABLE_JIT=1` — not just `demean_nb`. The initial cut only ran `test_ses`+`test_demean`, which left those two modules unsurfaced. Full run still green (278 passed / 6 skipped, ~9m, weekly-only). Happy to trim it back if you'd prefer a smaller set.

lazizbekravshanov:
Thanks for taking this the last mile, @s3alfisc — `b918c34d` looks great and matches the intent cleanly. Gating the no-JIT steps behind `schedule`/`workflow_dispatch` is a nicer way to keep the slow run weekly-only than what I had, and dropping the `pyfixest/estimation/numba/*` ignore in `codecov.yml` is really the other half of the story — the no-JIT run makes those bodies *execute* so coverage can see them, and un-ignoring them means that coverage now actually counts. Happy with the design as-is.

One small thing to sanity-check on your end (non-blocking): since the no-JIT steps only run on the weekly `schedule`/`workflow_dispatch`, the numba files won't be covered on ordinary push/PR runs, but they're no longer ignored either — so regular PR coverage reports may show them as uncovered until the weekly job carries coverage forward. If your Codecov carryforward handles that, ignore me; just flagging in case it surprises anyone on a normal PR. Thanks again for the reviews and for finishing it off!

s3alfisc:
Hi @lazizbekravshanov, good point! Made a small final adjustments. In general, I do not expect too much development with numba going forward, so things should be fine as they are now. Thanks for the PR!

s3alfisc:
@all-contributors please add @lazizbekravshanov for infra.

PR #1383, opened 2026-07-06
Add pyfixest agent toolkit (.claude/ commands + sub-agents)

mdizadi:
## Summary

A thin layer of Claude Code slash-commands + sub-agents over the existing
`.agents/feature-pr.md` workflow — one command per phase (`/discover`,
`/implement-slice`, `/validate-numerics`, `/review-pr`, `/rewrite-history`),
a locator/analyzer sub-agent split for parallel discovery, and plans
persisted as checkbox files in `.claude/plans/` (gitignored).

**`.agents/feature-pr.md` itself is untouched** — it's the reference file,
and this PR doesn't presume to edit it. Six small behaviors the commands
rely on that go beyond what `feature-pr.md` currently documents (plan
persistence, interaction rules, a plan-mismatch stop, a staleness check, a
Phase 3 completion check, and a named escalation-to-replan tier) are
implemented at the command level for now, and written up as proposed diffs
with rationale in `.claude/README.md` → "Proposed changes to
.agents/feature-pr.md" — happy to fold any/none of these into the reference
file, whatever you'd prefer.

## Test plan

- [x] Dry run (`/discover` + `/review-pr`, no changes) on PR #1275
- [x] Full run (`/discover` → `/implement-slice` × 4 → `/review-pr` × 2) on
      PR #1275, including a mid-flight rebase onto a large upstream GLM
      refactor — the staleness-check and Phase-3-check proposals came
      directly out of gaps that run exposed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1373, opened 2026-07-04
feat: add Conley spatial HAC standard errors

genrichez:
Closes #500.
This adds Conley (1999 ) spatial HAC standard errors as a new vcov type in pyfixest. Users can compute standard errors that account for spatial correlation between observations within a given distance.
**Usage :** 
```
import pyfixest as pf

fit = pf.feols("y ~ x1 + x2", data=df)
fit.vcov({"conley": {"cutoff": 100, "lat": "latitude", "lon": "longitude"}})
fit.summary()
```
**What it does:**
The implementation computes the Conley spatial HAC meat matrix by weighting the outer product of score vectors (X_i * e_i) by a distance-based kernel. Pairs of observations within the cutoff distance (in km, using Haversine) contribute to the spatial correlation estimate.
Two kernels are supported:

- bartlett (default): linear decay from 1 at distance 0 to 0 at the cutoff

- uniform: equal weight for all pairs within the cutoff

**Implementation:**

- New module pyfixest/estimation/post_estimation/conley_se.py with the spatial kernel computation

- Uses numba for acceleration when available, falls back to pure numpy

- Wired into the existing vcov() method on Feols as a new branch alongside iid, hetero, CRV, HAC
**Tests:** 15 tests in tests/test_conley_se.py covering Haversine distance, meat matrix properties, brute-force validation, size control under IID errors, and detection of spatial correlation.

PR #1410, opened 2026-07-25
Consolidate agent instructions and add house-style checks

s3alfisc:
Consolidates the unmerged `agent-style-guide` commits, the two evidence-backed
proposals from #1383, and an audit of #1373 / #1370 / #1363 / #1275.

Two problems the audit found:

- `AGENTS.md`'s "mirror the nearest implementation" beat its own recipes,
  because the codebase doesn't follow them — `Literal` aliases live in five
  places, so three of four PRs skipped `literals.py`. Fixed with a
  known-stale-precedents table.
- Rules were stated but not checkable. All four PRs missed their quartodoc
  entry. `scripts/check_house_style.py` now holds the mechanical rules
  (pre-commit hook + `pixi run check-style`), and the prose it replaced is gone.

Also adds four rules the audit found uncovered (no override hooks on model
classes, a test through the public API, dataclass fields, stale precedents) and
#1383's staleness check and re-plan tier.

Net: 769 lines of prose instead of ~1232 if combined as written. `AGENTS.md`
now states the ratchet — a checkable rule belongs in a check, and adding a rule
means deleting one.

Test plan:

- [x] `pixi run lint` green, including the new `house-style` hook
- [x] `pixi run test-py` — 732 passed, 0 failed
- [x] Diff checks verified against #1370 (deleted test, missing quartodoc entry)
      and #1373 (numba, importlib, test classes)

Supersedes #1383 and `claude/pyfixest-agent-style-guide-6cd554`.

### DIFF
--- a/.agents/align-pr.md
+++ b/.agents/align-pr.md
@@ -1,0 +1,157 @@
+# Workflow: align a contributor PR with house style
+
+Audit an existing PR against `AGENTS.md` and produce either a paste-ready
+review or a fixed-up local branch, so the maintainer does as little as
+possible. `.agents/feature-pr.md` is for *building* a change; this workflow is
+for *auditing and repairing* one that already exists. Read `AGENTS.md` first —
+it is the rulebook; this file is only the loop that applies it.
+
+Input: a PR number (given several, run the whole workflow per PR
+independently; never mix branches). Two modes, chosen per invocation:
+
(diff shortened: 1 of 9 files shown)


s3alfisc:
Closed automatically by a branch rename; continued in #1411 with the same commits.

PR #1421, opened 2026-07-26
Add regression tests for four formula-parsing bugs

s3alfisc:
Detected several minor bugs / regressions while reviewing this PR. 

### What's here

Add targeted snapshot tests against a released pyfixest version, including the bugs / regressions. 
These pass the CI here but are explicitly marked. Proper fixes will be added in follow up PRs. 

### The four bugs

| # | Symptom | Fixed in |
|---|---|---|
| 1 | `Y ~ (X1 + X2)^2` expands to `X1:2 + X2:2` and loses the interaction | PR 2 |
| 2 | `[log(X2) ~ Z1]` raises `FormulaicCompatibilityError` | PR 3 |
| 3 | `I(Y + Y2) ~ X1` is read as two outcomes and rejected | PR 4 |
| 4 | `C(np.floor(X2))` predicts all `NaN` on `newdata` | PR 5 |

PR #1422, opened 2026-07-26
Scope the fixest `^` interaction to the fixed effects segment

s3alfisc:
Stacked on #1421 — merge that first.

### Problem

`_FixedEffectsOperatorResolver` replaces formulaic's `^` operator for the
**whole** formula, but the fixest meaning of `^` (interact these fixed effects)
is only wanted inside the `|` segment. Formulaic documents `^` as an alias for
`**`, so everywhere else it should expand a power:

```python
pf.feols("Y ~ (X1 + X2)^2", data)
# expected: Intercept, X1, X2, X1:X2
# on this branch: Intercept, X1:2, X2:2   <- interaction silently lost
```

### Fix

Rewrite `^` to `:` during preprocessing, in the fixed effects segment only
(`_preprocess_fixest_fixed_effect_interactions`), and drop the custom operator
resolver so formulaic's own `^` survives everywhere else. `_str_split_by_sep`
now tracks brackets as well as parentheses so the split is safe around IV
blocks.

A `^` nested inside parentheses in the FE segment (`| I(x^2)`) is left alone.

### Verification

- `| f1^f2` still demeans on the interacted group — checked numerically against
  a manually built `f1_f2` column (coefficients and standard errors).
- `^` and `**` now produce identical expansions.
- Removes the `XFAIL_CARET` markers and the `ols-caret-power` snapshot marker
  added in #1421.

### DIFF
--- a/pyfixest/estimation/formula/parse.py
+++ b/pyfixest/estimation/formula/parse.py
@@ -22,9 +22,6 @@ from pyfixest.estimation.formula.formulaic_compat import (
     get_first_multistage_rhs,
     is_structured_formula,
 )
-from pyfixest.estimation.formula.transforms.fixed_effects_encoding import (
-    _FixedEffectsOperatorResolver,
-)
 from pyfixest.estimation.formula.utils import (
     _MULTIPLE_ESTIMATION_PATTERN,
     _get_position_of_first_parenthesis_pair,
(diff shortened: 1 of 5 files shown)


leostimpfle (commented):
I'd prefer either keeping the custom operator resolver (reserving `^` for fixed effect interactions) or changing the syntax to `:` entirely. Aesthetically, the latter is my preferred option because it more naturally reads like "interaction"

Although both changes are breaking, the internal reparsing creates maintenance burden and means we have to keep track of the translation downstream (e.g. in `predict`) which I would really like to avoid.

leostimpfle:
Closing this in favour of #1437

PR #1423, opened 2026-07-26
Derive formulaic `_hat` names from endogenous terms

s3alfisc:
Stacked on #1422 — merge that first.

### Problem

formulaic's MULTISTAGE syntax names the generated second-stage term after the
endogenous **term**, not its source variables
([`parser.py`](https://github.com/matthewwardrop/formulaic/blob/v1.2.1/formulaic/parser/parser.py#L354-L364) builds `Factor(str(t) + "_hat")`
from the term). So `[log(X2) ~ Z1]` generates `log(X2)_hat`.

`filter_multistage_endogenous_terms` derived the name from
`required_variables` and therefore looked for `X2_hat`, which is never present:

```python
pf.feols("Y ~ X1 | log(X2) ~ Z1", data)
# FormulaicCompatibilityError: expected generated second-stage terms
# ['X2_hat'] to be present before filtering
```

Any transformed endogenous variable hit this — the guard fired on a formula
that used to work.

### Fix

Pass the endogenous *terms* rather than their source variables, so the expected
name is built the same way formulaic builds it. The comment now records the
`log(X2)` vs `X2` distinction so it doesn't regress.

### Verification

Inline `[np.exp(X2) ~ Z1]` matches the same model with the transform
precomputed as a data column, on both coefficients and standard errors.
Removes the `XFAIL_TRANSFORMED_ENDOGENOUS` markers from #1421.

### DIFF
--- a/pyfixest/estimation/formula/formulaic_compat.py
+++ b/pyfixest/estimation/formula/formulaic_compat.py
@@ -69,12 +69,13 @@ def _get_single_multistage_block(rhs: formulaic.formula.Formula) -> Any:
 
 def filter_multistage_endogenous_terms(
     exogenous: formulaic.formula.Formula,
-    endogenous_variables: Iterable[Any],
+    endogenous_terms: Iterable[Any],
 ) -> formulaic.formula.SimpleFormula:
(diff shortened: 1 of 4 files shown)


leostimpfle (approved):
Good catch!

PR #1424, opened 2026-07-26
Treat a nested `+` in the dependent as one outcome

s3alfisc:
Stacked on #1423 — merge that first.

### Problem

Two places treated a `+` anywhere on the left hand side as a separator between
multiple dependent variables:

1. `_preprocess_fixest_multiple_dependents` tested `"+" in dependent`, so
   `I(Y + Y2)` was rewritten to a one-argument `sw(...)` and rejected.
2. `Formula.__post_init__` counted `dependent.required_variables`, which is 2
   for `I(Y + Y2)` — one expression evaluated from two columns.

```python
pf.feols("I(Y + Y2) ~ X1", data)
# FormulaSyntaxError: 'sw(...)' requires at least 2 arguments, got 1
```

### Fix

Split the left hand side on **top-level** `+` only, so a `+` nested inside a
transform stays part of the expression, and count dependent *terms* instead of
source variables.

`Y + Y2 ~ X1` still expands to two models and still emits its
`DeprecationWarning`.

### Verification

`I(Y + Y2) ~ X1` estimates one model whose coefficients match the same
regression on a precomputed `Y + Y2` column. Removes the
`XFAIL_DEPENDENT_PLUS` markers and the `ols-identity-sum` snapshot marker from
#1421.

### DIFF
--- a/pyfixest/estimation/formula/parse.py
+++ b/pyfixest/estimation/formula/parse.py
@@ -81,10 +81,12 @@ class Formula:
                 f"Received {len(self._formula.rhs)}:\n"
                 f"{self._formula}"
             )
-        if len(self.dependent.required_variables) > 1:
+        # Count terms, not source variables: `I(Y + Y2)` is one dependent
+        # expression evaluated from two columns.
+        if len(self.dependent) > 1:
             raise FormulaSyntaxError(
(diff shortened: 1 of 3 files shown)


leostimpfle (approved):
Thanks!

PR #1425, opened 2026-07-26
Detect unseen categorical levels on evaluated factors

s3alfisc:
Stacked on #1424 — merge that first.

### Problem

`_rows_with_unseen_categories` treated `ModelSpec.factor_variables` as the
values a factor encodes. It is [documented](https://github.com/matthewwardrop/formulaic/blob/v1.2.1/formulaic/model_spec.py#L306-L311)
as the variables used to *evaluate* the factor — for `C(np.floor(X2))` that is
`{X2, np.floor, C}`, while `factor_contrasts[...].levels` holds the levels of
`floor(X2)`.

The check therefore compared raw `X2` values against floor levels, found no
match, and marked every row unseen:

```python
fit = pf.feols("Y ~ C(np.floor(X2))", data)
fit.predict(newdata=data.iloc[:20])
# 20 NaN   (master: 20 finite predictions)
```

`C(f1)` worked only by luck — there the source column *is* the encoded value.
Any categorical factor built from an expression, or from more than one column,
was affected.

### Fix

Re-encode each contrast factor on its own, at full rank, from the stored
`encoder_state`. Formulaic casts an unseen level to `NaN`, which dummy-encodes
to an all-zero row, while every seen level contributes exactly one indicator —
so the row sum is the test and no expression has to be evaluated by hand. This
uses only `ModelSpec` / `get_model_matrix`, the documented reuse path.

The `i()` branch keeps its variable-based lookup: pyfixest's own transform
records the actual column name in its state, and it handles `ref=` and binning,
which the full-rank trick cannot distinguish from an unseen level.

`iter_model_spec_categorical_levels` is renamed `iter_i_categorical_levels`
since it now covers only the `i()` path.

### Verification

- `C(np.floor(X2))`, `C(np.floor(X2)):X1` and `C(f1 + f2)` predict finitely on
  `newdata` and match in-sample fits.
- A genuinely unseen evaluated level still returns `NaN`, and unseen `C()`/`i()`
  levels are unaffected.
- Removes the last markers from #1421 — the suite is back to no xfails.

### DIFF
--- a/pyfixest/estimation/formula/formulaic_compat.py
+++ b/pyfixest/estimation/formula/formulaic_compat.py
@@ -2,15 +2,17 @@
 
 from __future__ import annotations
 
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterable, Iterator, Mapping
 from typing import TYPE_CHECKING, Any
 
 import formulaic
 import formulaic.formula
+import numpy as np
 import pandas as pd
 from formulaic.parser.types import Factor
 
 from pyfixest.estimation.formula.transforms.factor_interaction import (
+    bin_mapping_state_key,
     is_contrast_state_key,
     variable_from_contrast_state_key,
 )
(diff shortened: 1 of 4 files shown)


leostimpfle (approved):
LGTM!

I really hope this will be fixed upstream at some point :)

Issue #1236, opened 2026-03-08
NA handling for predict and predict with new data


leostimpfle:
Is this behaviour of [`pyfixest.estimation.models.feols_.Feols.predict`](https://github.com/py-econometrics/pyfixest/blob/493d739932e7cd42a3bfc5479b33443967817a66/pyfixest/estimation/models/feols_.py#L1974) intentional? It seems slightly odd that we drop NaNs and singletons in the `newdata is None` case but not in the `newdata is not None` case.

```python
import numpy as np

import pyfixest as pf

data = pf.get_data(N=1_000, seed=0, model="Feols")
fit = pf.feols("Y ~ X1", data = data)

np.isnan(fit.predict(newdata=None)).any()  # False (NaNs and singletons are dropped)
np.isnan(fit.predict(newdata=data)).any()  # True (NaNs and singletons are kept)
```

s3alfisc:
I think this is matching fixest behavior? Basically, the idea is - if `new data = None`, then we predict based on the estimation sample for which we made sure to drop NaNs. If we pass `new data`, this is treated as a new data set (for out of sample prediction) and `predict` reports that it cannot work on NaNs.

lazizbekravshanov:
I looked into this and I think you are right that the core behavior is intentional and matches fixest. When newdata is None we return predictions for the estimation sample, where NaN rows and singleton fixed effect levels were already dropped at fit time, so the result has length _N with no NaNs. When newdata is passed we return one prediction per input row and set NaN for rows we cannot predict. The repo already asserts this parity: tests/test_predict_resid_fixef.py checks that len(predict()) matches R fitted.values and that len(predict(newdata)) matches R predict(newdata), and test_predict_nas compares row for row including NaNs.

The one thing that looked off is that unseen fixed effect levels already emit a warning, but a NaN in a right hand side covariate produced a NaN prediction silently. I have a small branch that keeps the returned values exactly the same and just adds a warning for those covariate NaN rows, so predict reports that it cannot work on NaNs in both cases, plus a docstring that documents the None vs newdata behavior. Does that direction work for you, or would you prefer a docstring only change with no new warning? Happy to adjust before opening the PR.

PR #1429, opened 2026-07-28
feat(predict): warn on NaN predictions from newdata covariates

lazizbekravshanov:
Feols.predict returns predictions for the estimation sample when newdata is None (NaNs and singleton fixed effect levels already dropped), but returns one prediction per row when newdata is passed. Rows with an unseen fixed effect level already warn and return NaN; rows with a NaN in a right hand side covariate returned NaN silently.

Warn for those covariate NaN rows too, mirroring the unseen level warning, so predict no longer returns NaN predictions without notice. Document the NaN and singleton handling of both branches in the predict docstring. Behavior for returned values is unchanged.

Part of #1236

### DIFF
--- a/pyfixest/estimation/models/feols_.py
+++ b/pyfixest/estimation/models/feols_.py
@@ -1759,8 +1759,18 @@ class Feols(ResultAccessorMixin):
         Predict values of the model on new data.
 
         Return a flat np.array with predicted values of the regression model.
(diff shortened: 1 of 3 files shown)


lazizbekravshanov:
Closing this for now, it jumped ahead of the discussion on #1236. If the warning approach from the investigation comment looks right to you, I am glad to reopen it.

PR #1437, opened 2026-08-01
Use : as formulaic-native fixed effect interaction

leostimpfle:
This PR replace the `fixest`-style fixed-effect interaction syntax `f1^f2` with the formulaic-native syntax `f1:f2`. A custom preprocessing step ensures backwards-compatibility for now but `^` is deprecated. See #1422 for further background details.

### DIFF
--- a/pyfixest/estimation/formula/utils.py
+++ b/pyfixest/estimation/formula/utils.py
@@ -10,16 +10,17 @@ from pyfixest.errors import FormulaSyntaxError
 def _str_split_by_sep(string: str, separator: str = "+") -> list[str]:
     """
     Split on top-level *separator*, skipping any occurrences nested inside
-    parentheses.  The main use-case is splitting formula terms on ``+``
-    without breaking apart multi-estimation operators like ``sw(a, b + c)``.
+    brackets. The main use-case is splitting formula terms on ``+`` without
+    breaking apart multi-estimation operators like ``sw(a, b + c)`` or
+    Formulaic multistage expressions like ``[x ~ z1 + z2]``.
     """
     args: list[str] = []
     depth = 0
     current: list[str] = []
     for c in string:
-        if c == "(":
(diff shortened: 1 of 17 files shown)


pyfixest/estimation/formula/utils.py
```diff
@@ -10,16 +10,17 @@
...
+    brackets. The main use-case is splitting formula terms on ``+`` without
+    breaking apart multi-estimation operators like ``sw(a, b + c)`` or
+    Formulaic multistage expressions like ``[x ~ z1 + z2]``.
     """
     args: list[str] = []
     depth = 0
     current: list[str] = []
     for c in string:
-        if c == "(":
+        if c in "([{":
```
s3alfisc, line 21: Is this change needed for the current PR, or an unrelated fix?

s3alfisc (approved):
Can you please update the changelog? Then it is good to merge.

pyfixest/estimation/formula/utils.py
leostimpfle: It's needed only for backwards-compatibility of the `f1^f2` syntax in conjunction with backwards-compatibility for the three-sided IV syntax. For example, a formula `Y ~ 1 | f1^f2 | X ~ Z` is translated to `Y ~ [X ~ Z] | f1^f2` and then to `Y ~ [X ~ Z] | f1:f2`. The addition of `[{` ensures that the string parsing doesn't split at any parenthesis/bracket.

`{` is not strictly needed for this PR, but added it because it's valid formulaic syntax, so thought it would be good to include it, too: https://matthew.wardrop.casa/formulaic/latest/guides/grammar/

leostimpfle:
> Can you please update the changelog? Then it is good to merge.

Sounds good. I will also stack your fixes in the #1421 stack (apart from #1422).

PR #1445, opened 2026-08-02
Add model matrix construction for varying slopes

leostimpfle:


leostimpfle:
Hi @s3alfisc! I'm wondering if it might make sense to merge #1444 and #1445 independently of the complete varying slope implementation in #1448 and #1449.

#1444 implements the varying slope formula parsing and #1445 the model matrix construction. Both are currently blocked in the public API so shouldn't change any user-facing behaviour (`_supports_varying_slopes` currently returns `False` for any demeaner): https://github.com/py-econometrics/pyfixest/blob/ae7bf807326d67da9ba180b7b778f12f29b0873a/pyfixest/estimation/models/feols_.py#L421-L426 

My main motivation is to reduce the rebase/merge burden against a moving master as I'm currently not sure how long it will take to finalise the full implementation. The main blocker is that I have to rework #1448 for the changes to the `within` Python API @schroedk and I have implemented in https://github.com/py-econometrics/within/issues/237 and https://github.com/py-econometrics/within/issues/236. This depends on a new `within` release but the formula parsing and model matrix construction are entirely independent of this, so it might make sense to tackle them separately.

Long story short: I'd appreciate your feedback on #1444 and #1445 irrespective of the downstream PRs in the stack 🙂

PR #1448, opened 2026-08-02
Move from within to within-py

leostimpfle:
## Summary

- replace the custom PyO3 bridge to the Rust `within` crate with the `within-py` Python API
- use the upstream `Preconditioner` directly and preserve preconditioner reuse
- remove the now-unused Rust dependency and bridge
- keep the single-factor MAP shortcut unchanged

This PR is stacked on #1445 and intentionally contains no varying-slopes wiring. That will follow in a separate PR.

## Tests

- `pixi run --frozen -e py312-r pytest tests/test_demean.py tests/test_api.py -x -q --no-cov`
- `pixi run --frozen test-py`
- ruff format/check and mypy on changed Python files
- CI-pinned Pixi 0.62.2 accepts the v6 lockfile in frozen mode


s3alfisc:
Hi @leostimpfle , back in the day we purposefully decided to only depend on the within crate directly, and not on within; I have not thought about this a lot though. It was a recommendation from @schroedk back in the day. I suppose the core motivation was that relying on the Rust crate was more robust vs pulling in another python dependency from pypi?

s3alfisc:
On the other hand, probably a good idea to make all of the wiring work via the Python API first? We can always switch back to the Rust API once everything works (if we decide this is the better option)?

leostimpfle:
My main concern was code duplication of the `within-py` API, specifically the [`Effect`](https://github.com/leostimpfle/within/blob/dafe3306c21fa9db6ffb6987740a938fc355f8c9/python/within/_within.pyi#L203) class which is the entry point to the varying slopes demeaner and will be useful to structure the `ModelMatrix` for varying slopes on the `pyfixest` side.

It seemed strange to duplicate the Python API on both sides but there may be very good reasons that I'm not aware of, so keen to get @schroedk's input. Here's what GPT-5.6 has to say re tradeoffs:

> Yes—without `within-py`, handling effects and preconditioners becomes
>   more complicated. We would need to:
> 
>   - Define a bespoke PyO3 transport format.
>   - Duplicate validation and configuration mapping.
>   - Maintain our own persistent `Preconditioner` wrapper.
> 
>   Using the upstream `within.Effect` avoids this duplication.
> 
>   The concern about “another Python dependency from PyPI” is primarily
>   about distribution, not code ownership:
> 
>   - With the direct Rust dependency, the `within` crate is compiled into
>   the pyfixest wheel, so users install one binary wheel.
>   - With `within-py`, users install a second binary wheel, which must be
>   available for every supported platform.
>   - In theory, the crates.io and PyPI releases could become
>   unsynchronized, or one artifact could be unavailable.
> 
>   However, both packages are currently version `0.3.0`, come from the same
>   [`within` repository](https://github.com/py-econometrics/within), and
>   `within-py` uses ABI3 wheels with only NumPy as a dependency. The
>   additional packaging risk therefore seems modest.
> 
>   On balance, `within-py` appears cleaner because it provides:
> 
>   - No duplicated Python-facing types.
>   - No custom effect transport protocol.
>   - Less PyO3 glue.
>   - Upstream validation and object semantics.
>   - A direct correspondence between `DemeanCache` designs and
>   `within.Effect`.

schroedk:
@leostimpfle @s3alfisc the main blocker I would see is, that currently there are no wheels for 3.1xt and PyPy builds in within. So right know pyfixest would lose support for these python versions.  I created to issues that are blocking the switch, if we want to keep parity:

https://github.com/py-econometrics/within/issues/237
https://github.com/py-econometrics/within/issues/236

leostimpfle:
This will be implemented separately once `within` `0.4.0` lands: https://github.com/py-econometrics/within/milestone/3

PR #1449, opened 2026-08-02
Wire varying slopes to within backend

### LINKED ISSUE #1290: Fixed effects in slopes (FEIS) estimation
The R fixest package allows for estimation of fixed effect in slopes (FEIS) models by absorbing effects for categories interacted with a continuous covariate that defines the slopes. This can be very useful in cases where one wants to partial out a large number of heterogeneous slopes (e.g., in my case, separate slopes per industry representing heterogeneity in production function elasticities). This is implemented using a square bracket notation, e.g., the formula `y ~ x1 | industry[x2]` absorbs industry-specific slope coefficients on `x2` and returns a coefficient on `x1`.

I'm not sure how difficult this is to implement with the existing demeaning backends, but this would be a very helpful addition to pyfixest. To my knowledge, there is no Python package right now that handles FEIS regression.

### LINKED ISSUE #1365: Varying slopes syntax support
Over in the within repo, @schroedk is implementing varying slope support for the schwarz preconditioner: [link](https://github.com/py-econometrics/within/issues/64)

For us in pyfixest, this means we need to add support for varying slope formula syntax as in fixest via `f[z]`.

@leostimpfle fyi =) Related to #1290.

leostimpfle:
Basic implementation seems to work now @s3alfisc 

```python
import pyfixest
from pyfixest.demeaners import LsmrDemeaner

data = pyfixest.get_data()
fit = pyfixest.feols("Y ~ X1 | f1[Z1]", data=data, demeaner=LsmrDemeaner())
```

### DIFF
--- a/pyfixest/estimation/formula/parse.py
+++ b/pyfixest/estimation/formula/parse.py
@@ -128,6 +128,29 @@ class FixedEffectSpecification:
         )
 
 
+def _merge_fixed_effect_specifications(
+    specifications: tuple[FixedEffectSpecification, ...],
+) -> tuple[FixedEffectSpecification, ...]:
+    """Merge specifications sharing a level term in their original order."""
+    merged: dict[Term, tuple[bool, list[Term]]] = {}
+    for specification in specifications:
+        intercept, slopes = merged.setdefault(specification.levels, (False, []))
+        slopes.extend(slope for slope in specification.slopes if slope not in slopes)
+        merged[specification.levels] = (
+            intercept or specification.intercept,
+            slopes,
+        )
+
+    return tuple(
+        FixedEffectSpecification(
+            levels=levels,
+            intercept=intercept,
+            slopes=tuple(slopes),
+        )
+        for levels, (intercept, slopes) in merged.items()
+    )
+
+
 def _term_from_ast(node: ast.expr) -> Term:
     """Parse one extracted Python expression as one Formulaic term."""
     expression = ast.unparse(node)
@@ -345,8 +368,10 @@ class Formula:
         """Fixed effects represented as symbolic `within::Effect` terms."""
         if not self.is_fixed_effects:
             return ()
-        return tuple(
-            FixedEffectSpecification.from_term(term) for term in self.fixed_effects
+        return _merge_fixed_effect_specifications(
+            tuple(
+                FixedEffectSpecification.from_term(term) for term in self.fixed_effects
+            )
         )
 
     @property
--- a/pyfixest/estimation/models/feglm_.py
+++ b/pyfixest/estimation/models/feglm_.py
@@ -148,6 +148,7 @@ class Feglm(Feols):
             self._Y.drop(na_separation, axis=0, inplace=True)
             self._X.drop(na_separation, axis=0, inplace=True)
             self._fe.drop(na_separation, axis=0, inplace=True)
+            self._fe_slopes.drop(na_separation, axis=0, inplace=True)
             self._data.drop(na_separation, axis=0, inplace=True)
             if self._weights_df is not None:
                 self._weights_df.drop(na_separation, axis=0, inplace=True)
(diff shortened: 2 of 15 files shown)


pyfixest/estimation/formula/parse.py
```diff
@@ -345,8 +368,10 @@ def fixed_effect_specifications(self) -> tuple[FixedEffectSpecification, ...]:
         """Fixed effects represented as symbolic `within::Effect` terms."""
         if not self.is_fixed_effects:
             return ()
-        return tuple(
-            FixedEffectSpecification.from_term(term) for term in self.fixed_effects
+        return _merge_fixed_effect_specifications(
```
leostimpfle, line 371: Could we preserve the structural formula intercept choice for slope-only effects? `Formula.exogenous` removes the intercept term for every fixed-effect model, and `second_stage` then reconstructs `Y ~ X1`. For `Y ~ X1 - 1 | f1[[X2]]`, reparsing that formula restores an implicit intercept; because the slope-only effect does not absorb it, the fitted model incorrectly includes `Intercept`. This also needs to be preserved in IV first- and second-stage matrices.

pyfixest/estimation/models/feglm_.py
```diff
@@ -148,20 +148,21 @@ def prepare_model_matrix(self):
             self._Y.drop(na_separation, axis=0, inplace=True)
             self._X.drop(na_separation, axis=0, inplace=True)
             self._fe.drop(na_separation, axis=0, inplace=True)
+            self._fe_slopes.drop(na_separation, axis=0, inplace=True)
```
leostimpfle: The `"fe"` separation check above receives every level column through `fe=self._fe`, including slope-only effects. A slope-only term such as `g[[z]]` does not add a group intercept, so an all-zero outcome group is not necessarily separated when `z` varies within the group; treating it as an ordinary FE can incorrectly remove valid rows. Only intercept-bearing effects should be passed to the `"fe"` check. The `"ir"` path also needs to reuse the active `within` demeaner for its auxiliary `feols` fits; otherwise varying-slope models fall back to `MapDemeaner` and raise.

pyfixest/estimation/models/feols_.py
```diff
@@ -461,14 +464,15 @@ def prepare_model_matrix(self):
         self._depvar = self._Y.columns[0]
 
         self._has_fixef = self._fe is not None
-        self._fixef = (
-            str(self.FixestFormula.fixed_effects).replace(" ", "")
-            if self.FixestFormula.is_fixed_effects
-            else None
-        )
+        if self.FixestFormula.is_fixed_effects:
+            self._fixef = "+".join(
```
leostimpfle, line 468: This makes `_fixef` contain only the fixed-effect level names, which is useful for downstream data lookup, but `summary()` also uses `_fixef` as its display label. Consequently, `feols("Y ~ X1 | f1[Z1]", ...)` prints `Fixed effects: f1` and hides the varying slope. Could the summary display the symbolic fixed-effect formula from `FixestFormula.fixed_effects` (or a separate display value) while keeping `_fixef` in its operational level-only form?

s3alfisc:
This is awesome! Very cool 🚀

tests/test_vs_fixest.py
```diff
@@ -1541,6 +1541,126 @@ def get_data_r(fml, data):
...
 
+@pytest.mark.against_r_core
+@pytest.mark.parametrize(
+    ("estimator", "fml", "weights"),
+    [
+        ("feols", "Y ~ X1 | f1[X2]", None),
+        ("feols", "Y ~ X1 | f1[[X2]]", "weights"),
+        ("feols", "Y ~ Z2 | f1[X2] | X1 ~ Z1", None),
+        ("fepois", "Y ~ X1 | f1[X2]", None),
+        ("feglm", "Y_bin ~ X1 | f1[X2]", None),
```
leostimpfle, line 1593: Add terms with multiple slopes, e.g., `f1[Z1, Z2]`

PR #1451, opened 2026-08-03
perf: halve model matrix construction time

schroedk:
Two changes to model matrix construction. No behavior change — `prepare_model_matrix` is roughly twice as fast.

## What

**`create_model_matrix`** built the set of NA-dropped rows with `set(range(n)).difference(lhs.index)`: a Python set holding every row label, with `set.difference` iterating the kept `RangeIndex` one element at a time. On a million rows that is 68ms, and it produces the empty set whenever nothing was dropped — the common case. Now a boolean mask.

**`ModelMatrix._process`** dropped rows by label list and read the `dependent` / `fixed_effects` properties (each one a full column-block copy) for checks that only need the column names. It now selects rows with a mask and inspects the stored column lists. `Feols.prepare_model_matrix` re-derived the surviving rows with `index.isin`, which `na_index` already records; it now drops those labels directly and does nothing when nothing was dropped.

`ModelMatrix` also owns the dropped-label set instead of mutating one passed in by its caller, and keeps it as the `frozenset` its only readers — the demean caches — actually want.

+49 / −37 across two files.

## Runtime

AKM panel, 3 FE (worker × firm × year), 1 covariate. `prepare_model_matrix` / whole `feols` call:

| rows | before | after | feols before | feols after |
|---:|---:|---:|---:|---:|
| 100k | 22ms | 14ms | 42ms | 35ms |
| 1M | 174ms | 90ms | 356ms | 258ms |
| 3M | 524ms | 275ms | 1.03s | 0.79s |
| 6M | 1.06s | 0.54s | 2.29s | 1.77s |
| 10M | 1.76s | 0.88s | 3.82s | 2.96s |
| 20M | 3.58s | 1.78s | 15.2s | 13.9s |

The cost is linear in rows and the saving is a flat **178 → 89 ns/row**, so the halving holds at every size rather than being amortized away.

Per-row cost does depend on fixed-effect cardinality. In a long panel with fixed cardinality (50k workers × 5k firms, growing T) it is 44 ns/row rather than 89, and `prepare_model_matrix` still accounts for a stable ~35% of a `feols` call from 1M to 20M rows.

## Testing

`test_vs_fixest` and `against_r_core` show no new failures — the only failures are pre-existing flaky `torch_mps` float32 comparisons, which vary between 28 and 32 across repeat runs on unmodified `master`.

A 25-scenario differential harness against `master` (coefficients, SEs, `_N`, `_na_index`, surviving index, warning strings) is byte-identical on all scenarios: inf/NaN in Y, X, weights and offset, singleton drops, `log(0)`, IV, weights, `i()`, `C()`, all-integer data, `fepois`, `quantreg`, multiple estimation, `split=`, `predict`/`fixef` after dropping, and the empty and single-row frames. Verified under pandas 2.3.3 as well as 3.0.1.

## Note for reviewers

Both rewrites rely on labels equalling positions, which `create_model_matrix` establishes by resetting the index in place. Where `master` degraded silently, the new code raises: an out-of-range label gives `IndexError` in `_dropped_rows`, and a missing label gives `KeyError` in the `feols_.py` drop. No caller can reach either today, and the precondition is documented at `_dropped_rows`.

### DIFF
--- a/pyfixest/estimation/formula/model_matrix.py
+++ b/pyfixest/estimation/formula/model_matrix.py
@@ -365,8 +365,9 @@ def create_model_matrix(
         }
         | {**capture_context(context)},
     )
-    drop_rows: set[int] = set(range(n_observations)).difference(
-        model_matrix[_ModelMatrixKey.main]["lhs"].index
+    drop_rows = _dropped_rows(
+        kept=model_matrix[_ModelMatrixKey.main]["lhs"].index,
+        n_observations=n_observations,
     )
     return ModelMatrix(
         model_matrix,
@@ -374,6 +375,17 @@ def create_model_matrix(
         drop_singletons=drop_singletons,
         drop_intercept=drop_intercept,
     )
+
+
+def _dropped_rows(kept: pd.Index, n_observations: int) -> set[int]:
+    """Row labels in `range(n_observations)` that `kept` does not contain.
+
+    `create_model_matrix` resets the data index beforehand, so row labels and
+    row positions coincide and the complement can be taken with a mask.
+    """
+    is_kept = np.zeros(n_observations, dtype=bool)
+    is_kept[kept.to_numpy()] = True
+    return set(np.flatnonzero(~is_kept).tolist())
 
 
 def _get_formulaic_formula(
--- a/pyfixest/estimation/formula/model_matrix.py
+++ b/pyfixest/estimation/formula/model_matrix.py
@@ -158,7 +158,7 @@ class ModelMatrix:
         n_dropped = int(is_dropped.sum())
         if not n_dropped:
             return
-        self._na_index |= frozenset(self._data.index[is_dropped].tolist())
+        self._na_index = self._na_index.union(self._data.index[is_dropped].tolist())
         self._data = self._data.loc[~is_dropped]
         warnings.warn(f"{n_dropped} {reason} dropped from the model.")
 
--- a/pyfixest/estimation/formula/model_matrix.py
+++ b/pyfixest/estimation/formula/model_matrix.py
@@ -7,6 +7,7 @@ import formulaic
 import numpy as np
 import pandas as pd
 from formulaic.parser import DefaultFormulaParser
+from numpy.typing import NDArray
 
 from pyfixest.core.detect_singletons import detect_singletons
 from pyfixest.estimation.formula import FORMULAIC_FEATURE_FLAG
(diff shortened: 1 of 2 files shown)


s3alfisc:
Oh, wow!

pyfixest/estimation/formula/model_matrix.py
```diff
@@ -141,19 +143,24 @@ def _process(self, dropped_rows: set[int], drop_singletons: bool = False) -> Non
...
+
+    def _drop(self, is_dropped: NDArray[np.bool_], reason: str) -> None:
+        """Drop the masked rows from `self._data` and add their labels to `na_index`.
+
+        `reason` completes the warning "{n} {reason} dropped from the model."
+        """
+        n_dropped = int(is_dropped.sum())
+        if not n_dropped:
+            return
+        self._na_index |= frozenset(self._data.index[is_dropped].tolist())
```
leostimpfle: Nitpick but I think we can avoid generating the intermediate`frozenset` using

```python
self._na_index = self._na_index.union(self._data.index[is_dropped].tolist())
```

leostimpfle (commented):
Thanks @schroedk! Looks good to me apart from a minor nitpick

pyfixest/estimation/formula/model_matrix.py
```diff
@@ -101,37 +108,32 @@ def _collect_columns(self, model_matrix: formulaic.ModelMatrix) -> None:
...
-    def _process(self, dropped_rows: set[int], drop_singletons: bool = False) -> None:
-        if self.dependent.shape[1] != 1:
+    def _process(self, drop_singletons: bool = False) -> None:
+        if self._dependent is None or len(self._dependent) != 1:
             # If the dependent variable is not numeric, formulaic's contrast encoding kicks in
             # creating multiple columns for the dependent variable
             # TODO: Make this check more explicit?
             raise TypeError("The dependent variable must be numeric.")
-        if self.endogenous is not None and self.endogenous.shape[1] != 1:
+        if self._endogenous is not None and len(self._endogenous) != 1:
```
s3alfisc, line 123: Do we have both `self.endogeneous` and `self.endogenous` now @leostimpfle ?

s3alfisc: Ah yes, maybe we could make the naming a bit clearer in a separate PR?

leostimpfle: Yeah it's not ideal (colums versus data)

pyfixest/estimation/formula/model_matrix.py
schroedk: Done

leostimpfle (approved):

PR #1456, opened 2026-08-08
Document fitted-model result accessors

layan985:


docs/getting-started.qmd
```diff
@@ -75,6 +75,34 @@ pf.etable(
...
+
+For models with fixed effects, within R-squared measures are also available:
+
+```{python}
+fit_fe._r2_within
+fit_fe._adj_r2_within
+```
+
+The leading underscore indicates that these model-level attributes are internal attributes rather than part of the stable public method API. See the [Function Reference](reference/index.qmd) for the full set of model methods and attributes.
+
```
leostimpfle: I would remove this as it's standard Python convention rather than anything specific to `pyfixest`

leostimpfle (changes requested):
Thanks! This looks good to me apart form one comment

leostimpfle (approved):

leostimpfle:
LGTM! @s3alfisc I'll let you decide whether to merge

s3alfisc:
Thanks both!

PR #1460, opened 2026-08-12
fix: honor predict() lsqr tolerance after caching

CodingSelim:
`predict()` takes `atol`/`btol` for the lsqr solve that recovers the fixed effects, but the call site only checked `if self._sumFE is None`, so once the fixed effects were cached the tolerance you passed later was ignored and you silently got whatever the first call computed.

```python
fresh = pf.feols(fml, data=df)
tight = fresh.predict(newdata=df, atol=1e-12, btol=1e-12)   # err 9.4e-13

cached = pf.feols(fml, data=df)
cached.predict(newdata=df)                                   # caches at 1e-6
cached.predict(newdata=df, atol=1e-12, btol=1e-12)           # err 6.5e-06
```

same arguments, seven orders of magnitude apart depending on what ran before. i checked against an exact lstsq solve on the FE dummies, which reproduces the fitted values to ~5e-15, so the tight one is the right answer.

only recomputing when the request is tighter than what's cached, never looser, since going the other way would throw away precision someone explicitly asked for. `test_fixef_interacted_labels` already relies on that, it calls `fixef(atol=1e-12)` then a default `predict()` and expects the accurate values to survive.

added a regression test covering both directions. ran the non-R suite against a clean-tree baseline and the failure set is identical, though i couldn't run the rpy2 comparisons locally so those are down to CI.

### DIFF
--- a/docs/changelog.qmd
+++ b/docs/changelog.qmd
@@ -177,6 +177,11 @@ We also removed:
   no not-yet-treated observations), instead of failing later with an opaque
   broadcasting `ValueError` in the vcov computation
   ([#1244](https://github.com/py-econometrics/pyfixest/issues/1244)).
+- `predict()` now honors its `atol` and `btol` arguments on every call. The fixed
+  effects recovered by `lsqr()` were cached after the first call and reused
+  regardless of the tolerance requested later, so `predict(atol=1e-12)` silently
+  returned values computed at the earlier, looser tolerance. Cached fixed effects
+  that are already tighter than requested are still reused.
 
 
 ## PyFixest 0.60.0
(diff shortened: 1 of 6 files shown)


s3alfisc:
Thanks @CodingSelim! I'll have to think about this one for a day, but I think you get it right. Could you please update the docstrings for the `atol` and `btol` args with the info that whenever users set the args lower, we recompute?

s3alfisc:
Plus, could you move the test into `test_predict_resid_fixef` or the test file with a similar name?

CodingSelim:
done both. the `atol`/`btol` docs on `predict()` now spell out that passing a value below the cached one recomputes the fixed effects at the tighter tolerance while a looser one reuses them, and i put the same note on the `fepois` and `feglm` predict overrides since they document the args separately. test moved into `test_predict_resid_fixef.py`.

one thing i noticed while in there, `fepois.predict` and `feglm.predict` both list `atol` and `btol` twice in the same Parameters block, looks like an old copy paste. left it alone to keep this diff small.

PR #1499, opened 2026-09-03
refactor(estimation): retain formula state and route lifecycle hooks

s3alfisc:
## Summary

Base layer of the estimation-state stack; #1516 and #1518–#1525 are stacked on it (#1500 and #1501 were closed as superseded).

Fitted models now retain the formula `ModelMatrix` as read-only formula state instead of copying its roles into a separate record, and the generic fit runner delegates response validation, post-fit work, and result expansion through a structural `FittedModel` protocol instead of estimator-name dispatch. Numerical results are unchanged.

- GLM separation filters the sample through the new `ModelMatrix.without_rows()`, which returns a copy and records the dropped rows in `na_index`. On master the post-separation `_na_index` was stale, so the demeaning and preconditioner cache key described the pre-separation sample, and `_Y_untransformed` kept its pre-separation length.
- The runner no longer passes `data=` to `vcov()`: `vcov()` already reads the model's retained estimation data, and `QuantregMulti` children carry their own.
- `QuantregMulti.get_performance()` is removed; the old `spec.method == "feols"` gate never reached it.
- The architecture guide records the pre-refactor estimator-state lifecycle and the stack's vocabulary.

## Verification

Release contract 1355/1355 against pyfixest 0.60.0, fast live-R 22/22, targeted Python suites, broad Python-only suite (pre-existing `torch_mps` demeaner-backend failures excluded), ruff, and mypy all pass. Deferred to CI: full R suites, HAC, docs render.

## Compatibility / risks

Retaining the `ModelMatrix` keeps the design frame alive for the model's lifetime: about 20 MB on a 200k x 13 design, roughly doubling per-model retained memory. `lean=True` and `store_data=False` free it, and the layers stacked above are its first consumers.

### DIFF
--- a/pyfixest/estimation/formula/model_matrix.py
+++ b/pyfixest/estimation/formula/model_matrix.py
@@ -1,7 +1,10 @@
+from __future__ import annotations
+
+import copy
 import warnings
 from collections.abc import Mapping
 from dataclasses import dataclass
-from typing import Any, Final
+from typing import Any, Final, TypeAlias, cast
 
 import formulaic
 import numpy as np
@@ -15,6 +18,8 @@ from pyfixest.estimation.formula.formulaic_compat import flatten_model_matrix
 from pyfixest.estimation.formula.parse import Formula
 from pyfixest.estimation.formula.utils import _get_weights
 from pyfixest.utils.utils import capture_context
+
+_ModelSpecMapping: TypeAlias = Mapping[str, formulaic.ModelSpec]
 
 
 @dataclass(frozen=True, kw_only=True)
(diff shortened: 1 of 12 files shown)


pyfixest/estimation/formula/model_matrix.py
```diff
@@ -26,6 +30,47 @@ class _ModelMatrixKey:
     offset: str = "offset"
 
 
+@dataclass(frozen=True, slots=True, kw_only=True)
+class FormulaData:
```
leostimpfle: I like this! Just curious: Is there are reason for why `FormulaData` is needed separately to `ModelMatrix`? Why not keep `ModelMatrix` as the single source of truth?

s3alfisc: I do not think it is needed; basically i was about to say the same thing to my LLM agent but then they all decided to be on a break this afternoon ...

PR #1507, opened 2026-09-04
refactor(estimation): split Feols into BaseRegression and an OLS leaf

s3alfisc:
Layer 5 of the estimation-state stack, based on `refactor/capability-contract` (#1504).

`Feols` no longer plays three roles. A new `BaseRegression` owns everything every estimator shares: construction, formula publication, weights and within-data helpers, storage guards, the capability contract, vcov dispatch, inference, tidy accessors, report binding, and the shared post-estimation methods with estimator hooks in place of `_method` comparisons. `Feols` is the OLS leaf, `Feiv` extends it, and `Feglm` and `Quantreg` are separate leaves. Consumers are typed against a `FittedResult` protocol, `FixestMulti` is generic in its result class, and the public entry points return precise types. Numerical results are invariant.

- **Breaking:** the OLS-only methods `wildboottest`, `ccv`, `decompose`, `update`, `evalue`, and `pvalue_savi` are defined on `Feols` only. A GLM, Poisson, or quantile result raises `AttributeError` carrying the capability reason instead of `NotImplementedError`; fit-level restrictions are unchanged.
- Fixed: the GLM `predict` overrides silently dropped `interval="prediction"`; it now raises like `se_fit=True`.
- `maketables` recognizes results via `isinstance(model, (Feols, Fepois, Feiv))`, so `pyfixest.report` registers an extractor that tests `BaseRegression`. The upstream fix belongs in maketables.
- quartodoc resolves inherited members, so no reference links move.

## Verification

Passed locally at the head: release contract 1355/1355 after every slice, `tests/test_vs_fixest.py` 5104 passed, fast live-R 22/22, targeted suites, broad Python-only suite (pre-existing `torch_mps` failures excluded), quartodoc build, ruff, mypy. Deferred to CI: HAC, docs render.


pyfixest/estimation/models/base_regression_.py
```diff
@@ -0,0 +1,2389 @@
...
+            )
+
+        return WithinLinearData(
+            response=within_data.response,
+            design=design,
+            instruments=within_data.instruments,
+            endogenous=within_data.endogenous,
+        )
+
+    def _set_within_data(self, within_data: WithinLinearData) -> None:
```
leostimpfle, line 630: Do we need these methods mutating the class's state or could we just use properties that delegate to the underlying object (`within_data` in this case)? For example,

```python
@property
def _k(self) -> int:
    return self._within_data.design.shape[1]
```

leostimpfle (commented):
Just a simple question after a superficial glance :)

PR #1519, opened 2026-09-05
refactor(estimation): add named demeaning cache

s3alfisc:
## Summary

The demeaning cache now stores named float64 arrays instead of a DataFrame, selecting cached columns by position and returning them read-only in the model's requested order, so repeated fits reuse demeaned data without rebuilding frames or risking one model corrupting another's columns. For the moment the DataFrame-shaped entry point survives as demean_yx_frames, which converts at the boundary and returns pd.DataFrames so linear and IV call sites are unchanged; https://github.com/py-econometrics/pyfixest/pull/1520 drops it and moves both to the array-native demean_yx.

### DIFF
--- a/docs/changelog.qmd
+++ b/docs/changelog.qmd
@@ -51,15 +51,20 @@ fit3 = pf.feols("Y ~ X1 + X2 | f1", data = df)
 
 ### Inference Types
 
-- Weighted models now reject `ccv()` explicitly. `Feols.update()` remains a
-  coefficient-only calculation and rejects `inplace=True`, weights, IV, and
-  non-OLS models instead of mixing scales.
-- Non-Poisson GLMs reject `CRV3` instead of silently refitting through the
+- `ccv()` now rejects weighted models instead of estimating from a mix of
+  weighting scales.
+- `ccv()` now raises `NotImplementedError` instead of `AssertionError` for
+  model types it does not support.
+- `Feols.update()` now rejects `inplace=True`. It remains a coefficient-only
+  Sherman-Morrison calculation, and no longer leaves the fitted result in a
+  partially updated state.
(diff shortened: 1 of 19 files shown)


s3alfisc:
The demeaning cache now stores named `float64` arrays instead of a `DataFrame`, selecting cached columns by position and returning them read-only in the model's requested order, so repeated fits reuse demeaned data without rebuilding frames or risking one model corrupting another's columns. For the moment the DataFrame-shaped entry point survives as `demean_yx_frames`, which converts at the boundary and returns `pd.DataFrame`s so linear and IV call sites are unchanged; #1520 drops it and moves both to the array-native `demean_yx`.

Review follow-ups since the last push, as three self-contained commits:

- `refactor(demean): name the cache locals after what they hold` — the `YX`/`YX_demeaned`/`yx_names` vocabulary replaces `requested_data`/`requested_values`, `uncached_*` replaces `new_*`, `used` becomes `used_preconditioner`, and `DemeanedData`'s docstring gains an `Attributes` section with the frozen/read-only rationale under `Notes`.
- `fix(demean): hand out read-only arrays from every selection path` — selecting a cached prefix returned a read-only view while selecting out of order returned a writable copy, so whether a model could write into its own demeaned data depended on the order the shared cache happened to fill up in. Both paths are read-only now, asserted across cache miss, in-order hit, and out-of-order hit.
- `docs(changelog): describe what the demeaning cache layer changes` — the previous entry described only the column-order naming and omitted the array-backed cache entry.

Checks: `tests/test_demean.py` (50 passed, 2 skipped) and `test-release-contract` (1355 passed, no drift) both green on the pushed tip; the naming commit is covered by an equivalent run on its own state.

PR #1533, opened 2026-09-10
refactor(estimation): expose canonical fitted components

s3alfisc:
Expose fitted-model components for inspection and replace superseded private array aliases.
Other things: align multi-quantile retention with OLS, enable child predictions and covariance updates, and drop GLM residuals under lean=True.


s3alfisc:
Expose fitted-model components for inspection and replace superseded private array aliases.
Other things: align multi-quantile retention with OLS, enable child predictions and covariance updates, and drop GLM residuals under lean=True.

PR #1564, opened 2026-09-15
fix(report): select iplot coefficients from the i() term registry

s3alfisc:
`iplot()` now derives `i()` coefficient names from formulaic's term registry instead of matching `::` in coefficient names, mirroring how fixest's `iplot` consults `model_matrix_info`. A regressor whose column name contains `::` was previously plotted as an interaction; it now raises the existing "no i() syntax" error. The private `Feols._icovars` attribute is removed; `report.visualize.iplot` was its only consumer.

The new helper `i_term_columns` lives in `formulaic_compat.py` next to the prediction-time `i()` helpers and reuses their contrast-state marker. It reads the retained model spec, so it works under `lean=True` and `store_data=False`.

Checks: targeted compat, plot, DiD, and error tests, ruff, and mypy passed. No numerical change, so no live-R run; release contract not run (no local baseline).


s3alfisc:
@leostimpfle if you have the time, could you take a quick look? I'll let you know when, still haven't really checked too thoroughly myself.

pyfixest/estimation/formula/formulaic_compat.py
```diff
@@ -164,6 +164,32 @@ def rows_with_unseen_contrast_levels(
...
+    pyfixest's ``__contrasts_<var>__`` key, so a factor is an ``i()`` call
+    exactly when its encoder state carries such a key. A term contributes its
+    columns when any of its factors is such a call, which also covers
+    ``i(f):x`` style interactions. Column names are never inspected, so user
+    columns containing ``::`` are not misclassified.
+    """
+    i_factor_exprs = {
+        factor_expr
+        for factor_expr, value in rhs_spec.encoder_state.items()
+        if any(is_contrast_state_key(key) for key in _unpack_encoder_state(value)[1])
```
leostimpfle, line 182: Stumbled over the naming here because it wasn't immediately clear to me that `is_contrast_state_key` only refers to `i`-encoded variables. (`formulaic` also refers to "contrasts".)

tests/test_formulaic_compat.py
```diff
@@ -152,6 +153,47 @@ def test_encoder_state_guard_raises_loudly() -> None:
...
+    ids=["no_i", "i", "i_continuous", "i_in_interaction"],
+)
+def test_i_term_columns(data: pd.DataFrame, fml: str, expected: list[str]) -> None:
+    """i_term_columns returns exactly the columns produced by i() terms."""
+    fit = pf.feols(fml, data=data)
+
+    columns = i_term_columns(fit._model_spec["second_stage"].rhs)
+
+    assert columns == expected
+    assert set(columns) <= set(fit._coefnames)
```
leostimpfle, line 173: Nitpick: I would find `set(columns).issubset(fit._coefnames)` more readable

pyfixest/report/visualize.py
```diff
@@ -202,12 +204,13 @@ def iplot(
         rename_models = {}
 
     for x, fxst in enumerate(list(models)):
-        if fxst._icovars is None:
+        icovars = i_term_columns(fxst._model_spec[_ModelMatrixKey.main].rhs)
+        if not icovars:
             raise ValueError(
                 f"The {x} th estimated model did not have ivars / 'i()' model syntax."
```
leostimpfle, line 210: Nitpick (and unrelated to your change): Would change the message to something like

`f"Model {x+1} cannot be plotted with `iplot` because it does not use the `i` syntax: {fxst._fml!r}"`

tests/test_plots.py
```diff
@@ -159,6 +159,15 @@ def test_iplot_error(data):
         iplot(fit4)
 
 
+@pytest.mark.extended
+def test_iplot_rejects_double_colon_column_without_i(data):
+    """A '::' in a column name must not be mistaken for i() syntax."""
+    renamed = data.rename(columns={"X1": "a::b"})
+    fit = feols(fml="Y ~ Q('a::b') + X2", data=renamed, vcov="iid")
+    with pytest.raises(ValueError, match="did not have ivars"):
```
leostimpfle, line 167: Update if error message in `iplot` is changed (see above)

s3alfisc:
Thanks for the feedback @leostimpfle!

